PR #2, opened 2021-04-08
Install the runner

chris-sanders:


src/charm.py
```diff
@@ -2,58 +2,77 @@
...
-                {"fortune": "A bug in the code is worth two in the documentation."}
-            )
+        self.unit.status = MaintenanceStatus("Installing runner")
+        self._runner.download()
+        self._runner.setup_env()
+        self.unit.status = BlockedStatus("Waiting for registration")
+
+    def _on_register_action(self, event):
+        if self._stored.registered:
+            pass  # TODO raise this an an error
```
Cynerva, line 36: Why not do this now?

src/charm.py
```diff
@@ -2,58 +2,77 @@
...
+# def _on_fortune_action(self, event):
+#     # Note: you need to uncomment the example in the actions.yaml file for this to work (ensure
+#     # to not just leave the example, but adapt to your needs for actions commands)
+#     fail = event.params["fail"]
+#     if fail:
+#         event.fail(fail)
+#     else:
+#         event.set_results(
+#             {"fortune": "A bug in the code is worth two in the documentation."}
+#         )
```
Cynerva, line 75: Consider removing these commented out methods.

src/charm.py
```diff
@@ -2,58 +2,77 @@
...
-
     _stored = StoredState()
 
     def __init__(self, *args):
         super().__init__(*args)
         self.framework.observe(self.on.install, self._on_install)
-        self.framework.observe(self.on.config_changed, self._on_config_changed)
-        self.framework.observe(self.on.fortune_action, self._on_fortune_action)
-        self._stored.set_default(things=[])
+        # self.framework.observe(self.on.config_changed, self._on_config_changed)
```
Cynerva, line 23: Consider removing this commented out line.

tests/unit/test_charm.py
```diff
@@ -2,34 +2,36 @@
...
 
 class TestCharm(unittest.TestCase):
-    def test_config_changed(self):
+    # def test_config_changed(self):
+    #     harness = Harness(GithubRunnerOperator)
+    #     self.addCleanup(harness.cleanup)
+    #     harness.begin()
+    #     self.assertEqual(list(harness.charm._stored.things), [])
+    #     harness.update_config({"thing": "foo"})
+    #     self.assertEqual(list(harness.charm._stored.things), ["foo"])
```
Cynerva, line 18: Consider removing this commented out method.

Cynerva (changes requested):
Looks pretty good, but I'd like to see the copyright notices fixed, and an answer for juju-no-proxy handling. Either implementation to handle it, or documentation (in the charm README?) to warn users of the limitation.

src/runner.py
```diff
@@ -0,0 +1,70 @@
...
+        if http_proxy:
+            self.proxies["http"] = http_proxy
+        if https_proxy:
+            self.proxies["https"] = https_proxy
+        self.session = requests.Session()
+        if self.proxies:
+            self.session.proxies.update(self.proxies)
+        self.jinja = Environment(loader=FileSystemLoader("templates"))
+
+    def download(self):
```
johnsca, line 33: It's generally recommended to handle assets like this as resources, to ensure that all units are running the same version and to provide an explicit upgrade mechanism. That said, the runner by definition needs to have access to the GH API server and using that directly means we don't have to manage a copy of it.

tests/integration/test_charm.py
```diff
@@ -1,25 +1,33 @@
...
-
-import yaml
 
 import pytest
+from runner import Runner
 
 log = logging.getLogger(__name__)
 
 
+async def file_contents(unit, path):
```
johnsca, line 9: Maybe useful to add this to pytest-operator.

tests/unit/test_runner.py
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env python3
+from unittest.mock import patch
+
+
+@patch("shutil.chown")
+def test_download(chown, runner):
+    runner.download()
```
johnsca, line 7: Technically, this isn't a unit test since it performs an actual download, meaning this could fail depending due to network or disk issues.

johnsca (approved):
LGTM for a first pass but I left some non-blocking comments below.

johnsca:
Ah, @Cynerva beat me to it.  +1 to his requested changes.

src/charm.py
```diff
@@ -2,58 +2,77 @@
 # Copyright 2021 Chris
```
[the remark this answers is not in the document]
chris-sanders: Ha, I didn't even notice that. This is what charmcraft creates. I'll update it.

templates/env.j2
```diff
@@ -0,0 +1,4 @@
+http_proxy={{proxies['http']}}
+https_proxy={{proxies['https']}}
```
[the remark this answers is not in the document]
chris-sanders: Good catch, I'll add that too.

src/charm.py
chris-sanders: Yup I'll add a log and return the status.

src/charm.py
chris-sanders: Will do.

actions.yaml
```diff
@@ -5,10 +5,12 @@
...
 # It ties in to the example _on_fortune_action handler in src/charm.py
 # 
- fortune:
-   description: Returns a pithy phrase.
-   params:
-     fail:
-       description: "Fail with this message"
-       type: string
-       default: ""
+register:
```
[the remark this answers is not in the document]
chris-sanders: Because, as far as I can tell Github requires user intervention. Tokens are provided 1 at a time and are time limited.

src/runner.py
chris-sanders: The runner also does automatic upgrades, so you can't guarantee the version since it will self update before it runs a job if it is behind.

tests/unit/test_runner.py
chris-sanders: I agree, I started with this for faster testing, and then checked that the download worked in the integration testing. Do you think it's best to put a pytest.skip on it or just remove it? It's a lot easier to test the downloading locally and I just never went back and removed it. It does seem helpful so I'm inclined to put it behind a flag or something.

tests/integration/test_charm.py
chris-sanders: I agree, I dug this up from my previous charms. Do you think it belongs in pytest-operator or libjuju?

src/charm.py
```diff
@@ -1,58 +1,59 @@
...
-            )
+        self.unit.status = MaintenanceStatus("Installing runner")
+        self._runner.download()
+        self._runner.setup_env()
+        self.unit.status = BlockedStatus("Waiting for registration")
+
+    def _on_register_action(self, event):
+        if self._stored.registered:
+            event.set_results({"status": "Registration aborted, already registered"})
+            logger.error("Can not re-register, unit is already registered")
```
Cynerva, line 36: Missing a `return` here? Where's the abort?

Cynerva (changes requested):

src/charm.py
chris-sanders: I guess I removed it when I added the logging, thanks I'll put that back in.

Cynerva (approved):
LGTM

PR #5, opened 2021-04-16
Spawn runners in containers

chris-sanders:
Use a personal access token to allow automatic runner registration and
removal. Each runner is now installed in a LXD container and runs one
job after which the container is destroyed. The charm checks on a
configuable interval how many runners it has and creates new ones to
maintain a configured quantity.


johnsca (commented):
A few comments below, but I was also trying to figure out if there was a way to add an action to the charm to generate / rotate a token which would have to be set on the org / repo as a GitHub secret and could then be validated before launching the job. I think that just having the GH secret on the org / repo should be enough to prevent external PRs from launching the runner, but if we somehow forgot to include it on a repo, it could re-open the hole, so it would be nice if it could be validated on each job request.

src/charm.py
```diff
@@ -3,58 +3,105 @@
...
+    def _on_stop(self, event):
+        self._remove_cron("check-runners")
+        self._runner.remove_runners()
+
+    def _add_cron(self, event, interval):
+        """Add a cron job for the provided event to run at the provided interval."""
+        root_cron = CronTab(user="root")
+        unit = self.unit.name
+        dispatch = str(self.charm_dir / "dispatch")
+        command = f"juju-run {unit} 'JUJU_DISPATCH_PATH={event} {dispatch}'"
```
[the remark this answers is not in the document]
chris-sanders: This fires an event, which runs _reconcile_runners in the charm. While that does use the runner, it does use context from the juju config to understand how many runners are suppose to be available and the type of virtualization to use. It also sets unit status and provides logging so you're aware what the runner is doing. I don't think I can easily decouple this process, and the charm needs to be able to do it as well so I'd prefer not to have 2 code paths doing the same thing in different ways.

chris-sanders:
I really don't love this solution, but I'll go ahead and explain it.
```python
    def create(self, image, virt="container", wait=False):
        """Create a runner"""
        instance = self._create_instance(image=image, virt=virt)
        instance.start(wait=True)
        time.sleep(5)  # Wait seems to be aggressive
```
I'm able to in very specific situations get pylxd to become unresponsive to commands. At first it was file.put which failed, but replacing that with a subprocess then triggers later when commands are executed in the container. I've determined that after starting an instance, if you wait before proceeding this no longer happens. Hence using pdb makes it always pass.

I have triggered this with a 1 second sleep, but I've yet to see it trigger with a 5 second sleep. I've reviewed the information that LXD has available and I can't find anything that would tell me the instance is running. Wait already waits for the status, which is what you would expect to have to wait for. Any work around that executes code in the container, has the potential to trigger the issue as well. Therefore the only option I've found is to simply sleep a few seconds after starting the container.

johnsca:
Oh, one other thing.  It might be worth calling out that doing a `juju destroy-model` doesn't give the charm a chance to clean up the runners.

chris-sanders:
> Oh, one other thing. It might be worth calling out that doing a `juju destroy-model` doesn't give the charm a chance to clean up the runners.
> 
> **Edit:** Actually, I thought it was a bit odd that removing the model didn't let the charm clean up (it's actually `destroy-controller` that does an unclean shutdown), so I tested it again by just doing `juju remove-application` and it said that it was running the `stop` hook in the status but the runners didn't get removed from the repo. Not sure what's going on there.

I think I know what's going on here. When shutting down this charm takes one more pass to remove *offline* runners. I haven't implemented enough logic to track runners/unit. I think that would be a nice feature to add, but didn't go that far with the first implementation. Because it doesn't know if another unit could be hosting the runners it only removes runners that github reports as being offline. I could try and stop the LXD's on the instance before cleaning up but it takes a while for the runners to show up as offline and I didn't want to try and loop waiting until I was convinced they were offline. The better solution would be to add unit name's to the runner names and nuke all runners for the charm on exit.

I don't think it's a big lift to add, but I am hesitant to do so right now as I'm concerned I could be wrong and additional errors will have to be addressed.

README.md
```diff
@@ -2,11 +2,26 @@
...
+after which the LXD is destroyed preventing the job from interacting with any other jobs. The charm will periodically check the number of currently active runners and spawn a new
+runner as necessary. Both the interval of the check and the number of runners to maintain are configurable.
 
 ## Usage
 
-TODO
+There are two mandatory configuration options, `path` and `token`. Path is used to determine the organization or repository that the runner will be registered to. This value can
+not be changed, the charm should be redeployed with a new value instead. The token is a github personal access token. This token requires `repository-specific agents: repo` for
+repositories and `organisation-wide agents: repo and admin:org` for use with an organization. This is necessary because the charm will be creating and removing runners in order to
+make runners single use which protects jobs from leaking information to other jobs run on the same runner.
```
johnsca, line 14:
```suggestion
not be changed, the charm should be redeployed with a new value instead. The token is a [GitHub Personal Access Token (PAT)](https://github.com/settings/tokens) (note: this is not the same as the token given in the Add a Runner instructions). This token requires `repository-specific agents: repo` ("Full control of private repositories" on the website) for
use with repositories and `organisation-wide agents: repo and admin:org` ("Full control of private repositories" and "Full control of orgs and teams, read and write org projects" on the website) for use with an organization. This is necessary because the charm will be creating and removing runners in order to
make runners single use which protects jobs from leaking information to other jobs run on the same runner.
```

src/charm.py
```diff
@@ -3,58 +3,116 @@
...
+        comment = f"Charm cron for {event}"
+
+        comments = [c for c in root_cron.find_comment(comment)]
+        if comments:
+            return  # Already exists
+
+        unit = self.unit.name
+        dispatch = str(self.charm_dir / "dispatch")
+        command = f"juju-run {unit} 'JUJU_DISPATCH_PATH={event} {dispatch}'"
+        job = root_cron.new(command=command, comment=comment)
```
johnsca, line 103: As written, it won't update the interval if it changes.

```suggestion
        try:
            job = next(root_cron.find_comment(comment))
        except StopIteration:
            unit = self.unit.name
            dispatch = str(self.charm_dir / "dispatch")
            command = f"juju-run {unit} 'JUJU_DISPATCH_PATH={event} {dispatch}'"
            job = root_cron.new(command=command, comment=comment)
```

johnsca (changes requested):
Suggestions below.

johnsca (approved):

PR #9, opened 2021-05-19
Privileged is required for snaps in unconfined containers

chris-sanders:
Apparently you need to be privileged *and* unconfined to output stdout in a classic snap in a LXD container. The other option is remove both priviliged and the unconfined profile and it will also run but be denied access to stdout. I'm leaning toward getting the debug logging.

### DIFF
--- a/src/runner.py
+++ b/src/runner.py
@@ -225,6 +225,7 @@ class Runner:
         if not self.lxd.profiles.exists("runner"):
             config = {
                 "security.nesting": "true",
+                "security.privileged": "true",
                 "raw.lxc": "lxc.apparmor.profile=unconfined",
             }
             devices = {}


johnsca:
I don't see a way around this other than a fundamental change to the architecture of how the ephemeral runners are managed. We can't reasonably accept no info about why a bootstrap fails if we consider that we are likely going to want to change or modify the environments at some point (the initial lack of bootstrap failure info directly blocked realizing that the vSphere provider needed some additional bootstrap params).

Eventually, I'd really like to see the runner operator be able to manage runner instances at the same virtualization level as itself (i.e., being a "stack" which is able to manage the scale of runner instance units).

johnsca (approved):

PR #10, opened 2021-05-21
Load apparmor profiles

chris-sanders:
After several tests, I've found that loading the profiles via apparmor_parser allows juju to run. However, I believe it's racing as some calls return an error (239). I've not seen any errors on the first call to /etc/apparmor.d/*snap-confine*.

I've also added an option to not split the command that is passed into the check_output call. This was necessary so to launch bash for the shell expansion.

I've done a fresh install with this build, and it has been getting past bootstrap again. This run was made against this build of the charm: https://github.com/charmed-kubernetes/charm-kubernetes-worker/runs/2635431830?check_suite_focus=true

### DIFF
--- a/src/runner.py
+++ b/src/runner.py
@@ -69,6 +69,7 @@ class Runner:
                 ],
             )
             self._start_runner(instance)
+            self._load_aaprofile(instance)
         except RuntimeError as e:
             instance.stop(wait=True)
             raise e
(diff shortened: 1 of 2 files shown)


Cynerva (commented):
> I believe it's racing as some calls return an error (239).

What happens in this case? Will the runner work as it should?

johnsca:
We've been using this for some time and it's been working, so going ahead with merge.  There's a backlog item to support VM-based runners.

PR #11, opened 2021-09-03
Refactor runner management logic

### LINKED ISSUE #4: Unregister runner
Runners can be removed from Github directly as well as via a command on the runner. If a runner can unregister itself it would be good to implement an action to handle this and unregistered when the unit is removed. This requires looking at the removal token it may not be possible to self-remove without an admin providing a token.

johnsca:
This refactor addresses several issues with how the runners are managed and kept up to date, including:

  * There were several possible infinite loops.
  * The runner binary update check was done every time a runner was created.
  * The runner binary update check wasn't using the API token and thus could hit the lower unauthenticated API limits.
  * Due to any and all of the above, the charm could get stuck in a blocked hook execution which would prevent it from recovering any runners.
  * Local runners could end up unregistered and never cleaned up, eventually blocking replacement runner registration.

Fixes #4

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -11,10 +11,14 @@ options:
       default: 3
       type: int
       description: "Quantity of runners to maintain"
-  check-interval:
-      default: "*/1 * * * *"
-      type: string
-      description: "Cron interval to check runners, each deployed unit will check on this interval"
+  reconcile-interval:
+      default: 5
+      type: int
+      description: "How frequently, in minutes, the runners should be reconciled"
+  update-interval:
+      default: 60
+      type: int
+      description: "How frequently, in minutes, the charm should check for new versions of the runner binary"
   virt-type:
       default: "container"
       type: string
--- a/src/charm.py
+++ b/src/charm.py
@@ -9,13 +9,17 @@ from ops.charm import CharmBase
 from ops.framework import EventBase, StoredState
 from ops.main import main
 from ops.model import ActiveStatus, BlockedStatus, MaintenanceStatus
-from runner import Runner
+from runner import RunnerManager, RunnerError
 
 logger = logging.getLogger(__name__)
 
 
-class CheckRunnersEvent(EventBase):
-    """Event to request a runner check"""
+class ReconcileRunnersEvent(EventBase):
+    """Event representing a periodic check to ensure runners are ok."""
+
+
+class UpdateRunnerBinEvent(EventBase):
+    """Event representing a periodic check for new versions of the runner binary."""
 
 
 class GithubRunnerOperator(CharmBase):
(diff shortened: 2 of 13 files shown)


src/charm.py
```diff
@@ -3,115 +3,267 @@
...
+        if not (token and path):
+            return None
+        return RunnerManager(path, token, self.app.name, self.config["virt-type"])
 
     def _on_install(self, event):
         self.unit.status = MaintenanceStatus("Installing packages")
-        self._runner.install()
+        RunnerManager.install_deps()
+        runner_manager = self._get_runner_manager()
+        if runner_manager:
```
knkski, line 75: As much as I like `Option`s from other languages, since you're already using exceptions below, seems like you might want to embrace that and have `self._get_runner_manager()` throw an exception and catch it instead of checking for `None` here.

src/charm.py
```diff
@@ -3,115 +3,267 @@
...
+            self.unit.status = MaintenanceStatus("Installing runner binary")
+            try:
```
knkski, line 77: Seems like you could simplify this to be something like:

```python
try:
    self.unit.status = MaintenanceStatus('foo')
    foo()
    self.unit.status = MaintenanceStatus('bar')
    bar()
except RunnerError as e:
    logger.exception(e)
    self.unit.status = BlockedStatus(e.message)
except Exception as e:
    logger.exception(e)
    self.unit.status = BlockedStatus(str(e))

```

src/charm.py
```diff
@@ -3,115 +3,267 @@
...
+        self._disable_event_timer("reconcile-runners")
+        runner_manager = self._get_runner_manager()
+        if runner_manager:
+            try:
+                runner_manager.clear()
+            except Exception:
+                # Log but ignore error since we're stopping anyway.
+                logger.exception("Failed to clear runners")
+
+    def _render_event_tmpl(self, tmpl_type, event_name, context):
```
knkski, line 234: Minor nit, but this seems better named as `_write_event_tmpl`. Offhand, I was surprised that `self._render_event_tmpl(...)` didn't return a rendered template

src/charm.py
```diff
@@ -3,115 +3,267 @@
...
+        if runner_bin_url != self._stored.runner_bin_url:
+            self.unit.status = MaintenanceStatus("Updating runner binary")
+            try:
+                runner_manager.update_runner_bin(runner_bin_url)
+            except Exception as e:
+                logger.exception("Failed to update runner binary")
+                self.unit.status = BlockedStatus(f"Failed to update runner binary: {e}")
                 return
+            self._stored.runner_bin_url = runner_bin_url
+            # TODO: Flush existing runners? What if they're processing a job?
```
knkski, line 133: Does this need to exist at the Juju level? I ask because a number of services in Charmed Kubeflow handle the reconciliation of similar things themselves. The equivalent here would probably be something like the operator installing an extra systemd timer that handles flushing existing runners if they're done processing everything and there's a new binary.

If this needs to exist at the Juju level (i.e. supporting `juju run-action update-runner-bin`), then the above doesn't work as well

config.yaml
```diff
@@ -11,10 +11,14 @@ options:
       default: 3
```
knkski, line 11: I can't comment on this directly in this file since it's not being changed, but we may want to use something like `org_or_repo` instead of `path`. When reading the code, I thought `path` was referring to where the binary was stored locally or something, but that's a different variable.

johnsca:
Much of this would be obviated by https://github.blog/changelog/2021-09-20-github-actions-ephemeral-self-hosted-runners-new-webhooks-for-auto-scaling/ if we can figure out how to make the webhook part work to be notified of incoming worker requests.

src/charm.py
```diff
@@ -3,114 +3,267 @@
...
+            self.unit.status = MaintenanceStatus("Starting runners")
+            try:
+                runner_manager.reconcile(self.config["quantity"])
+            except RunnerError as e:
+                logger.exception("Failed to start runners")
+                self.unit.status = BlockedStatus(f"Failed to start runners: {e}")
+            else:
+                self.unit.status = ActiveStatus()
+        else:
+            self.unit.status = BlockedStatus("Missing token or org/repo path config")
```
addyess, line 93: It feels to me @knkski was suggesting something like this:


```suggestion
        RunnerManager.install_deps()    # raises RunnerInstallDepsError, results in a charm error
        try:
            runner_manager = self._get_runner_manager()   # could raise RunnerConfigError
            self.unit.status = MaintenanceStatus("Installing runner binary")
            self._stored.runner_bin_url = runner_manager.get_latest_runner_bin_url()  # Could raise RunnerBinaryError
            runner_manager.update_runner_bin(self._stored.runner_bin_url)  # Could raise RunnerBinaryError
            self.unit.status = MaintenanceStatus("Starting runners")
            runner_manager.reconcile(self.config["quantity"])  # Could raise RunnerReconcileError
        except RunnerConfigError:
            logger.exception("Failed to update runner binary")
            self.unit.status = BlockedStatus("Missing token or org/repo path config")
        except RunnerBinaryError as e:
            logger.exception("Failed to update runner binary")
            self.unit.status = BlockedStatus(f"Failed to update runner binary: {e}")
        except RunnerReconcileError as e:
            logger.exception("Failed to start runners")
            self.unit.status = BlockedStatus(f"Failed to start runners: {e}")
        else:
            self.unit.status = ActiveStatus()
```

templates/dispatch-event.timer.j2
```diff
@@ -0,0 +1,11 @@
+[Unit]
+Description=Timer to dispatch {{event}} event periodically
+Requires=ghro.{{event}}.service
+
+[Timer]
+Unit=ghro.{{event}}.service
+OnUnitInactiveSec={{interval}}m
+RandomizedDelaySec={{jitter}}m
```
addyess, line 8: oh, ok you're using systemd to emit `juju-run action` events on the current agent -- eliminating the need for cron.  Nifty

tox.ini
```diff
@@ -1,5 +1,8 @@
 [flake8]
 max-line-length = 88
+ignore =
+  E203
```
addyess, line 4: whitespace before ':' operator?
https://www.flake8rules.com/rules/E203.html

addyess (approved):
I didn't see anything glaring here, though i think ken's comment does make it more readable.  However these changes would lead to MORE refactor... so i'm 50/50

tox.ini
addyess: [answering my own question](https://github.com/charmed-kubernetes/github-runner-operator/pull/11/files#diff-535292833fc918d3f2f1afb1208a622104ad63cf55e4b4b522292c06ba84c369R184)

PR #16, opened 2021-11-24
Replace once with ephemeral

### LINKED ISSUE #15: Ephemeral runners created with the deprecated `--once` flag instead of `--ephemeral`
See the following blog post, for instance https://github.blog/changelog/2021-09-20-github-actions-ephemeral-self-hosted-runners-new-webhooks-for-auto-scaling/

jack-w-shaw:
Use the newly supported `--ephemeral` on configure instead of `-once` on run.

This means that runners are automatically de registered once they have completed a job. Meaning charm code can be simplified somewhat.

In doing so, I have also included a bug fix which means runners are now correctly de registered when units are stopped

Fixes #15


johnsca (approved):
LGTM, nice.

Issue #12, opened 2021-10-07
Improve VM support


johnsca:
Currently, a given deployment can only have either LXD runners or VM runners, but not a mix of both. The config would need to change to support it, but the logic is mostly in place already and it would make managing mixed sets of runners much easier.

r-f-g:
@ijansky @addyess This problem was solved by PR #18, right?

However, I tried to use VM in github-runner and they are unusable due to low resources. VMs are created with 1 cpu and 1GB ram. Can we add `limits.cpu=4` and `limits.memory=8GiB` to `runner` profile, or add them as a config option?

addyess:
So, does the github runner really need a huge beefy machine to run a little node script?  I suppose the work being asked of the VM may be intense -- which kinda makes you wonder if the test environment should spawn a larger VM to do the dirty work

Here are the docs for github hosted runners:
https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners

> Hardware specification for Windows and Linux virtual machines:
> 2-core CPU
> 7 GB of RAM memory
> 14 GB of SSD disk space

Do you suppose these specs would serve as a better default in the VM profile?

I support `limits.cpu=2` and `limits.memory=7GiB`.  is this fair?

r-f-g:
I think if we want to run `microk8s` or `microstack` on these runners, then we need a bigger VM.
`microstack` is recommended to be used with a multi-core cpu, 8 GB RAM and 100 GB disk space according to [doc](https://ubuntu.com/tutorials/install-openstack-on-your-workstation-and-launch-your-first-instance#1-overview).
`microk8s` is recommended to be used with 4 GB RAM and 20 GB disk space according to [doc](https://microk8s.io/docs/getting-started).

I understand that in CI it doesn't make much sense to always have huge VMs, but this is only used to run functional tests, so I see potential there. At least if I want to have those runners just for a specific repo, it should be up to me to decide on the size of those VMs.
Because of this, I think that these values could be configurable via config options.

addyess:
cool, this is reasonable. can you raise a PR to address it?

r-f-g:
Yes, sure.

r-f-g:
Now I think that this issue could be closed. What do you think @addyess?

addyess:
yeah, super job

PR #18, opened 2022-02-05
Improve VM support

ijansky:
Enable a mix of containers and virtual machines as instance runners. This PR should address issue #12.

There are a few other changes that were necessary to build the charm.

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -7,15 +7,15 @@ options:
       default: ""
       type: string
       description: "Personal Access Token, required for auto registerting with the given url"
-  quantity:
-      default: 3
-      type: int
-      description: "Quantity of runners to maintain"
   check-interval:
       default: "*/1 * * * *"
       type: string
       description: "Cron interval to check runners, each deployed unit will check on this interval"
-  virt-type:
-      default: "container"
-      type: string
-      description: "Virtualization type for runners, either container or virtual-machine"
+  containers:
+      default: 3
+      type: int
+      description: "Quantity of container runners to maintain"
+  virtual-machines:
+      default: 0
+      type: int
+      description: "Quantity of virtual machine runners to maintain"
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,6 @@
 ops
 requests
-jinja2
+jinja2 < 3.0.0
 ghapi
 pylxd
 python-crontab
--- a/src/charm.py
+++ b/src/charm.py
@@ -20,6 +20,10 @@ class CheckRunnersEvent(EventBase):
 
 class GithubRunnerOperator(CharmBase):
     _stored = StoredState()
+    _virt_type_config_options = {
+        "container": "containers",
+        "virtual-machine": "virtual-machines",
+    }
 
     def __init__(self, *args):
         super().__init__(*args)
@@ -71,21 +75,24 @@ class GithubRunnerOperator(CharmBase):
         event.set_results({"status": "Completed runner check"})
 
     def _reconcile_runners(self):
-        delta = self.config["quantity"] - self._runner.active_count()
-        logger.info(f"Reconciling {delta} runners")
-        while delta > 0:
-            self.unit.status = MaintenanceStatus(f"Installing {delta} runner(s)")
-            try:
-                self._runner.create(image="ubuntu", virt=self.config["virt-type"])
-            except RuntimeError as e:
-                logger.error("Failed to create runner")
-                logger.error(f"Error: {e}")
-            delta = self.config["quantity"] - self._runner.active_count()
+        for virt_type in self._virt_type_config_options.keys():
+            target_quantity = self.config[self._virt_type_config_options[virt_type]]
+            delta = target_quantity - self._runner.active_count(virt_type)
+            logger.info(f"Reconciling {delta} runners")
+            while delta > 0:
+                self.unit.status = MaintenanceStatus(f"Installing {delta} runner(s)")
+                try:
+                    self._runner.create(virt_type, "ubuntu")
+                except RuntimeError as e:
+                    logger.error("Failed to create runner")
+                    logger.error(f"Error: {e}")
+                delta = target_quantity - self._runner.active_count(virt_type)
         self.unit.status = ActiveStatus("Active and registered")
 
     def _on_stop(self, event):
         self._remove_cron("check-runners")
-        self._runner.remove_runners()
+        for virt_type in self._virt_type_config_options.keys():
+            self._runner.remove_runners(virt_type)
 
     def _add_cron(self, event, interval):
         """Add a cron job for the provided event to run at the provided interval."""
--- a/src/runner.py
+++ b/src/runner.py
@@ -22,6 +22,7 @@ logger = logging.getLogger(__name__)
 class Runner:
     release_url = "https://api.github.com/repos/actions/runner/releases/latest"
     runner_path = Path("/opt/github-runner")
+    runner_prefix = "runner-"
     env_file = runner_path / ".env"
 
     def __init__(self, path, token):
@@ -55,10 +56,10 @@ class Runner:
         cmd = "sudo lxd init --auto"
         subprocess.check_output(cmd.split())
 
-    def create(self, image, virt="container", wait=False):
+    def create(self, virt_type, image, wait=True):
         """Create a runner"""
-        instance = self._create_instance(image=image, virt=virt)
-        instance.start(wait=True)
+        instance = self._create_instance(virt_type, image, wait)
+        instance.start(wait)
         try:
             self._install_binary(instance)
             self._configure_runner(instance)
@@ -71,24 +72,22 @@ class Runner:
             self._start_runner(instance)
             self._load_aaprofile(instance)
         except RuntimeError as e:
-            instance.stop(wait=True)
+            instance.stop(wait=wait)
             raise e
 
-    def active_count(self):
+    def active_count(self, virt_type):
         """Return the number of active runners"""
-        count = 0
-        for container in self.lxd.containers.all():
-            if container.name.startswith("runner-"):
-                count += 1
-        return count
+        return len(self._get_instances(virt_type))
 
-    def remove_runners(self):
+    def remove_runners(self, virt_type):
         """Remove runners"""
         api = self.api
         repo = None
         if "/" in self.path:
             owner, repo = self.path.split("/")
-        hosted_runners = [container.name for container in self.lxd.containers.all()]
+        hosted_runners = [
+            container.name for container in self._get_instances(virt_type)
+        ]
         for runner in self._get_runners()["runners"]:
             if runner.name in hosted_runners:
                 logger.info(f"Deregistering runner {runner.name}")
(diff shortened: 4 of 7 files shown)


requirements.txt
```diff
@@ -1,6 +1,6 @@
 ops
 requests
-jinja2
+jinja2 < 3.0.0
```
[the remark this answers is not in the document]
ijansky: Good catch, addressed.

addyess:
There's a mix of changes going in with https://github.com/charmed-kubernetes/github-runner-operator/pull/11 that will require some merging / refactoring of this PR.  Would you have time/be willing to adjust after those changes?

src/charm.py
```diff
@@ -71,21 +75,24 @@ def _on_check_runners_action(self, event):
...
+                    logger.error("Failed to create runner")
+                    logger.error(f"Error: {e}")
+                delta = target_quantity - self._runner.active_count(virt_type)
         self.unit.status = ActiveStatus("Active and registered")
 
     def _on_stop(self, event):
         self._remove_cron("check-runners")
-        self._runner.remove_runners()
+        for virt_type in self._virt_type_config_options.keys():
+            self._runner.remove_runners(virt_type)
```
addyess, line 95: Why does the `_virt_type_config_options` dict exist in charms?  it seems linked to the `Runner` class.  If you pushed it into there, you wouldn't have to have these loops in `charms.py`. 

Maybe if this was in `Runner.virt_types` you could iterate on it when you needed it in `charms.py` but leave `remove_runners()` and `active_count` etc down in there. 

It seems like there are operations where you don't care if it's a container or VM, you just want to remove all the runners, or reconcile all the runners.  Dealing with which type isn't specifically important. Here's kinda what i'm thinking

```python
def remove_runners(virt_type=None):
    """
    Removes all the runners of a specific type, or if None specified remove all.
    """
    ...
```

config.yaml
```diff
@@ -7,15 +7,15 @@ options:
...
-      description: "Quantity of runners to maintain"
   check-interval:
       default: "*/1 * * * *"
       type: string
       description: "Cron interval to check runners, each deployed unit will check on this interval"
-  virt-type:
-      default: "container"
-      type: string
-      description: "Virtualization type for runners, either container or virtual-machine"
+  containers:
```
addyess, line 14: changing this name from `quantity` to `containers` will impact already existing deployments.  Can you deprecate `quantity` but leave it in there and support it?

addyess (changes requested):
A few changes around config and some information hiding

ijansky:
Thank you for the review. I will definitely find some time to rewrite this pull request once #11 lands in the master.

src/charm.py
ijansky: The reason for keeping it in the charm.py was to make runner.py virt_type agnostic and to access related config values in one place. I will take a look at your suggestion though.

config.yaml
ijansky: Is there a recommended way of deprecating config values, or is it acceptable to add a prefix to the description and code adequate logic in the code? 
I suppose that virt-type should be taken into account as well, correct? It would make sense I guess.

addyess: Usually, add a description to the existing config like:

```yaml
    description: |
       DEPRECATED:  use `containers` instead
       <existing description>.  
```

then, make `containers` default to something like `None` or some such field where you inherit the config value of `quantity`.   I've seen this as a general rule, not sure if it's best practice.

ijansky:
The source branch was updated with the latest code from the master branch and it is ready for a new review.

src/runner.py
```diff
@@ -52,6 +52,10 @@ def is_unregistered(self):
             return False
         return self.local and self.local.status == "Running"
 
+    @property
+    def virt_type(self):
+        return self.local.type if self.local else None
```
addyess, line 57: i haven't tried this but can this be 
```suggestion
        return (self.local or self.remote).type
```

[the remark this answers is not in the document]
addyess: trying this out locally, i found the remote runners do provide these labels

src/runner.py
```diff
@@ -158,32 +161,33 @@ def get_info(self):
...
+    def reconcile(self, virt_type, quantity):
         """Bring runners in line with target."""
         active_runners = []
         stale_runners = []
         for runner in self.get_info():
-            if runner.is_active:
-                active_runners.append(runner)
-            else:
-                stale_runners.append(runner)
+            if runner.virt_type == virt_type:
```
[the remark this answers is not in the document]
ijansky: Now that I think about it, the current implementation will probably not scale with additional units. If the reconciliation method finds 4 containers and 4 virtual machines, it will not create new ones, no matter how many units were deployed. That does not sound like the expected behavior. 

I will test it further.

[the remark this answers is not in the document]
addyess: I'm glad you caught that from the README.  I guess we can do whatever is necessary.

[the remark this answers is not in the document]
ijansky: Testing the code on more than one unit confirmed that units were operating on runners that did not belong to them (runners that were not running on the unit). This is fixed now. The scaling behavior is aligned with the documentation:
* when a unit is added, the unit spawns N containers and M virtual machines;
* when a unit is removed, only containers and virtual machines started by that unit are removed.

src/runner.py
```diff
@@ -27,34 +27,20 @@ def __init__(self, name, local, remote):
         self.remote = remote
 
     @property
-    def is_active(self):
```
ijansky, line 30: The logic is simplified to distinguish online and offline runners.

src/runner.py
```diff
@@ -111,6 +97,7 @@ def install_deps(cls):
         """Install dependencies"""
         subprocess.run(["snap", "install", "lxd"], check=True)
         subprocess.run(["lxd", "init", "--auto"], check=True)
+        subprocess.run(["apt", "install", "-qy", "cpu-checker", "libvirt-clients", "libvirt-daemon-driver-qemu"], check=True)
```
ijansky, line 100: Install packages necessary for QEMU based virtual machines.

src/runner.py
```diff
@@ -171,47 +158,54 @@ def get_info(self):
...
-        active_runners = []
-        stale_runners = []
-        for runner in self.get_info():
-            if runner.virt_type == virt_type:
-                if runner.is_active:
-                    active_runners.append(runner)
-                else:
-                    stale_runners.append(runner)
+        runners = [r for r in self.get_info() if r.virt_type == virt_type]
+        online_runners = [r for r in runners if r.is_online and r.is_local]
```
ijansky, line 162: The `r.is_local` condition ensures that the RunnerManager operates only on runners that belong to its unit.

src/runner.py
```diff
@@ -171,47 +158,54 @@ def get_info(self):
...
             logger.info(f"Removing extra {virt_type} runners: {runner_names}")
             for runner in old_runners:
                 self._remove_runner(runner)
         return delta
 
     def clear(self):
-        """Clear out all existing runners."""
-        runners = self.get_info()
+        """Clear out existing local runners."""
+        runners = [r for r in self.get_info() if r.is_local]
```
ijansky, line 189: Do not stop runners belonging to other units.

src/runner.py
```diff
@@ -171,47 +158,54 @@ def get_info(self):
...
 
     def create(self, image, virt="container"):
         """Create a runner"""
         instance = self._create_instance(image=image, virt=virt)
         instance.start(wait=True)
+
+        # Wait for the instance to boot
+        for attempt in range(8):
+            try:
+                self._check_output(instance, "who")
```
ijansky, line 203: The call to `instance.start(wait=True)` returns as soon as the instance started, which does not mean that it has fully booted up and is ready to execute commands. Let's run a few wait iterations to give the instance time to boot up.

src/runner.py
addyess: agree.  In fact this is 
```suggestion
        local_online_runners = [r for r in runners if r.is_online and r.is_local]
```

src/runner.py
```diff
@@ -158,48 +166,56 @@ def get_info(self):
...
-            else:
-                stale_runners.append(runner)
-
-        if stale_runners:
-            runner_names = ", ".join(r.name for r in stale_runners)
-            logger.info(f"Cleaning up stale runners: {runner_names}")
-            for runner in stale_runners:
+        runners = [r for r in self.get_info() if r.virt_type == virt_type]
+        online_runners = [r for r in runners if r.is_online and r.is_local]
+        offline_runners = [r for r in runners if r.is_offline]
```
addyess, line 173: and this just lists all offline runners -- which all need to be cleaned up, local or not.  This too is ok

src/runner.py
```diff
@@ -158,48 +166,56 @@ def get_info(self):
...
+        # Clean up stale (offline) runners
+        if offline_runners:
+            runner_names = ", ".join(r.name for r in offline_runners)
+            logger.info(f"Cleaning up stale {virt_type} runners: {runner_names}")
+            for runner in offline_runners:
                 self._remove_runner(runner)
 
-        delta = num_runners - len(active_runners)
+        # Add/Remove runners to match the target quantity
+        delta = quantity - len(online_runners)
```
addyess, line 183: This answers the question "how many more runners do I need to start locally"
```suggestion
        delta = quantity - len(local_online_runners)
```

src/runner.py
```diff
@@ -158,48 +166,56 @@ def get_info(self):
...
         if delta > 0:
-            logger.info(f"Adding {delta} additional runners")
+            logger.info(f"Adding {delta} additional {virt_type} runners")
             for i in range(delta):
-                self.create(image="ubuntu", virt=self.virt_type)
+                self.create(image="ubuntu", virt=virt_type)
         elif delta < 0:
-            active_runners.sort(key=lambda r: r.created_at)
```
addyess, line 183: Does this disagree with the docs?  Which is the right path?  Immediately removing the runners or letting them run eventually and not renewing them when they're gone

> Decreasing the `containers` or `virtual-machines` config values will not shut down runners.

src/runner.py
ijansky: Looking at this from the perspective of someone using the charm and has access to a "slow" cloud, then failing the deployment because of a too short deadline will be a showstopper. Catching it in a try-except block will not solve it, nor does the 2 minute interval. At the same time, it is not a good idea to wait indefinitely.
What about making the timeout configurable with a default of 2 minutes and wrapping the wait in a try-except block?

src/runner.py
ijansky: Yeah, I should check that the runner is in fact "Idle" and not active before stopping it. Good catch.

ijansky: I have rewritten the readme file - now it should be fully aligned with the actual behavior.

src/runner.py
ijansky: I used the `reconcile-interval` config value as the upper limit for how long to wait for an instance to start. It gives operators the ability to configure the `reconcile-interval` such that it is long enough for at least one runner to start. At the same time it "motivates" the operator to use a sufficiently long interval to ensure stability based on the environment.

addyess (approved):
LGTM  👍🏻

PR #23, opened 2022-11-01
More descriptive error in case of being unable to check for runner updates

mthaddon:
Currently we catch an exception while executing `runner_manager.get_latest_runner_bin_url()` in the `_on_install` method, but we don't catch the exception in the `_on_update_runner_bin` method. If someone deploys the charm with no `path` and `token` set, then nothing will be done in the `_on_install` method, but it could then fail later on when this is added.

I've confirmed that in an environment that doesn't have external access to github this will put the charm into BlockedStatus and report a message of `Failed to check for runner updates: <urlopen error [Errno 110] Connection timed out>` in juju status.

This PR also includes some linting fixes, and some minor copyright and doc updates.


jdkandersson (approved):

metadata.yaml
```diff
@@ -1,6 +1,7 @@
-# Copyright 2021 Chris
+# Copyright 2021 Canonical Ltd.
```
arturo-seijas, line 1: I miss several entries in this file: docs, issues, maintainers...

src/charm.py
```diff
@@ -121,8 +122,13 @@ def _on_update_runner_bin(self, event):
...
             return
         old_status = self.unit.status
-        self.unit.status = MaintenanceStatus("Checking for runner updates")
-        runner_bin_url = runner_manager.get_latest_runner_bin_url()
+        try:
+            self.unit.status = MaintenanceStatus("Checking for runner updates")
+            runner_bin_url = runner_manager.get_latest_runner_bin_url()
+        except urllib.error.URLError as e:
+            logger.exception("Failed to check for runner updates")
+            self.unit.status = BlockedStatus(f"Failed to check for runner updates: {e}")
```
arturo-seijas, line 130: Do we really want to block here? Was if it's a transient error?

yhaliaw: Maybe it is better to switch to Maintenance status, as update binary is being trigger on a schedule (default to 1 hour).
However, the default one hour retry is a bit long, so block status may be a good choice.

Currently, failure to get the URL in `_on_install` method also return blocked status. So that should be changed as well, if we are changing to maintenance status.

yhaliaw (approved):

metadata.yaml
mthaddon: Indeed - those can come in later changes though.

src/charm.py
mthaddon: Yeah, I'd agree we should leave this as `BlockedStatus` for now and re-assess switching this to something else later (and updating all other entries in this charm that do something similar at the same time).

mthaddon:
There are some test failures here, but these are testing on versions of python the service doesn't support, so I think we can merge this for now and deal with other clean up in future changes.

PR #26, opened 2022-12-13
Code linting and cleanup

yhaliaw:


yhaliaw:
The integration test not a part of the workflow. The next story is to rework the integration tests. It will be enable then.

The unittest coverage is below 85% currently. A following story is to convert all unit test to pytest and add more tests. The coverage will be enable with that story.

charmcraft.yaml
```diff
@@ -10,7 +10,7 @@ parts:
 bases:
   - build-on:
     - name: "ubuntu"
-      channel: "20.04"
+      channel: "22.04"
     run-on:
     - name: "ubuntu"
-      channel: "20.04"
+      channel: "22.04"
```
mthaddon, line 16: I think we should be careful about changing this. Doing series upgrades for machine charms is non-trivial, and means anyone running the current charm on focal won't see there's a new version available.

metadata.yaml
```diff
@@ -2,17 +2,17 @@
 # See LICENSE file for licensing details.
 name: github-runner
 display-name: GitHub runner
-description: |
+description: >
```
arturo-seijas, line 5: Can you also add `issues` and `maintainers`?

src/charm.py
```diff
@@ -48,29 +69,33 @@ def __init__(self, *args):
...
-            self.on.reconcile_runners_action, self._on_reconcile_runners_action
-        )
-        self.framework.observe(
-            self.on.flush_runners_action, self._on_flush_runners_action
-        )
+        self.framework.observe(self.on.check_runners_action, self._on_check_runners_action)
+        self.framework.observe(self.on.reconcile_runners_action, self._on_reconcile_runners_action)
+        self.framework.observe(self.on.flush_runners_action, self._on_flush_runners_action)
+
+    def _get_runner_manager(self, token: Optional[str] = None, path: Optional[str] = None):
```
arturo-seijas, line 76: Can you add the return types too?

src/charm.py
```diff
@@ -48,29 +69,33 @@ def __init__(self, *args):
...
+            token (Optional[str]): GitHub personal access token to manager the runners with.
+                Defaults to None.
+            path (str): GitHub repo path in the format '<org>/<repo>', or the GitHub org name.
+                Defaults to None.
+        """
         if token is None:
             token = self.config["token"]
         if path is None:
             path = self.config["path"]
         if not (token and path):
```
arturo-seijas, line 89: I think `if not toker or not path` would be easier to read

src/charm.py
```diff
@@ -94,19 +119,29 @@ def _on_install(self, event):
...
+            event (ConfigChangedEvent): Event of configuration change.
+        """
+        self._event_timer.ensure_event_timer("update-runner-bin", self.config["update-interval"])
+        self._event_timer.ensure_event_timer(
+            "reconcile-runners", self.config["reconcile-interval"]
+        )
 
         if self.config["path"] != self._stored.path:
-            prev_runner_manager = self._get_runner_manager(path=self._stored.path)
+            prev_runner_manager = self._get_runner_manager(path=str(self._stored.path))
```
arturo-seijas, line 142: Why do you need this cast?

mthaddon (changes requested):
I don't think we can just switch to jammy from focal since - haven't done a thorough review of other changes as this may affect other things (type hints, etc.).

src/event_timer.py
```diff
@@ -0,0 +1,92 @@
...
+            "event": event_name,
+            "interval": interval,
+            "jitter": interval / 4,
+            "timeout": timeout or (interval * 30),
+            "unit": self.unit_name,
+        }
+        self._render_event_template("service", event_name, context)
+        self._render_event_template("timer", event_name, context)
+        # Binding for systemctl do no exist, so `subprocess.run` used.
+        subprocess.run(["/usr/bin/systemctl", "daemon-reload"], check=True)  # nosec B603
```
arturo-seijas, line 71: Have you considered handling the exceptions that might be thrown here?

src/event_timer.py
```diff
@@ -0,0 +1,92 @@
...
+
+    def disable_event_timer(self, event_name: str):
+        """Disable the systemd timer for the given event.
+
+        Args:
+            event_name (str): Name of the juju event to disable.
+        """
+        # Don't check for errors in case the timer wasn't registered.
+        # Binding for systemctl do no exist, so `subprocess.run` used.
+        subprocess.run(  # nosec B603
```
arturo-seijas, line 87: Same as above

README.md
```diff
@@ -2,11 +2,11 @@
 
 ## Description
 
-This charm creates self-hosted GitHub runners. Each unit of this charm will start a configurable number of LXD based containers and virtual 
+This machine charm creates self-hosted GitHub runners. Each unit of this charm will start a configurable number of LXD based containers and virtual
 machines to host GitHub runners. Each runner performs only one job, after which it unregisters from GitHub to ensure that each job runs in
 a clean environment.
 
-The charm will periodically check the number of idle runners and spawn or destroy runners as necessary to maintain the configured number of 
+The charm will periodically check the number of idle runners and spawn or destroy runners as necessary to maintain the configured number of
```
arturo-seijas, line 9: "to maintain the configured number of runners" -> "to match the number provided by configuration"?

README.md
```diff
@@ -2,11 +2,11 @@
 
 ## Description
 
-This charm creates self-hosted GitHub runners. Each unit of this charm will start a configurable number of LXD based containers and virtual 
+This machine charm creates self-hosted GitHub runners. Each unit of this charm will start a configurable number of LXD based containers and virtual
```
arturo-seijas, line 5: "to host GitHub runners" -> "to host them"?

README.md
```diff
@@ -25,9 +25,9 @@ The number of runners on a single unit is configured using two configuration opt
 For example, if the charm is deployed with 2 units `juju deploy <charm> -n 2` and the `containers` value of 3 is in use,
 there will be a total of 6 container based runners, three on each unit.
 
-## Reconcilation
+## Reconciliation
 
-Each unit will periodically check the number of idle runners at an interval that is configured using the `check-interval` setting to maintain the configured number of runners. During the check, all offline runners are unregistered from GitHub and corresponding containers or virtual machines are destroyed. 
+Each unit will periodically check the number of idle runners at an interval that is configured using the `check-interval` setting to maintain the configured number of runners. During the check, all offline runners are unregistered from GitHub and corresponding containers or virtual machines are destroyed.
```
arturo-seijas, line 30: "Each unit will periodically check the number of idle runners at an interval that is configured using the `check-interval` setting to maintain the configured number of runners." -> "Each unit will periodically check the number of idle runners at the interval specified by `check-interval` to maintain the appropriate number."

arturo-seijas: "During the check, all offline" -> "During the check, all the offline"

config.yaml
```diff
@@ -1,49 +1,62 @@
 options:
   path:
-      default: ""
       type: string
-      description: "Either an organization name or repository in 'org/repo' format"
-  token:
       default: ""
+      description: >
+        The path of the GitHub repo to register the self-hosted runners. Either an organization name
```
arturo-seijas, line 6: "repo" -> "repository"

config.yaml
```diff
@@ -1,49 +1,62 @@
...
-      description: "Personal Access Token, required for auto registerting with the given url"
-  quantity:
-      default: 3
-      type: int
-      description: |
-        DEPRECATED: use `containers` and `virtual-machines`
-        Quantity of runners to maintain
+      default: ""
+      description:
+        The GitHub Personal Access Token for auto-registering the self-hosted runners.
```
arturo-seijas, line 12: "auto-registering" -> "registering"? or maybe "for the self-hosted runners to self register"?

config.yaml
```diff
@@ -1,49 +1,62 @@
...
       default: 0
+      description: >
+        The target quantity of container runners. This charm will spawn or destroy container
+        runners to match this setting.
+  virtual-machines:
       type: int
-      description: "Quantity of virtual machine runners to maintain"
+      default: 0
+      description: >
+        The target quantity of virtual machine runners. This charm will spawn or destroy virtual
```
arturo-seijas, line 23: "target quantity" -> number

config.yaml
```diff
@@ -1,49 +1,62 @@
...
+        machines runners to match this setting.
   vm-cpu:
-      default: 2
       type: int
-      description: "Number of CPUs for VMs."
+      default: 2
+      description: >
+        The number of CPUs used by each virtual machine runner. Does not affect container runners.
```
arturo-seijas, line 29: "by each" -> "per"

config.yaml
```diff
@@ -1,49 +1,62 @@
...
   vm-memory:
-    default: 7GiB
     type: string
-    description: "VM memory size."
+    default: 7GiB
+    description: >
+        The memory size used by each virtual machine runner. Does not affect container runners.
```
arturo-seijas, line 34: "by each" -> "per"

config.yaml
```diff
@@ -1,49 +1,62 @@
...
   vm-disk:
-    default: 10GiB
     type: string
-    description: "VM disk size."
+    default: 10GiB
+    description: >
+        The root disk size used by each virtual machine runner. Does not affect container runners.
```
arturo-seijas, line 39: "by each" -> "per"

config.yaml
```diff
@@ -1,49 +1,62 @@
...
   reconcile-interval:
-      default: 5
       type: int
-      description: "How frequently, in minutes, the runners should be reconciled"
+      default: 5
+      description: >
+        The number of minutes between each reconciliation between the current state of runners and
```
arturo-seijas, line 44: "The number of minutes" -> "Minutes"
"between the current state of runners" -> "between the current state"

config.yaml
```diff
@@ -1,49 +1,62 @@
...
+        the intended state of the runners. On reconciliation the charm polls the state of runners,
+        such as, the number of container runners, and see if it needs make adjustments.
   update-interval:
-      default: 60
       type: int
-      description: "How frequently, in minutes, the charm should check for new versions of the runner binary"
+      default: 60
+      description: >
+        The number of minutes between each check for new versions of the runner binary.
```
arturo-seijas, line 51: "The number of minutes" -> "Minutes"

metadata.yaml
```diff
@@ -2,17 +2,17 @@
...
   Creates a cluster of self-hosted github runners
-summary: |
-  This charm creates self-hosted GitHub runners. Each unit of this charm will start
-  a configurable number of LXD based containers and virtual machines to host GitHub
-  runners. Each runner performs only one job, after which it unregisters from GitHub
-  to ensure that each job runs in a clean environment.
+summary: >
+  This charm creates self-hosted GitHub runners. Each unit of this charm will start a configurable
+  number of LXD based containers and virtual machines to host GitHub runners. Each runner performs
```
arturo-seijas, line 9: "to host GitHub runners" -> "to host them"

src/charm.py
yhaliaw: For mypy type checking.
The `_get_runner_manager` method takes a `Optional[str]` but the `self._stored.path is `ObjectEvents` type.
Due to `self._stored` is `StoredState` object.

This mypy error shows up without casting to str:
```
static run-test: commands[0] | mypy /home/user/work/charm/github-runner-operator/src/ /home/user/work/charm/github-runner-operator/tests/
src/charm.py:144: error: Argument "path" to "_get_runner_manager" of "GithubRunnerOperator" has incompatible type overloaded function; expected "Optional[str]"  [arg-type]
```

charmcraft.yaml
yhaliaw: The problem is it is not possible to build the charm on 20.04 any more.

Doing `charmcraft pack` will result in the following error:
```
2022-12-14 09:16:40.188 :: 2022-12-14 01:16:39.876 package: rust-all
2022-12-14 09:16:40.188 :: 2022-12-14 01:16:39.879 Parts processing error: Package 'rust-all' has unmet dependencies: llvm-14.
```

This charm requires `rust-all` package.

On focal, the dependency `llvm-14` is not available:
https://packages.ubuntu.com/focal/rust-all

It is possible to install `rust-all` on jammy:
https://packages.ubuntu.com/jammy/rust-all

The requirement of `rust-all` stems from the python package `cryptography` which is needed by the python package `pylxd`.

Either we remove the dependency on `pylxd` or find a different way to manage the rust dependency (I already put this in the github runner design doc improvement list).

src/charm.py
yhaliaw: Fixed.

src/charm.py
yhaliaw: Fixed.

README.md
yhaliaw: fixed

README.md
yhaliaw: fixed

README.md
yhaliaw: fixed

config.yaml
yhaliaw: fixed

config.yaml
yhaliaw: fixed.

config.yaml
```diff
@@ -1,49 +1,62 @@
...
+      description:
+        The GitHub Personal Access Token for auto-registering the self-hosted runners.
   containers:
-      default: 0
       type: int
-      description: "Quantity of container runners to maintain"
-  virtual-machines:
       default: 0
+      description: >
+        The target quantity of container runners. This charm will spawn or destroy container
```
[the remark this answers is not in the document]
yhaliaw: fixed

config.yaml
yhaliaw: Fixed.

config.yaml
yhaliaw: fixed

config.yaml
yhaliaw: fixed

config.yaml
yhaliaw: fixed

config.yaml
yhaliaw: fixed

config.yaml
```diff
@@ -1,49 +1,62 @@
...
+        The root disk size used by each virtual machine runner. Does not affect container runners.
   reconcile-interval:
-      default: 5
       type: int
-      description: "How frequently, in minutes, the runners should be reconciled"
+      default: 5
+      description: >
+        The number of minutes between each reconciliation between the current state of runners and
+        the intended state of the runners. On reconciliation the charm polls the state of runners,
+        such as, the number of container runners, and see if it needs make adjustments.
```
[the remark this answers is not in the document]
yhaliaw: fixed

config.yaml
yhaliaw: fixed

metadata.yaml
yhaliaw: Fixed

pyproject.toml
```diff
@@ -0,0 +1,53 @@
...
+
+
+[tool.pytest.ini_options]
+minversion = "6.0"
+log_cli_level = "INFO"
+
+# Formatting tools configuration
+[tool.black]
+line-length = 99
+target-version = ["py38"]
```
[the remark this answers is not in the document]
merkata: My bad, was looking at `tox.ini` like I've mentioned, this is present in `pyproject.toml` in other charm repos :/

src/event_timer.py
yhaliaw: Updated.
Can you review again?

src/event_timer.py
yhaliaw: Updated.
Can you review again?

charmcraft.yaml
mthaddon: If it's possible to build on jammy and run on focal, that sounds good to me.

Could we add integration tests for focal and jammy? It looks like the (very basic) integration tests currently just test it deploys and goes into blocked state, and this would presumably be on the default series (which you're switching to jammy here) if I'm understanding it correctly.

src/charm.py
arturo-seijas: Thx for the explanation

src/charm.py
```diff
@@ -95,19 +122,39 @@ def _on_install(self, event):
...
+        """
+        try:
+            self._event_timer.ensure_event_timer(
+                "update-runner-bin", self.config["update-interval"]
+            )
+            self._event_timer.ensure_event_timer(
+                "reconcile-runners", self.config["reconcile-interval"]
+            )
+        except TimerEnableError as ex:
+            logger.exception("Failed to start the event timer")
```
arturo-seijas, line 147: I think it would be nice here to log the original error message to gain some insight on what did actually happened when executing the command

charmcraft.yaml
jdkandersson: The integration tests are the next Jira story

src/charm.py
```diff
@@ -1,18 +1,31 @@
...
 from ops.framework import EventBase, StoredState
 from ops.main import main
 from ops.model import ActiveStatus, BlockedStatus, MaintenanceStatus
-from runner import RunnerManager, RunnerError, VMResources
+
+from event_timer import EventTimer, TimerDisableError, TimerEnableError
+from runner import RunnerError, RunnerManager, VMResources
+
+if TYPE_CHECKING:
+    from ops.model import JsonObject
```
jdkandersson, line 28: Out of curiosity, what does this do?

src/charm.py
```diff
@@ -49,29 +70,35 @@ def __init__(self, *args):
...
+
+    def _get_runner_manager(
+        self, token: Optional[str] = None, path: Optional[str] = None
+    ) -> Optional[RunnerManager]:
+        """Get a RunnerManager instance, or None if missing config.
 
-    def _get_runner_manager(self, token=None, path=None):
-        """Get a RunnerManager instance, or None if missing config."""
+        Args:
+            token (Optional[str]): GitHub personal access token to manager the runners with.
```
jdkandersson, line 83: Do we need the type in the docstring if it is defined in the type hints?

src/charm.py
```diff
@@ -49,29 +70,35 @@ def __init__(self, *args):
...
+                Defaults to None.
+            path (str): GitHub repository path in the format '<org>/<repo>', or the GitHub
+                organization name. Defaults to None.
+        """
```
jdkandersson, line 87: Our standard includes a description of return values, do you think it is applicable here? https://github.com/canonical/is-charms-contributing-guide/blob/main/CONTRIBUTING.md#docstrings

src/charm.py
```diff
@@ -95,19 +122,39 @@ def _on_install(self, event):
...
+            logger.exception("Failed to start the event timer")
+            self.unit.status = BlockedStatus(
+                f"Failed to start timer for regular reconciliation and binary update checks: {ex}"
+            )
 
         if self.config["path"] != self._stored.path:
-            prev_runner_manager = self._get_runner_manager(path=self._stored.path)
+            prev_runner_manager = self._get_runner_manager(
+                path=str(self._stored.path)
+            )  # Casting for mypy checks.
```
jdkandersson, line 155: Out of curiosity, what is the type before the cast?

src/charm.py
```diff
@@ -141,7 +193,12 @@ def _on_update_runner_bin(self, event):
             # TODO: Flush existing runners? What if they're processing a job?
         self.unit.status = old_status
 
-    def _on_reconcile_runners(self, event):
+    def _on_reconcile_runners(self, event: ReconcileRunnersEvent) -> None:
+        """Handle the reconcile of runners.
```
jdkandersson, line 197: `reconcile` -> `reconciliation`?

src/charm.py
```diff
@@ -235,7 +317,15 @@ def _on_stop(self, event):
                 # Log but ignore error since we're stopping anyway.
                 logger.exception("Failed to clear runners")
 
-    def _reconcile_runners(self, runner_manager):
+    def _reconcile_runners(self, runner_manager: RunnerManager) -> Dict[str, "JsonObject"]:
```
jdkandersson, line 320: Out of curiosity, what is `JsonObject`?

src/event_timer.py
```diff
@@ -0,0 +1,112 @@
+# Copyright 2021 Canonical
+# See LICENSE file for licensing details.
+
+"""EventTimer for scheduling dispatch of juju event on regular intervals."""
+
+import subprocess  # nosec B404
```
jdkandersson, line 6: Would it be worth including a comment here why subprocess is needed?

src/event_timer.py
```diff
@@ -0,0 +1,112 @@
...
+
+    event: str
+    interval: float
+    jitter: float
+    timeout: float
+    unit: str
+
+
+class EventTimer:
+    """Manages the timer to emit juju event at regular intervals.
```
jdkandersson, line 32: `event` -> `events`?

src/event_timer.py
```diff
@@ -0,0 +1,112 @@
...
+    def __init__(self, unit_name: str):
+        """Construct the timer manager.
+
+        Args:
+            unit_name (str): Name of the juju unit to emit events to.
+        """
+        self.unit_name = unit_name
+        self._jinja = Environment(loader=FileSystemLoader("templates"), autoescape=True)
+
+    def _render_event_template(self, template_type: str, event_name: str, context: EventConfig):
```
jdkandersson, line 49: Would a docstring be valuable here?

src/event_timer.py
```diff
@@ -0,0 +1,112 @@
...
+        try:
+            # Binding for systemctl do no exist, so `subprocess.run` used.
+            subprocess.run(["/usr/bin/systemctl", "daemon-reload"], check=True)  # nosec B603
+            subprocess.run(  # nosec B603
+                ["/usr/bin/systemctl", "enable", f"ghro.{event_name}.timer"], check=True
+            )
+            subprocess.run(  # nosec B603
+                ["/usr/bin/systemctl", "start", f"ghro.{event_name}.timer"], check=True
+            )
+        except Exception as ex:
```
jdkandersson, line 90: Would it be worth being more specific about which exception to catch? Similar comment for the other `except` below

src/runner.py
```diff
@@ -1,53 +1,158 @@
...
+import pylxd.models
+import requests
 from ghapi.all import GhApi
+from jinja2 import Environment, FileSystemLoader
 from pylxd.exceptions import LXDAPIException, NotFound
 
 logger = logging.getLogger(__name__)
-VMResources = namedtuple("VMResource", ["cpu", "memory", "disk"])
+
+VMResources = namedtuple("VMResources", ["cpu", "memory", "disk"])
```
jdkandersson, line 46: Would it be worth changing this to a `typing.NamedTuple`?

src/runner.py
```diff
@@ -1,53 +1,158 @@
...
     @property
-    def is_idle(self):
-        return self.is_online and self.remote.busy is False
+    def is_idle(self) -> bool:
+        """Whether the runner is idle.
+
+        Returns:
+            bool: Whether the runner is idle.
+        """
+        # Redundant None check for mypy type check.
```
jdkandersson, line 116: Out of curiosity, why is it redundant? the type for `remote` is `remote: Optional[RemoteRunner]` which means it could be `None`?

src/runner.py
```diff
@@ -1,53 +1,158 @@
...
+        return self.remote is not None and self.is_online and self.remote["busy"] is False
```
jdkandersson, line 117: Would it be clearer if these were re-ordered to something like `return self.is_online and self.remote is not None and self.remote["busy"] is False`? It just keeps the `remote` checks together

src/runner.py
```diff
@@ -1,53 +1,158 @@
...
 
     @property
-    def is_offline(self):
-        return self.remote and self.remote.status != "online"
+    def is_offline(self) -> bool:
+        """Whether the runner is offline.
+
+        Returns:
+            bool: Whether the runner is offline.
```
jdkandersson, line 124: Does the retur type have to be provided if we already define the return type as a hint?

src/runner.py
```diff
@@ -1,53 +1,158 @@
...
     @property
-    def virt_type(self):
+    def virt_type(self) -> Optional[str]:
+        """Return the virtualization type of the runner.
+
+        Returns:
+            Optional[str]: Virtualization type of the runner.
+        """
         if self.is_local:
+            # Assert for mypy type check. This check is already done by `is_local`.
```
jdkandersson, line 154: Out of curiosity, what is the `mypy` error if this `assert` is not there?

src/runner.py
```diff
@@ -56,29 +161,49 @@ def virt_type(self):
...
-    pass
+class RunnerStartError(RunnerError):
+    """Error for runner start failure."""
 
 
 class RunnerManager:
+    """Class for querying and controlling the runners.
+
+    Attributes:
+        proxies (Dict[str, str]): Mapping of proxy env vars.
```
jdkandersson, line 187: Would it be worth defining these as typed attributes on the class?

src/charm.py
yhaliaw: The `exception` method automatically logs the trace stack in it context.

For example, running the following script:
```
import logging

logger = logging.getLogger()


if __name__ == "__main__":
    try:
        int("two")
    except Exception as ex:
        logger.exception("Convert error")
```

Outputs:
```
Convert error
Traceback (most recent call last):
  File "/home/user/work/charm/github-runner-operator/test.py", line 8, in <module>
    int("two")
ValueError: invalid literal for int() with base 10: 'two'
```

`Logger.exception` is just `Logger.error` with the arg `exc_info` set to true. If `exc_info` is set to true it will print out the stack trace.

I only learn of this by reading this code base, since it was using `Logger.exception`.

src/charm.py
yhaliaw: The `ActionEvent.set_result` has a input type of `_SerializedData` which is defined as `Dict[str, JsonObject]`.
So we need to import `JsonObject` for a function output.

The problem here is the definition of `JsonObject` it is guarded with `if TYPE_CHECKING:`. So the definition of `JsonObject` only exists when `TYPE_CHECKING` is true, which is controlled by Python.

`if TYPE_CHECKING:` in our code is needed to prevent import `JsonObject` when it is not defined.

In practical term, when running mypy, `TYPE_CHECKING` will equal to true, and `JsonObject` will be imported to satisfy type checking. When running the charm, `TYPE_CHECKING` will equal to false, and a non-existence class will not be import. Removing the `if TYPE_CHECKING` will cause import error when running the charm.

Side note: I have tried setting the type hint of a function return to `Dict[str, Dict[str, int]]`, but mypy does not realize that type can satisfy `Dict[str, JsonObject]`.

src/runner.py
yhaliaw: `src/runner.py:154: error: Item "None" of "Optional[Any]" has no attribute "type"  [union-attr]`

src/charm.py
yhaliaw: `self._stored` => `BoundStoredState`
`self._stored.path` => `ObjectEvents`

src/event_timer.py
yhaliaw: I was think the reasoning should be documented at each location where `subprocess` is used.
If no `subprocess` is used the import will cause linting for unused import.

There is some advantage to this.
- Line comments cause formatter to add an extra black line between imports.
- Each instance of `subprocess` usage should have it own reason. Since usage throughout a file may be different.

src/runner.py
yhaliaw: The `self.is_online` already checks that `self.remote` is not None.

src/runner.py
yhaliaw: After some throught, attributes are meant to be publicly accessible. So would it be better if they are typed in the docstring? So someone using the class only need to read the docstrings and function signature to know how to use the class?

src/charm.py
yhaliaw: It is a type defined by ops framework to represent JSON-line object in Python.
For example, the juju action result could JSON, so the typing for the method to set the action result involves `JsonObject`.

In ops framework:
```
class ActionEvent(EventBase):
...
    def set_results(self, results: '_SerializedData'):
...
```
with:
```_SerializedData = Dict[str, 'JsonObject']```

File: src/charm.py (hunk 23 above)
Comment (jdkandersson) [reply src/charm.py:149]: Didn't know that either, that is indeed quite handy

File: src/runner.py (hunk 39 above)
Comment (jdkandersson) [reply src/runner.py:192]: Does `mypy` read the types from the docstring?

Comment (yhaliaw) [reply src/runner.py:192]: No, but types are inferred from the input of the `__init__` and functions called.

File: charmcraft.yaml (hunk 1 above)
Comment (mthaddon) [reply charmcraft.yaml:16]: Ok, thanks. If you've manually tested it in the meantime I think that's fine for now - the other option would be to keep the default (based on ordering of the series entries in metadata.yaml) as focal for now so we'd ensure it can continue to be deployed on focal until we get the integration tests in place (obviously we can also just choose not to publish to stable until we have those tests in place). But in any case I'll close out this comment thread.

File: .github/workflows/publish_docs.yaml [hunk 40]
```diff
@@ -0,0 +1,20 @@
...
+      - uses: actions/checkout@v3
+      - name: Publish documentation
+        uses: canonical/upload-charm-docs@main
+        id: publishDocumentation
+        with:
+          discourse_host: discourse.charmhub.io
+          discourse_api_username: ${{ secrets.DISCOURSE_API_USERNAME }}
+          discourse_api_key: ${{ secrets.DISCOURSE_API_KEY }}
+      - name: Show index page
+        run: echo '${{ steps.publishDocumentation.outputs.index_url }}'
```
arturo-seijas, line 20: Missing blank line at EOF

.github/workflows/publish_docs.yaml
```diff
@@ -0,0 +1,20 @@
+name: Public Docs
```
arturo-seijas, line 1: Move this file to the [operator workflows repository](https://github.com/canonical/operator-workflows) and refactor the existing ones if needed to resuse it

docs/how-to-guides/contributing.md
```diff
@@ -0,0 +1,69 @@
...
+juju model-config logging-config="<root>=INFO;unit=DEBUG"
+# Configure the machine resource created by the model
+juju set-model-constraints mem=8G cores=2 root-disk=50G
+# Deploy the charm (Assuming you're on amd64)
+juju deploy ./github-runner_ubuntu-22.04-amd64_ubuntu-20.04-amd64.charm
+```
+
+## Canonical Contributor Agreement
+
+Canonical welcomes contributions to the Indico Operator. Please check out our [contributor agreement](https://ubuntu.com/legal/contributors) if you're interested in contributing to the solution.
```
arturo-seijas, line 69: Indico Operator?

metadata.yaml
```diff
@@ -2,17 +2,21 @@
 # See LICENSE file for licensing details.
 name: github-runner
 display-name: GitHub runner
-description: |
+maintainers:
+  - launchpad.net/~canonical-is-devops
+docs: https://discourse.charmhub.io/t/github-runner-documentation-overview/7804
+issues: https://github.com/canonical/github-runner-operator/issues
+summary: >
```
arturo-seijas, line 9: Why the `>`?

metadata.yaml
```diff
@@ -2,17 +2,21 @@
...
-  to ensure that each job runs in a clean environment.
+description: >
+  This charm creates self-hosted GitHub runners. Each unit of this charm will start a configurable
+  number of LXD based containers and virtual machines to host them. Each runner performs only one
+  job, after which it unregisters from GitHub to ensure that each job runs in a clean environment.
 
-  The charm will periodically check the number of idle runners and spawn or destroy
-  runners as necessary to maintain the configured number of runners. Both the reconciliation
-  interval and the number of runners to maintain are configurable.
+  The charm will periodically check the number of idle runners and spawn or destroy runners as
```
arturo-seijas, line 16: "destroy runners" -> "destroy them"

pyproject.toml
```diff
@@ -0,0 +1,46 @@
...
+exclude_dirs = ["/venv/"]
+[tool.bandit.assert_used]
+skips = ["*/*test.py", "*/test_*.py"]
+
+# Testing tools configuration
+[tool.coverage.run]
+branch = true
+
+[tool.coverage.report]
+# TODO: Enable this after migrate unittest to pytest
```
arturo-seijas, line 14: Why? Aren't we running tests right now?

src/charm.py
arturo-seijas: Same here

merkata: TIL, thanks @yhaliaw ! :)

pyproject.toml
yhaliaw: The current test coverage is 39%.
There is a story to rewrite the unit test with pytest and increase the coverage.

Currently also working a story for integration test.

tests/unit/test_runner.py
```diff
@@ -4,6 +4,7 @@
 
 import pytest
 from pylxd.exceptions import LXDAPIException
+
```
arturo-seijas, line 7: The license header is missing in this file

README.md
```diff
@@ -44,4 +44,4 @@ Testing is run via tox and pytest. To run the full test run:
 
     tox
 
-Dependencies are installed in virtual environments. Integration testing requires a juju controller to execute. These tests will use the existing controller, creating an ephemeral model for the tests which is removed after the testing. If you do not already have a controller setup, you can configure a local instance via LXD, see the [upstream documentation][https://juju.is/docs/lxd-cloud] for details.
\ No newline at end of file
+Dependencies are installed in virtual environments. Integration testing requires a juju controller to execute. These tests will use the existing controller, creating an ephemeral model for the tests which is removed after the testing. If you do not already have a controller setup, you can configure a local instance via LXD, see the (upstream documentation)[https://juju.is/docs/lxd-cloud] for details.
```
jdkandersson, line 47: Is the link right? I thought it was `[<title>](<URL>)`

docs/how-to-guides/contributing.md
```diff
@@ -0,0 +1,69 @@
+# Contributing
```
jdkandersson, line 1: Discourse actually has an updated version of this which has some nice changes (e.g., linking to our contributing guide): https://github.com/canonical/discourse-k8s-operator/blob/main/docs/how-to-guides/contributing.md

.github/workflows/publish_docs.yaml
yhaliaw: It turns out the publish charm already includes publish docs as a part of it.
This is file is unneeded.

jdkandersson (approved):
Nice work!

weiiwang01 (approved):

mthaddon (approved):
Looks great, thanks

PR #28, opened 2023-01-10
Refactor runner manager

yhaliaw:
Refactor the logic of the RunnerManager of runner.py into two classes.
Increase reliability and observability of runner-related operations.
Some minor refactor to charm.py for better usability.


src/runner_manager.py
```diff
@@ -0,0 +1,358 @@
...
+    TODO:
+        See if more information should be shared by the charm.
+    """
+
+    name: str
+    online: bool
+    offline: bool
+
+
+class RunnerManager:
```
jdkandersson, line 44: docstring?

src/runner_manager.py
```diff
@@ -0,0 +1,358 @@
...
+            self.proxies["no_proxy"] = no_proxy
+
+        self.session = requests.Session()
+        # TODO: Review the retry strategy
+        adapter = requests.adapters.HTTPAdapter(
+            max_retries=urllib3.Retry(
+                total=10, backoff_factor=0.3, status_forcelist=[500, 502, 503, 504]
+            )
+        )
+        self.session.mount("http://", adapter)
```
jdkandersson, line 87: Should we disable HTTP altogether and only allow HTTPS?

src/runner_manager.py
```diff
@@ -0,0 +1,358 @@
...
+            fastcore.net._opener = opener
+
+        self._github = GhApi(token=token)
+        self._jinja = jinja2.Environment(
+            loader=jinja2.FileSystemLoader("templates"), autoescape=True
+        )
+        self._lxd = pylxd.Client()
+
+        if "/" in path:
+            owner, repo = path.split("/")
```
jdkandersson, line 104: What is the user has put more than 1 `/` in the path? Should we check that input first?

src/runner_manager.py
```diff
@@ -0,0 +1,358 @@
...
+                "install",
+                "-qy",
+                "cpu-checker",
+                "libvirt-clients",
+                "libvirt-daemon-driver-qemu",
+            ],
+        )
+        cls.runner_bin_path.parent.mkdir(parents=True, exist_ok=True)
+
+        logger.info("Finished installing charm dependencies.")
```
jdkandersson, line 140: Are we guaranteed to have successfully installed all of these dependencies?

src/runner_manager.py
```diff
@@ -0,0 +1,358 @@
...
+        if isinstance(self.path, GitHubRepo):
+            runner_bins = self._github.actions.list_runner_applications_for_repo(
+                owner=self.path.owner, repo=self.path.repo
+            )
+        elif isinstance(self.path, GitHubOrg):
+            runner_bins = self._github.actions.list_runner_application_for_org(org=self.path.org)
+
+        logger.debug("Response of runner binary list: %s", runner_bins)
+
+        for bin in runner_bins:
```
jdkandersson, line 161: there is a nice trick using `next`, e.g., `next(binary.download_url for binary in runner_bins if binary.os == os_name and bin.architecture == arch_name)`, you can give a default or handle the exception (maybe `StopIteration`? in case nothing is found. I was also wondering, would there ever be multiple versions available for a combination of os_name and architecture? Or only one?

src/runner_manager.py
```diff
@@ -0,0 +1,358 @@
...
+            download_url (str): URL to download the runner binary.
+
+        TODO:
+            Deletion first cause no runner binary on download failure. See if this cause problems.
+        """
+        logger.info("Downloading runner binary from: %s", download_url)
+
+        # Remove any existing runner bin file
+        if self.runner_bin_path.exists():
+            self.runner_bin_path.unlink()
```
jdkandersson, line 183: Should the unlock happen just after the download? Also, would it be worth creating the file for the new binary first, then unlicking and then rename the file? That way, the time no binary file for the runner is available is minimised. Not sure if it matters, e.g., what would happen if the file is temporarily unavailable

src/runner_manager.py
```diff
@@ -0,0 +1,358 @@
...
+        # Download the new file
+        response = self.session.get(download_url)
```
jdkandersson, line 185: Not sure if this is streamed? Might be worth it to minimise memory depending on the binary size

src/runner_manager.py
```diff
@@ -0,0 +1,358 @@
...
+
+        logger.info("Download of runner binary return status code: %i", response.status_code)
```
jdkandersson, line 187: Should we log the `sha` or some other identifying information as well? Also, should we check the download has not been messed with? Not sure if we can verify it against a signature, for example

src/runner_manager.py
```diff
@@ -0,0 +1,358 @@
...
+            quantity: Number of intended runners.
+            resources: Configuration of the virtual machine resources.
+
+        Returns:
+            Difference between intended runners and actual runners.
+        """
+        runners = self._get_runners()
+
+        # Clean up offline runners
+        offline_runners = [r for r in runners if not r.online]
```
jdkandersson, line 227: Should we use something more descriptive than `r`?

src/runner_manager.py
```diff
@@ -0,0 +1,358 @@
...
+                    self.app_name,
+                    self.path,
+                    self.proxies,
+                    self._generate_runner_name(),
+                )
+                runner.create(
+                    self.image, resources, self.runner_bin_path, self.reconcile_interval, token
+                )
+        elif delta < 0:
+            idle_runners = [r for r in online_runners if not r.busy]
```
jdkandersson, line 262: In what case would the runners be busy and we would have too many runners?

src/runner_manager.py
```diff
@@ -0,0 +1,358 @@
...
+            for runner in offline_runners:
+                runner.remove()
+
+        # Add/Remove runners to match the target quantity
+        online_runners = [r for r in runners if r.exist and r.online]
+        delta = quantity - len(online_runners)
+        if delta > 0:
+            logger.info("Getting registration token for GitHub runners.")
+            if isinstance(self.path, GitHubRepo):
+                token = self._github.actions.create_registration_token_for_repo(
```
jdkandersson, line 241: How fine grained is the creation of this token?

src/runner_manager.py
```diff
@@ -0,0 +1,358 @@
...
+        """Generate a runner name based on charm name.
+
+        Returns:
+            Generated name of runner.
+
+        TODO:
+            Consider name collusion. LXD would fail to create VM on name collusion.
+        """
+        # Generated a suffix for naming propose, not used as secret.
+        suffix = "".join(random.choices(string.ascii_lowercase + string.digits, k=6))  # nosec B311
```
jdkandersson, line 301: what about `uuid`, would that be useful here?

src/utilities.py
```diff
@@ -0,0 +1,113 @@
...
+
+import logging
+import subprocess  # nosec B404
+import time
+from typing import Callable, Optional, Sequence, Type, TypeVar
+
+logger = logging.getLogger(__name__)
+
+# Return type of decorated function
+R = TypeVar("R")
```
jdkandersson, line 14: Would it be worth expanding on the variable name? Maybe `TReturn`? I think there is some convention, perhaps `mypy` enforces it

src/utilities.py
```diff
@@ -0,0 +1,113 @@
...
+    tries: int = 1,
+    delay: float = 0,
+    max_delay: Optional[float] = None,
+    backoff: float = 1,
+    logger: logging.Logger = logger,
+) -> Callable[[Callable[..., R]], Callable[..., R]]:
+    """Parameterize the decorator for adding retry to functions.
+
+    Args:
+        exception: Exception type to be retried. Defaults to Exception.
```
jdkandersson, line 28: Is it worth re-documenting the default value? It is already noted in the function signature?

src/utilities.py
```diff
@@ -0,0 +1,113 @@
...
+        delay: Time in seconds to wait between retry. Defaults to 0.
+        max_delay: Max time in seconds to wait between retry. Defaults to None.
+        backoff: Factor to increase the delay by each retry. Defaults to 1.
+        logger: Logger for logging. Defaults to module logger.
+
+    Returns:
+        The function decorator for retry.
+
+    TODO:
+    For Python 3.10+, PEP 612 allows to specify the parameter of callable using `ParamSpec`. Do
```
jdkandersson, line 39: I think with `typing_extensions` and `from __future__ import annotations` this should already be possible

src/utilities.py
```diff
@@ -0,0 +1,113 @@
...
+
+        Returns:
+            Callable[..., R]: The resulting function with retry added.
+        """
+
+        def fn_with_retry(*args, **kwargs) -> R:
+            """Function with retry"""
+            remain_tries, current_delay = tries, delay
+
+            while True:
```
jdkandersson, line 59: what about `for _ in range(tries)`? That way we avoid accidental infinite loops?

src/utilities.py
```diff
@@ -0,0 +1,113 @@
...
+        """Decorate function with retry.
+
+        Args:
+            fn (Callable[..., R]): The function to decorate.
+
+        Returns:
+            Callable[..., R]: The resulting function with retry added.
+        """
+
+        def fn_with_retry(*args, **kwargs) -> R:
```
jdkandersson, line 55: There is a nice method that helps make sure that the wrapped function still looks like the original function: https://docs.python.org/3/library/functools.html#functools.wraps

src/utilities.py
```diff
@@ -0,0 +1,113 @@
...
+        check: Whether to throw error on non-zero exit code. Defaults to True.
+
+    Returns:
+        Output on stdout.
+
+    TODO:
+        Update `event_timer.py` to use this function.
+    """
+    result = subprocess.run(cmd, capture_output=True)  # nosec B603
+    logger.debug("Command %s returns: %s", " ".join(cmd), result.stdout)
```
jdkandersson, line 99: Should we also log just before executing the command with the command to be executed?

src/utilities.py
```diff
@@ -0,0 +1,113 @@
...
+
+    if check:
+        try:
+            result.check_returncode()
+        except subprocess.CalledProcessError as err:
+            logger.error(
```
jdkandersson, line 105: I think there was a `logger.exception` which automatically writes the call stack to the log as well. Might be worth adding that to the contributing guide

requirements.txt
```diff
@@ -3,3 +3,5 @@ requests
 jinja2
 ghapi
 pylxd
+# Newer version does not work with default OpenSSL version on focal.
```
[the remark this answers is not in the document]
yhaliaw: Cyptography is a dependency for pylxd.

As mentioned before regarding the pylxd problem. I am thinking about remove the dependency on pylxd.

Weii has point out that craft providers does not use pylxd:
https://github.com/canonical/craft-providers/blob/main/craft_providers/lxd/lxc.py

src/github_type.py
```diff
@@ -0,0 +1,46 @@
+# Copyright 2022 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+"""Type of returned data from GitHub web API."""
+
+
+from typing import List, TypedDict
```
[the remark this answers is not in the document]
yhaliaw: Python 3.8 is targeted.

src/runner.py
```diff
@@ -88,596 +47,416 @@ class LxdInstanceConfig(TypedDict):
...
-            remote: GitHub API information of runner.
+            exist: Whether the runner instance exists on LXD. Defaults to False.
+            online: Whether GitHub marks this runner as online. Defaults to False.
+            busy: Whether GitHub marks this runner as busy. Defaults to False.
+            instance: LXD instance of the runner if already created. Defaults to None.
         """
-        self.name = name
-        self.local = local
-        self.remote = remote
+        # Dependency injection to share the instances across different `Runner` instance.
```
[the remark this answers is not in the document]
yhaliaw: Concern over resource usages.

I have not look into the implementation, so it might not be an issue. However, it might be possible that sharing the same instances across multiple runners would allow for reusing TCP connection or websockets, etc. 

Depending on how it is implemented it is still possible for it to share these underlying resource even if we create a `pylxd.Client`, `GhApi`, etc. per runner.

src/github_type.py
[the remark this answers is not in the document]
jdkandersson: Ah, so with the `from __future__ import annotations` you should be able to drop the `List` import and just use the inbuilt `list` for type hints

src/github_type.py
```diff
@@ -0,0 +1,46 @@
...
+
+
+class RunnerApplication(TypedDict, total=False):
+    """Information on a single runner application."""
+
+    os: str
+    architecture: str
+    download_url: str
+    filename: str
+    sha256_checksum: str  # NotRequired
```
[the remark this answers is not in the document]
yhaliaw: I am currently trying out the `NotRequired` from: https://peps.python.org/pep-0655/.

Imported it from `typing_extensions`, since the lib is already used elsewhere.

src/runner.py
```diff
@@ -88,596 +47,416 @@ class LxdInstanceConfig(TypedDict):
...
+        if isinstance(self.path, GitHubRepo):
+            self._github.actions.delete_self_hosted_runner_from_repo(
+                owner=self.path.owner, repo=self.path.repo, runner_id=self.name
+            )
+        elif isinstance(self.path, GitHubOrg):
+            self._github.actions.delete_self_hosted_runner_from_org(
+                org=self.path.org, runner_id=self.name
+            )
 
+        if self.instance is None:
```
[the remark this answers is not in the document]
yhaliaw: It is possible for a runner to be still registered on GitHub but the LXD instance is gone.
In which case, the `remove` method should delete the runner on GitHub and does nothing else.

src/runner_manager.py
yhaliaw: It will solve the problem, but the name would be really long (and may cause problems).

src/runner_manager.py
yhaliaw: The created token should expires in a hour and is used for registering runners:
https://docs.github.com/en/rest/actions/self-hosted-runners?apiVersion=2022-11-28#create-a-registration-token-for-an-organization

src/runner_manager.py
yhaliaw: busy means GitHub has assigned a job to that runner.

Changes to the configuration on the number of runners while some existing runners are executing job can result in above scenario.

src/runner_manager.py
yhaliaw: This was my initial thought, but then I realized if there is a updated version of the binary, and the charm fails to download it. Should the charm use the old version? or wait until it can download the newest version? 
I am unsure as this might have security implications (e.g., security bug fixes in new version).

src/runner_manager.py
yhaliaw: I checked API endpoint again.
It seems to only return one version of os/arch, and it is returning the `sha256_checksum` (I thought it did not).
Although these two are not stated by the API. 
`sha256_checksum` is a optional field in the API, and nothing about one version per os/arch is mentioned.

src/runner_manager.py
yhaliaw: Failure to install should result in an exception due to the command returning non-zero exit code.

I have move this to charm.py and will handle the exception there.

src/runner_manager.py
yhaliaw: It seems there is not a configuration that disables HTTP for requests.

We need to ensure HTTPS is used when we query stuff from the web.

src/runner_manager.py
jdkandersson: Yes that is a good point, that was one of the security feedback items to harden the update process. Maybe it would be better to delete the binary and stop accepting runs as soon as a new binary is released? And log/ alert it on download failure until manual intervention?

src/runner_manager.py
jdkandersson: it should be ok regarding the os/arch, I'm guessing the API just returns the latest release. I was thinking it might return binaries from earlier releases which is why I asked the question

src/runner_manager.py
jdkandersson: What if we mount `None` instead of `adapter` on `http://`?

src/runner_manager.py
jdkandersson: Ah I see, makes sense

src/runner_manager.py
jdkandersson: And do you know if the token is the minimally possible token for the runners?

src/runner_manager.py
jdkandersson: What kind of problems?

src/runner.py
```diff
@@ -88,596 +47,416 @@ class LxdInstanceConfig(TypedDict):
...
+                # this is just a fall-back.
+                pass
+        else:
+            # We somehow have a non-running instance which should have been
+            # ephemeral. Try to delete it and allow any errors doing so to
+            # surface.
+            try:
+                self.instance.delete(wait=True)
+            except Exception as err:
+                raise RunnerRemoveError() from err
```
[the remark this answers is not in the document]
gregory-schiano: Has it been change ? I don't see any message still

src/runner.py
```diff
@@ -88,596 +47,416 @@ class LxdInstanceConfig(TypedDict):
...
-    # TODO: Verify if this violates bandit B108.
-    runner_bin_path = Path("/var/cache/github-runner-operator/runner.tgz")
-    env_file = Path("/opt/github-runner/.env")
+        instance = self._lxd.instances.create(config=instance_config, wait=True)
+        self.exist = True
+        return instance
 
-    def __init__(self, path: str, token: str, app_name: str, reconcile_interval: int) -> None:
-        """Construct RunnerManager object for creating and managing runners.
+    @retry(tries=10, delay=1, logger=logger)
```
gregory-schiano, line 228: We also have here embedded retries, possible to refactor to avoid this situation ?

src/runner.py
```diff
@@ -88,596 +47,416 @@ class LxdInstanceConfig(TypedDict):
...
-        )
-        cls.runner_bin_path.parent.mkdir(parents=True, exist_ok=True)
-
-    def get_latest_runner_bin_url(self) -> Optional[str]:
-        """Get the URL for the latest runner binary.
+        if not self._lxd.profiles.exists("runner"):
+            logger.info("Creating runner LXD profile")
+            profile_config = {
+                "security.nesting": "true",
+                "security.privileged": "true",
```
gregory-schiano, line 239: Aren't nesting and privileged risky ? I can see some possible risks possibility with a nested instance (like bypassing potential apparmor restrictions, access to /proc and /sys)

src/runner.py
```diff
@@ -88,596 +47,416 @@ class LxdInstanceConfig(TypedDict):
...
-            runner_bins = self._api.actions.list_runner_applications_for_org(org=self.path)
-        for runner_bin in runner_bins:
-            if runner_bin.os == os_name and runner_bin.architecture == arch_name:
-                return runner_bin.download_url
-        return None
+            logger.info("Found existing runner LXD profile")
 
-    def update_runner_bin(self, download_url: str) -> None:
-        """Download a runner file, replacing the current copy.
+    @retry(tries=10, delay=1, logger=logger)
```
gregory-schiano, line 249: Same comment concerning retry

src/runner.py
```diff
@@ -88,596 +47,416 @@ class LxdInstanceConfig(TypedDict):
...
+        wait_interval = 15  # seconds
+        attempts = int(reconcile_interval * 60 / wait_interval)
 
-        self._clean_unused_profiles()
+        @retry(exception=RunnerExecutionError, tries=attempts, delay=wait_interval, logger=logger)
+        def connect_test():
+            """Check able to connect to network."""
+            try:
+                self._execute(["/usr/bin/who"])
+                # TODO Better network test? Ping `github.com` too much results in throttling.
```
gregory-schiano, line 318: Perhaps a DNS call ? like nslookup

src/runner.py
```diff
@@ -88,596 +47,416 @@ class LxdInstanceConfig(TypedDict):
...
-        runners = [r for r in self.get_info() if r.is_local]
-        runner_names = ", ".join(r.name for r in runners)
-        logger.info(f"Removing existing local runners: {runner_names}")
-        for runner in runners:
-            self._remove_runner(runner)
+        wait_interval = 15  # seconds
+        attempts = int(reconcile_interval * 60 / wait_interval)
 
-        self._clean_unused_profiles()
+        @retry(exception=RunnerExecutionError, tries=attempts, delay=wait_interval, logger=logger)
```
gregory-schiano, line 313: We're also nesting retries here + a sleep 60
The overall process could take a lot of time

src/runner.py
```diff
@@ -88,596 +47,416 @@ class LxdInstanceConfig(TypedDict):
...
 
-    def _remove_runner(self, runner: RunnerInfo) -> None:
-        """Remove a runner.
+    @retry(tries=10, delay=1, logger=logger)
+    def _install_binary(self, binary: Path) -> None:
+        """Load GitHub self-hosted runner binary on to the runner instance.
 
         Args:
-            runner: Information on the runner to remove.
+            binary: Path the compressed runner binary.
```
gregory-schiano, line 338: I think it's missing a `to`
`Path to the compressed runner's binary`

src/runner.py
```diff
@@ -88,596 +47,416 @@ class LxdInstanceConfig(TypedDict):
...
-        ]
-        self._check_output(instance, cmd)
-
-    def _start_runner(self, instance: pylxd.models.Instance) -> None:
-        """Start a runner that is already registered.
+        # Creating directory and putting the file are idempotent, and can be retried.
+        self.instance.files.mk_dir(str(self.runner_application))
+        self.instance.files.put(binary_path, binary.read_bytes())
+        self._execute(["tar", "-xzf", binary_path, "-C", str(self.runner_application)])
+        self._execute(["chown", "-R", "ubuntu:ubuntu", str(self.runner_application)])
```
gregory-schiano, line 356: Is it space-safe ?
A space in "runner_application" could lead to permission change on sensitive files (especially if it's running privileged)

src/runner.py
```diff
@@ -88,596 +47,416 @@ class LxdInstanceConfig(TypedDict):
...
 
-    Attributes:
-        proxies (Dict[str, str]): Mapping of proxy env vars.
-        path (str): GitHub repository path in the format '<org>/<repo>', or the GitHub org name.
-        app_name (str): An name for the set of runners.
-        reconcile_interval (int): Number of minutes between each reconciliation of runners state.
-    """
+        # Create runner instance.
+        instance_config: LxdInstanceConfig = {
+            "name": self.name,
```
gregory-schiano, line 211: No risk of instance already existing ?

src/runner.py
```diff
@@ -88,596 +47,416 @@ class LxdInstanceConfig(TypedDict):
...
-        """
-        script_contents = self._jinja.get_template("start.j2").render()
-        instance.files.put("/opt/github-runner/start.sh", script_contents, mode="0755")
-        self._check_output(
-            instance, "sudo chown ubuntu:ubuntu /opt/github-runner/start.sh".split()
-        )
-        self._check_output(instance, "sudo chmod u+x /opt/github-runner/start.sh".split())
-        self._check_output(instance, "sudo -u ubuntu /opt/github-runner/start.sh".split())
+        # Verify the env file is written to runner.
+        exit_code, _, _ = self.instance.execute(["test", "-f", str(self.config_script)])
```
gregory-schiano, line 359: No risk of injecting another command here using `&& other_command` ?

src/runner.py
```diff
@@ -88,596 +47,416 @@ class LxdInstanceConfig(TypedDict):
...
-            f"{','.join(labels)}",
-        ]
-        self._check_output(instance, cmd)
-
-    def _start_runner(self, instance: pylxd.models.Instance) -> None:
-        """Start a runner that is already registered.
+        # Creating directory and putting the file are idempotent, and can be retried.
+        self.instance.files.mk_dir(str(self.runner_application))
+        self.instance.files.put(binary_path, binary.read_bytes())
+        self._execute(["tar", "-xzf", binary_path, "-C", str(self.runner_application)])
```
gregory-schiano, line 355: No risk of injecting another command in runner_application using `&& runner_application`

gregory-schiano (commented):
Since I'm commenting while reading I took a pretty defensive approach in my comment since I don't have yet a global overview of which input can be trusted and which one can't, so some comments could be non-relevant

After a full first pass, I'll have a better view and could close myself some comments. But I can't finalize the review for today, will get back to it tomorrow

src/charm.py
```diff
@@ -233,17 +247,17 @@ def _on_check_runners_action(self, event: ActionEvent) -> None:
...
         except Exception as e:
             logger.exception("Failed to get runner info")
             event.fail(f"Failed to get runner info: {e}")
             return
 
         for runner in runner_info:
-            if runner.is_online:
+            if runner.online:
                 online += 1
                 runner_names.append(runner.name)
```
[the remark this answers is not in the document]
yhaliaw: The runner names should be unique.

This is enforce by LXD and GitHub.
This should not be any issue in the future, as UUID is planned to be used to generate runner names.

src/runner.py
```diff
@@ -88,596 +47,416 @@ class LxdInstanceConfig(TypedDict):
...
 
+        logger.info("Removing LXD instance for runner: %s", self.name)
 
-class RunnerError(Exception):
-    """Generic runner error."""
+        if isinstance(self.path, GitHubRepo):
+            self._github.actions.delete_self_hosted_runner_from_repo(
+                owner=self.path.owner, repo=self.path.repo, runner_id=self.name
+            )
+        elif isinstance(self.path, GitHubOrg):
```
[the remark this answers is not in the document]
yhaliaw: I wanted to use match statement, but with python 3.8 was unable to.
Do you think explicitly name the case is better like now?
or just convert one to the default case with `else`?

src/runner.py
yhaliaw: The path are hardcoded, and not dependent on user input.

src/runner.py
yhaliaw: The instance name will be generated with UUID v4, so this should not be an issue.

src/runner.py
yhaliaw: The `self.config_script` is a hardcoded value, user input not does not affect it.

src/runner.py
yhaliaw: The `self.runner_application` and `binary_path` are both hardcoded values.

src/runner.py
```diff
@@ -88,596 +47,416 @@ class LxdInstanceConfig(TypedDict):
...
-            return self.local.type
-        elif self.remote:
-            remote_virt_type = (
-                label["name"]
-                for label in self.remote["labels"]
-                if label["name"] in {"container", "virtual-machine"}
-            )
-            return next(remote_virt_type, None)
-        return None
+        @retry(tries=10, delay=10, logger=logger)
```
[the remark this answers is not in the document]
yhaliaw: All the retry within retry are for function that are waiting for something to succeed before moving on.
E.g., booting up an instance.

After some thought, rather than retry within retry it should be in sequence.
E.g., booting up with retries, then verify successfully boot up with retries.
So stacking retries are not an problem.

src/runner.py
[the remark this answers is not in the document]
gregory-schiano: Or simply doing 2 `if`

src/runner.py
```diff
@@ -88,596 +47,416 @@ class LxdInstanceConfig(TypedDict):
...
+            )
 
+        if self.instance is None:
+            return
 
-class RunnerCreateError(RunnerError):
-    """Error for runner creation failure."""
+        if self.instance.status == "Running":
+            self.instance.stop(wait=True)
+            time.sleep(30)
```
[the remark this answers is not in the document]
yhaliaw: Removed

src/runner.py
yhaliaw: I am unsure if these are needed.
This profile setting was there before.
We are request security team for a code review (in addition to the report they already produced) and they may help us understand the implication of this more.

src/charm.py
```diff
@@ -96,8 +95,15 @@ def _get_runner_manager(
             token = self.config["token"]
         if path is None:
             path = self.config["path"]
+
         if not token or not path:
             return None
+
+        if "/" in path:
```
jdkandersson, line 102: Should we check whether there are multiple "/" and raise/ return an error if that is the case? Is more than 1 slash an error?

src/runner.py
```diff
@@ -115,50 +131,56 @@ def create(
...
         binary: Path,
-        reconcile_interval: int,
         registration_token: str,
     ):
+        """Create the runner instance on LXD and register it on GitHub.
+
+        Args:
+            image: Name of the image to launch the LXD instance with.
+            resources: Resource setting for the LXD instance.
+            binary: Path to the runner binary.
```
jdkandersson, line 141: Would `binary_path` be a batter name for the variable?

src/runner.py
```diff
@@ -115,50 +131,56 @@ def create(
...
+            registration_token: Token to register the runner on GitHub.
+
+        Raises:
+            RunnerCreateError: _description_
```
jdkandersson, line 145: Should this description be populated?

src/runner.py
```diff
@@ -115,50 +131,56 @@ def create(
...
+        """
         logger.info("Creating runner: %s", self.name)
 
         self.instance = self._create_instance(image, resources)
 
         try:
-            self._start_instance(reconcile_interval)
+            self._start_instance()
+            # Wait some initial time for the instance to boot up
+            time.sleep(60)
```
jdkandersson, line 154: Is there a way to check whether it is done rather than wait a set amount of time? That way it would be possible to continue earlier?

src/runner.py
```diff
@@ -115,50 +131,56 @@ def create(
...
             self._register_runner(registration_token, labels=[self.app_name, image])
             self._start_runner()
         except Exception as err:
             self.instance.stop(wait=True)
-            try:
-                self.instance.delete(wait=True)
-            except Exception:  # nosec B110
-                # this is just a fall-back.
+
+            with suppress(Exception):
```
jdkandersson, line 163: Can we make this more explicit in which exception is being suppressed?

src/runner.py
```diff
@@ -167,27 +189,26 @@ def check_shutdown():
...
-
             try:
-                self.instance.delete(wait=True)
-            except Exception:  # nosec B110
+                self.instance.stop(wait=True, timeout=60)
+            except pylxd.exceptions.LXDAPIException as err:
+                logger.exception("Unable to gracefully stop runner within timeout.")
+                self.instance.stop(force=True)
+
+            with suppress(Exception):
```
jdkandersson, line 198: Similar comment as above, could we pick a more specific exception here?

jdkandersson (commented):
Not finished the review yet, will need to continue tomorrow

src/runner.py
yhaliaw: The next method called `_wait_boot_up` checks every 30 seconds if the instance completes the boot up.
This is just the initial wait prior to the first check.

src/charm.py
```diff
@@ -233,17 +247,17 @@ def _on_check_runners_action(self, event: ActionEvent) -> None:
         runner_names = []
 
         try:
-            runner_info = runner_manager.get_info()
+            runner_info = runner_manager.get_github_info()
         except Exception as e:
```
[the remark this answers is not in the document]
yhaliaw: This is catching any error and converting into a return value for the action.
Should all type of exception be catch here or still needs to be more focused.

src/runner.py
gregory-schiano: Well if that's not needed I'd suggest to remove them or at least the "privileged"

src/utilities.py
```diff
@@ -0,0 +1,120 @@
...
+logger = logging.getLogger(__name__)
+
+
+# Parameters of the function decorated with retry
+TParam = ParamSpec("TParam")
+# Return type of the function decorated with retry
+TReturn = TypeVar("TReturn")
+
+
+def retry(
```
gregory-schiano, line 23: Isn't there a lib doing this already ?

gregory-schiano (commented):
Another pass of review in the code
I'll have a second look at runner.py later

src/runner_manager.py
```diff
@@ -0,0 +1,360 @@
...
+
+    Used as a returned type to method querying runner information.
+
+    TODO:
+        See if more information should be shared by the charm.
+    """
+
+    name: str
+    online: bool
+    offline: bool
```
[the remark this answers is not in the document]
yhaliaw: These variable are set according to the `status` of runner returned by the API.

https://docs.github.com/en/rest/actions/self-hosted-runners?apiVersion=2022-11-28#list-self-hosted-runners-for-a-repository

The schema only mention this:
```
...
          "status": {
            "description": "The status of the runner.",
            "type": "string",
            "examples": [
              "online"
            ]
          },
...
```
The old version only checks for `status == "online"` and `status == "offline"` with the method `is_online` and `is_offline`, and I am unsure if there are any other status.

That said I should have used enum for this. Will be changing it.

src/runner_manager.py
```diff
@@ -0,0 +1,360 @@
...
+        else:
+            logger.warning(
+                "Checksum for runner binary is not found, verification of download is not done."
+            )
+
+        with self.runner_bin_path.open(mode="wb") as runner_bin_file:
+            runner_bin_file.write(response.content)
+
+        # Verify the file integrity.
+        with tarfile.open(self.runner_bin_path, "r:gz") as f:
```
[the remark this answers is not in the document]
yhaliaw: I think `is_tarfile` only checks whether it is a tar file.
The above line also tries to decompress with `gzip`.

src/charm.py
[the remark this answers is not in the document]
jdkandersson: Any thoughts @gregory-schiano? I think the reason is because juju expect charms not to raise exceptions, and if an exception is raised the charm will go into error state. Generally we shouldn't handle all exceptions and be more focussed, I can see why it might be ok in this case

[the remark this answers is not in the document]
jdkandersson: Do you think it is worth including a note explaining why general exceptions are caught in this case for future readers?

src/runner_manager.py
jdkandersson: I'm guessing it would refuse the run actions at that point, the theory is that there may be a security vulnerability with the runner binary we have given there is a new one, so the trade-off here is security versus usability. E.g., the more secure way is to stop using the current binary as soon as we become aware of a new binary. The usability perspective is to keep using a potentially vulnerable binary until we can successfully download the new binary

src/runner.py
jdkandersson: For my understanding, what is the reason for needing the initial wait period if the following command already has a variable wait limit in it?

yhaliaw: The only reason is due to the first check is pretty much an automatic failure, since virtual machine cannot boot up that fast.

src/runner_manager.py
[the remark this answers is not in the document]
gregory-schiano: Well to check if the tar is valid, it needs to uncompress it (or at least the header) as the Gzip compression is on top of the tar archiving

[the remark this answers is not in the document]
yhaliaw: Ok, I made the change to `is_tarfile`.

src/runner_manager.py
```diff
@@ -0,0 +1,360 @@
...
+                    hash,
+                )
+                raise RunnerBinaryError("Checksum mismatch for downloaded runner binary")
+        else:
+            logger.warning(
+                "Checksum for runner binary is not found, verification of download is not done."
+            )
+
+        with self.runner_bin_path.open(mode="wb") as runner_bin_file:
+            runner_bin_file.write(response.content)
```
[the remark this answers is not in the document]
yhaliaw: Done.

src/runner_manager.py
```diff
@@ -0,0 +1,360 @@
...
+                    resources,
+                    self.runner_bin_path,
+                    token.token,
+                )
+        elif delta < 0:
+            idle_runners = [runner for runner in online_runners if not runner.busy]
+            offset = min(-delta, len(idle_runners))
+            if offset != 0:
+                logger.info("Removing %i runner(s)", offset)
+                remove_runners = idle_runners[:offset]
```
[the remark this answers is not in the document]
yhaliaw: You are right this logic can cause removal of a busy runner.

The way to resolve this will involve the GitHub runner binary having an atomic operation that checking if the runner is currently busy and remove/unregister from GitHub if not busy.

There is two way to remove/unregister a runner, one method is executing the runner binary, the other is calling a RESTful API on GitHub. From the [documentation](https://docs.github.com/en/actions/hosting-your-own-runners/removing-self-hosted-runners#removing-a-runner-from-a-repository) is it unclear how these two method interacts with the runner being busy. I will need to do more research on this.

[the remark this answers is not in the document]
yhaliaw: I think this is not a high priority, as this is only possible due to configuration change on the number of the runners to maintain, which should occur infrequently.

src/runner_manager.py
```diff
@@ -0,0 +1,360 @@
...
+    def _get_runner_github_info(self) -> Dict[str, SelfHostedRunner]:
+        remote_runners_list: list[SelfHostedRunner] = []
+        if isinstance(self.path, GitHubRepo):
+            remote_runners_list = self._github.actions.list_self_hosted_runners_for_repo(
+                owner=self.path.owner, repo=self.path.repo
+            )["runners"]
+        if isinstance(self.path, GitHubOrg):
+            remote_runners_list = self._github.actions.list_self_hosted_runners_for_org(
+                org=self.path.org
+            )["runners"]
```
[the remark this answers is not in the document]
yhaliaw: It should return a empty list, which is the current observed behavior.

Sample response:
```
{
  "total_count": 0,
  "runners": [

  ]
}
```

According to the Response schema, the `runners` is a required field.
https://docs.github.com/en/rest/actions/self-hosted-runners?apiVersion=2022-11-28#list-self-hosted-runners-for-a-repository

src/charm.py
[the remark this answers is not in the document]
gregory-schiano: That's a good question, looking at the SDK I'd reach the same conclusion the Action handling doesn't expect exception and manages it through the `event.fail`
I don't see any mention of how exception should be managed in https://juju.is/docs/sdk/styleguide
And lastly I don't know how Action events will manage exceptions (will it really put the charm as blocked ?)

I'll ask on the juju or ops channel

src/runner_manager.py
gregory-schiano: As long as we accept that in the exceptional situation where there is a security vuln AND the new version is non downloadable (either due to an incident on our side or due to an issue on the site hosting the new version), the runners aren't available
Then it's fine

src/runner_manager.py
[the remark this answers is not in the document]
gregory-schiano: I agree, perhaps adding a TODO comment here, and/or creating a ticket on Jira to manage this case

src/runner_manager.py
```diff
@@ -0,0 +1,391 @@
...
+        is done to prevent security issues arising from outdated runner binary
+        containing security flaws. The newest version of runner binary should
+        always be used.
+
+        Args:
+            binary: Information on the runner binary to download.
+
+        TODO:
+            Convert to download of runner binary to streaming to file, rather than save in memory
+            then copy to file. Ignore if the file size is too small.
```
gregory-schiano, line 169: Is it still relevant with the changes for hash computation ?

src/runner_manager.py
```diff
@@ -0,0 +1,391 @@
...
+                if binary.sha256_checksum:
+                    hash.update(chunk)
+
+        logger.info("Finished download of runner binary.")
+
+        # Verify the checksum if checksum is present.
+        if binary.sha256_checksum:
+            if binary.sha256_checksum != hash.hexdigest():
+                logger.error(
+                    "Mismatch of excepted hash of runner binary (%s) and calculated hash (%s)",
```
gregory-schiano, line 203: To avoid the repetition of `of` I propose (and there's a typo for expected)
`The expected hash of runner binary (%s) doesn't match the calculated hash (%s)`

src/runner_manager.py
```diff
@@ -0,0 +1,391 @@
...
+                    binary.sha256_checksum,
+                    hash,
+                )
+                raise RunnerBinaryError("Checksum mismatch for downloaded runner binary")
+        else:
+            logger.warning("Checksum for runner binary is not found, download not verified.")
```
gregory-schiano, line 209: Do we allow this case ? The absence of hash could be the indication of some security issue like corrupted source

src/runner.py
```diff
@@ -85,599 +56,445 @@ class LxdInstanceConfig(TypedDict):
...
+            try:
+                self.instance.delete(wait=True)
+            except pylxd.exceptions.LXDAPIException as err:
+                raise RunnerRemoveError(f"Unable to remove {self.name}") from err
 
+    @retry(tries=5, delay=1, logger=logger)
+    def _create_instance(
+        self, image: str, resources: VirtualMachineResources, ephemeral: bool = True
+    ) -> pylxd.models.Instance:
+        """Create a instance of runner.
```
gregory-schiano, line 217: Change `a` to `an`

src/runner.py
```diff
@@ -85,599 +56,445 @@ class LxdInstanceConfig(TypedDict):
...
 
-class RunnerStartError(RunnerError):
-    """Error for runner start failure."""
+        Args:
+            image: Image to launch the instance hosting the runner.
```
gregory-schiano, line 220: Change to `Image used to launch the instance hosting the runner`

src/utilities.py
```diff
@@ -0,0 +1,135 @@
...
+        check: Whether to throw error on non-zero exit code.
+
+    Returns:
+        Output on stdout.
+
+    TODO:
+        Update `event_timer.py` to use this function.
+    """
+    logger.info("Executing command %s", cmd)
+    result = subprocess.run(cmd, capture_output=True)  # nosec B603
```
gregory-schiano, line 106: We should probably harden this `subprocess.run` which could be risky with injecting unexpected commands due to non validated input.
Best would be to set `shell=False` but this could be sometimes difficult to set
Otherwise using `shlex.split` which will ensure the params are properly quoted

gregory-schiano (commented):
Here's my second pass, we're almost there :)

src/charm.py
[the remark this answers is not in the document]
gregory-schiano: Feedback I got:
```
ppasotti
[2:45 PM]
The framework doesn't do much in terms of exception management. All uncaught exceptions will put the charm in error state until the same hook can be executed without raising an exception (juju will keep trying even if more events are queued) 
[2:45 PM]
action events aren't different
[2:46 PM]
so in general, a charm should never let an exception raise. because it's hard to recover from that and the user feedback is poor. Event.fail seems appropriate
```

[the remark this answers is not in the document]
yhaliaw: I will write some error/exception handling guideline on [is-charms-contributing-guide](https://github.com/canonical/is-charms-contributing-guide). So the team can review it.

src/runner_manager.py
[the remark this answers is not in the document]
jdkandersson: I'm not sure pylint will let you have TODO comments...

src/utilities.py
yhaliaw: The `subprocss.run` defaults to `shell=False`. So the code should run fine with explicit `shell=False` which I added.

src/charm.py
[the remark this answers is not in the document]
gregory-schiano: Yeap perfect, @jdkandersson I think we can resolve this discussion ?

src/runner.py
```diff
@@ -88,596 +47,416 @@ class LxdInstanceConfig(TypedDict):
...
-            Whether the runner is online.
-        """
-        return self.remote is not None and self.remote["status"] == "online"
+        try:
+            self._start_instance(reconcile_interval)
+            self._install_binary(binary)
+            self._configure_runner()
+            self._register_runner(registration_token, labels=[self.app_name, image])
+            self._start_runner()
+        except Exception as err:
```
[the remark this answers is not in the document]
gregory-schiano: Same as above I think we can resolve this discussion ?

src/runner_manager.py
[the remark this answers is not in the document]
gregory-schiano: @jdkandersson what do you think we should do here ?

src/utilities.py
yhaliaw: I had a discussion with @jdkandersson on this.
The decision was since this function was not too large, the benefit of using a lib does not outweigh the security risks.
Since lib such as https://github.com/jd/tenacity are owned under the author not an org.

src/runner_manager.py
```diff
@@ -0,0 +1,360 @@
...
+                online,
+                busy,
+                local_runner,
+            )
+
+        remote_runners = self._get_runner_github_info()
+        local_runners = {
+            instance.name: instance
+            for instance in self._lxd.instances.all()
+            if instance.name.startswith(f"{self.app_name}-")
```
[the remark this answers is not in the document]
yhaliaw: Name collusion on the level of app_name is an issue right now.
I have created a story on jira to address it in the future.

Individual LXD instance does not have this problem as UUID v4 is used as part of the name.

yhaliaw:
There is a story to refine unit test and update to use `pytest` in the future.

src/runner_manager.py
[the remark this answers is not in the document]
jdkandersson: Maybe just create an issue in the repo, that way we have an easy way of telling all the problems we should fix

jdkandersson (commented):
Haven't finished the review yet, just leaving the WIP so far

src/runner_manager.py
```diff
@@ -140,66 +144,75 @@ def get_latest_runner_bin_url(
...
-            logger.warning(
-                "Checksum for runner binary is not found, verification of download is not done."
-            )
+        if not binary["sha256_checksum"]:
+            logger.error("Checksum for runner binary is not found, unable to verify download.")
+            raise RunnerBinaryError("Checksum for runner binary is not found in GitHub response.")
+
+        hash = hashlib.sha256()
+
+        with tempfile.NamedTemporaryFile(delete=False) as tmp_file:
```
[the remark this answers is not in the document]
yhaliaw: Yes, the temporary file is not deleted after the context manager exits.

This is needed as the file (runner binary) needs to be stored for starting new runners.
The file is only deleted when a new version of the runner binary is available.

The `tempfile` lib is used for security reasons. 
Bandit will lint hardcoded tmp directory and recommend using `tempfile` for this.
https://bandit.readthedocs.io/en/latest/plugins/b108_hardcoded_tmp_directory.html
https://security.openstack.org/guidelines/dg_using-temporary-files-securely.html

src/runner_manager.py
yhaliaw: The registration token should only have access to register runners.
There is no setting related to access scope.

This is not the personal access tokens, which can be scoped to limit access.

src/runner_manager.py
```diff
@@ -0,0 +1,360 @@
...
+                config = RunnerConfig(
+                    self.app_name, self.path, self.proxies, self._generate_runner_name()
+                )
+                runner = Runner(self._github, self._jinja, self._lxd, config)
+                runner.create(
+                    self.image,
+                    resources,
+                    self.runner_bin_path,
+                    token.token,
+                )
```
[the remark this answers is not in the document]
yhaliaw: I created a issue to add concurrency to runner creation.

src/runner_manager.py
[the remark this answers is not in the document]
yhaliaw: I created an issue.

src/runner_manager.py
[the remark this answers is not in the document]
jdkandersson: Makes sense, is it worth it to add a comment documenting that?

src/charm.py
```diff
@@ -38,20 +43,19 @@ class UpdateRunnerBinEvent(EventBase):
...
     TODO:
-    * Remove support of LXD containers, due to security concerns.
     * Review the action results fields, and use TypedDict for the action results, and JSON-like
         return values.
     """
 
     _stored = StoredState()
 
-    def __init__(self, *args) -> None:
+    def __init__(self, *args, **kargs) -> None:
```
jdkandersson, line 56: I'm surprised that `flake8-docstrings-complete` is not complaining about the undocumented arguments here?

src/charm.py
```diff
@@ -91,48 +104,82 @@ def _get_runner_manager(
...
             path = self.config["path"]
+
         if not token or not path:
             return None
-        return RunnerManager(path, token, self.app.name, self.config["reconcile-interval"])
 
-    def _on_install(self, event: InstallEvent) -> None:
+        if "/" in path:
+            paths = path.split("/")
+            if len(paths) == 2:
```
jdkandersson, line 113: There is an opportunity to reduce nesting by:

```python
if len(paths) != 2:
    logger.error("Invalid path %s", path)
    return None

... # Happy path execution continue
```

src/charm.py
```diff
@@ -150,14 +197,18 @@ def _on_config_changed(self, event: ConfigChangedEvent) -> None:
             self.unit.status = BlockedStatus(
                 f"Failed to start timer for regular reconciliation and binary update checks: {ex}"
             )
+        # Safe guard against unexpected error.
+        except Exception as err:  # pylint: disable=broad-exception-caught
```
jdkandersson, line 201: Do you think it might be worth making this a decorator? We seem to repeat this `Exception` catching in a few functions: https://peps.python.org/pep-0318/

jdkandersson: Based on complexity, this could be a follow up issue, up to you

src/charm.py
```diff
@@ -328,30 +394,46 @@ def _reconcile_runners(self, runner_manager: RunnerManager) -> Dict[str, "JsonOb
...
+
+    @staticmethod
+    @retry(tries=10, delay=15, max_delay=60, backoff=1.5)
+    def _install_deps() -> None:
+        """Install dependencies."""
+        logger.info("Installing charm dependencies.")
+
+        # Binding for snap, apt, and lxd init commands are not available so subprocess.run used.
+        execute_command(["/usr/bin/apt", "remove", "-qy", "lxd", "lxd-client"], check=False)
+        execute_command(["/usr/bin/snap", "install", "lxd", "--channel=latest/stable"])
```
jdkandersson, line 418: For any of these commands, is there any issues with running them multiple times? E.g., a later command fails which triggers the retry which then leads to these commands being run again and they fail? Or are they idempotent?

src/event_timer.py
```diff
@@ -74,8 +74,10 @@ def ensure_event_timer(
 
         Raises:
             TimerEnableError: Timer cannot be started. Events will be not emitted.
+
+        TODO:
```
jdkandersson, line 78: Should we add an issue to this repo to convert all the `TODO` items to issues?

tests/unit/test_charm.py
```diff
@@ -1,109 +1,285 @@
...
+        assert harness.charm._get_runner_manager() is not None
+
+        # With invalid path.
+        assert harness.charm._get_runner_manager("mocktoken", "mock/invalid/path") is None
+
+    @patch("charm.RunnerManager")
+    @patch("pathlib.Path.write_text")
+    @patch("subprocess.run")
+    def test_on_install_failure(self, run, wt, rm):
+        """Test various error thrown during install."""
```
jdkandersson, line 164: We have this standard on writing documentation for tests, is it applicable here? https://github.com/canonical/is-charms-contributing-guide/blob/main/CONTRIBUTING.md#test-structure

jdkandersson (commented):
This is close now

src/charm.py
yhaliaw: Since `args` and `kargs` are standard convention, maybe `flake8-docstrings-complete` does not complain about it?

jdkandersson: I wrote the linter 😉  I can see that it is being used, I'll look into why it wasn't caught when I get some time. In the meantime, do you think it is worth documenting them in the docstring?

yhaliaw: Will do

src/runner_manager.py
yhaliaw: If mount `None` on session for `http://`, making requests on `http://` will throw the following error:
```
AttributeError: 'NoneType' object has no attribute 'send'
```
I will modified the code to disable HTTP.

jdkandersson: Maybe this can be a follow up ticket?

tests/unit/test_charm.py
yhaliaw: This file will be rewritten for the migrate to pytest story as charm test are current using `unittest` lib.
I did add some test to make the code coverage pass 85% for now.

jdkandersson: Yeh maybe we leave it given the tests need to be rewritten, good point

src/runner_manager.py
yhaliaw: This will be better as a followup.
The mypy failed for this as the `request` lib is expecting a `BaseAdapter` type rather than None.
The might be a better way to disable HTTP.

src/charm.py
yhaliaw: I reviewed them, and they all seem idempotent.

src/charm.py
yhaliaw: I wrote the decorators.

jdkandersson (dismissed):
Nice work! Thanks for the resilience, this charm is quite complicated

src/runner_manager.py
```diff
@@ -0,0 +1,392 @@
...
+        if not binary["sha256_checksum"]:
+            logger.error("Checksum for runner binary is not found, unable to verify download.")
+            raise RunnerBinaryError("Checksum for runner binary is not found in GitHub response.")
+
+        sha256 = hashlib.sha256()
+
+        # Use tempfile as recommend by bandit to avoid hardcoding the file
+        # path.  The `delete=False` arg save the file permanently for the charm
+        # to use start runners.
+        with tempfile.NamedTemporaryFile(delete=False) as tmp_file:
```
weiiwang01, line 189: You can leave the context manager and use the `tempfile.NamedTemporaryFile` directly if you don't need to delete the temp file. This will reduce nesting layers.

`tmp_file = tempfile.NamedTemporaryFile(delete=False)`

src/runner_manager.py
```diff
@@ -0,0 +1,392 @@
...
+                "Expected hash of runner binary (%s) doesn't match the calculated hash (%s)",
+                binary["sha256_checksum"],
+                sha256,
+            )
+            raise RunnerBinaryError("Checksum mismatch for downloaded runner binary")
+
+        # Verify the file integrity.
+        if not tarfile.is_tarfile(tmp_file.name):
+            logger.error("Failed to decompress downloaded GitHub runner binary.")
+            raise RunnerBinaryError("Downloaded runner binary cannot be decompressed.")
```
weiiwang01, line 209: Using the [`atexit`](https://docs.python.org/3/library/atexit.html) module to register a cleanup function in case of exception?

src/runner_manager.py
yhaliaw: The context manager is for closing the file, else `tmp_file.close()` still needs to be called.

src/runner_manager.py
yhaliaw: There is indeed resource leak (tmp files) on file integrity check failure.
I am working on fixing it.

src/runner.py
```diff
@@ -85,599 +56,438 @@ class LxdInstanceConfig(TypedDict):
...
-    def __init__(self, path: str, token: str, app_name: str, reconcile_interval: int) -> None:
-        """Construct RunnerManager object for creating and managing runners.
+        # Create runner instance.
+        instance_config: LxdInstanceConfig = {
+            "name": self.config.name,
+            "type": "virtual-machine",
+            "source": {
+                "type": "image",
+                "mode": "pull",
+                "server": "https://cloud-images.ubuntu.com/daily",
```
weiiwang01, line 249: Should we use a more stable version of Ubuntu instead of daily.

yhaliaw: Good idea.
Is there a link for the current stable release, or should I change this to say `jammy`?

weiiwang01: There isn't a link for stable release, so jammy is a good idea.

weiiwang01 (approved):

jdkandersson (approved):

PR #37, opened 2023-02-23
End to end test

yhaliaw:
- Implements a basic end to end test in which the charm is deployed, a runner is spawned, an GitHub workflow is ran. Part of the testing process will be manual for now, and can be automated later.
- Add bug fixes.
- Added some unit test for to pass coverage requirements.
- Add more logging for debugging.

Bug fixes includes:
- From testing, this method of setting proxy no longer works for ghapi library: https://github.com/fastai/ghapi/issues/50. Setting the env var for proxies are used instead. The env var should only affect the python process and child processes spawned.
- Change target series to jammy, to prevent dependency issues related to Rust binding of `cryptography` lib on focal.
- Use fixed file path for downloaded runner binary. Fix a missing binary problem, while avoid using `StoredState`.
- Fix a status related bug when executing `update-runner-bin` action.
- Fix typo related to GitHub organization API.

### DIFF
--- a/.github/workflows/end_to_end_test.yaml
+++ b/.github/workflows/end_to_end_test.yaml
@@ -14,6 +14,8 @@ jobs:
     runs-on: ${{ needs.get-runner-image.outputs.runs-on }}
     steps:
       - uses: actions/checkout@v3
+      - name: Initialize LXD
+        run: sudo lxd init --auto
       - name: Install charmcraft
         run: sudo snap install charmcraft --classic
       - name: Pack charm
--- a/.github/workflows/end_to_end_test.yaml
+++ b/.github/workflows/end_to_end_test.yaml
@@ -7,8 +7,17 @@ on:
 
 jobs:
   e2e-test:
-    name: end to end test
+    name: basic end to end test
     runs-on: [self-hosted, linux, x64, e2e-runner]
+    steps:
+      - name: Echo hello world
+        run: echo "hello world"
+  e2e-respawn-test:
+    name: respawning runner end to end test
+    runs-on: [self-hosted, linux, x64, e2e-runner]
+    strategy:
+      matrix:
+        runner: [1, 2]
     steps:
       - name: Echo hello world
         run: echo "hello world"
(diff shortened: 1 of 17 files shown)


.github/workflows/end_to_end_test.yaml
```diff
@@ -0,0 +1,16 @@
...
+on: 
+  pull_request:
+  workflow_call:
+  workflow_dispatch:
+
+jobs:
+  # Add a job for triggering the build, deploy, and configure the charm.
+  hello-world-test:
+    name: hello world test
+    runs-on: [self-hosted, linux, x64, demo-runner]
```
jdkandersson, line 12: Should we call them `e2e-test-runner` or something like that?

script/remove_runner.sh
```diff
@@ -0,0 +1,3 @@
+#!/usr/bin/env bash
+
+juju remove-application e2e-runner --force
```
jdkandersson, line 3: `--no-wait` and `--destroy-storage` as well?

.github/workflows/end_to_end_test.yaml
```diff
@@ -0,0 +1,18 @@
...
+  pull_request:
+  workflow_call:
+  workflow_dispatch:
+
+jobs:
+  e2e-test:
+    name: end to end test
+    runs-on: [self-hosted, linux, x64, e2e-runner]
+    strategy:
+      # Two tests are schedule to test if the charm is able to spawn new runners.
```
arturo-seijas, line 13: s/schedule/scheduled

tests/unit/mock.py
```diff
@@ -79,6 +79,8 @@ def start(self, wait: bool = True, timeout: int = 60):
 
     def stop(self, wait: bool = True, timeout: int = 60):
         self.status = "Stopped"
+        # Ephemeral virtual machine should delete on stop.
```
arturo-seijas, line 82: s/delete/be deleted

jdkandersson (dismissed):
Thanks!

jdkandersson (approved):

weiiwang01 (approved):

PR #39, opened 2023-03-14
Temporary fix for missing tools in runners

yhaliaw:
Self-hosted runners spawned by the GitHub runner charm does not have some common tooling users of GitHub Actions expects. This is due to the virtual machine hosting the self-hosted runners are created with the standard Ubuntu images.

Support for custom images based on [GitHub Actions Runner Images](https://github.com/actions/runner-images) is planned for the future and should be the permanent fix for this issue. In the mean time, installing the missing tools after starting up a virtual machine serves as a temporary fix.

### DIFF
--- a/src/runner.py
+++ b/src/runner.py
@@ -358,19 +358,25 @@ class Runner:
         # machines are created from custom images/GitHub runner image.
         logger.info("Setting up installation of docker...")
         self._execute(["/usr/bin/apt", "update"])
-        self._execute(["/usr/bin/apt", "install", "ca-certificates", "curl", "gnupg", "lsb-release"])
-        self._execute(["/usr/bin/mkdir", "-m", "0755", "-p", "/etc/apt/keyrings"])
         self._execute(
+            ["/usr/bin/apt", "install", "ca-certificates", "curl", "gnupg", "lsb-release"]
+        )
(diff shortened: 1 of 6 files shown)


tox.ini
```diff
@@ -93,7 +93,7 @@ commands =
 [testenv:static]
 description = Run static analysis tests
 deps =
-    bandit
+    bandit[toml]
     toml
```
arturo-seijas, line 97: You'll need to substitute toml for bandit[toml]

src/runner.py
```diff
@@ -354,6 +354,57 @@ def _install_binary(self, binary: Path) -> None:
         if self.instance is None:
             return
 
+        # TEMP: Install common tools used in GitHub Actions. This will be removed once virtual
+        # machines are created from custom images/GitHub runner image.
+        logger.info("Setting up installation of docker...")
+        self._execute(["/usr/bin/apt", "update"])
+        self._execute(["/usr/bin/apt", "install", "ca-certificates", "curl", "gnupg", "lsb-release"])
```
arturo-seijas, line 361: Aren't some of these already installed by default in Ubuntu? i.e. lsb-release?

arturo-seijas (commented):
Some comments inline

jdkandersson (approved):

yanksyoon (approved):
LGTM!

PR #43, opened 2023-03-22
Add more tools in self-hosted runners for GitHub Actions.

yhaliaw:
Some GitHub Actions used by operator-workflow uses the following tools:
- yq
- jq
- shellcheck

These tools are now installed by the LXD instance on start up.

### DIFF
--- a/src/runner.py
+++ b/src/runner.py
@@ -376,7 +376,7 @@ class Runner:
         self._execute(["/usr/bin/apt", "install", "-yq", "jq"])
 
         logger.info("Installing yq")
-        self._execute(["/usr/bin/apt", "install", "-yq", "yq"])
+        self._execute(["/usr/bin/snap", "install", "yq"])
 
         # The LXD instance is meant to run untrusted workload. Hardcoding the tmp directory should
         # be fine.
--- a/src/runner.py
+++ b/src/runner.py
@@ -369,6 +369,9 @@ class Runner:
         logger.info("Installing Python 3 pip...")
         self._execute(["/usr/bin/apt", "install", "-yq", "python3-pip"])
 
+        logger.info("Installing shellcheck")
+        self._execute(["/usr/bin/apt", "install", "-yq", "shellcheck"])
+
         # The LXD instance is meant to run untrusted workload. Hardcoding the tmp directory should
         # be fine.
         binary_path = "/tmp/runner.tgz"  # nosec B108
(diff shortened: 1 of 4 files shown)


src/runner.py
```diff
@@ -369,6 +369,15 @@ def _install_binary(self, binary: Path) -> None:
         logger.info("Installing Python 3 pip...")
         self._execute(["/usr/bin/apt", "install", "-yq", "python3-pip"])
 
+        logger.info("Installing shellcheck")
+        self._execute(["/usr/bin/apt", "install", "-yq", "shellcheck"])
```
jdkandersson, line 373: Perhaps it is worth adding a wrapping function around `self._execute` to install `apt` and `snap` packages that do the logging etc so that we can simplify this a bit? Or maybe that function even takes a list of apt packages and then installs them?

jdkandersson (dismissed):

jdkandersson (dismissed):

src/runner.py
```diff
@@ -499,3 +494,31 @@ def _execute(self, cmd: list[str], cwd: Optional[str] = None, **kwargs) -> str:
...
+    def _apt_install(self, packages: Iterable[str]) -> None:
+        """Installs the given APT packages.
+
+        This is a temporary solution to provide tools not offered by the base ubuntu image. Custom
+        images based on the GitHub action runner image will be used in the future.
+
+        Args:
+            packages: Packages to be install via apt.
+        """
+        self._execute(["/usr/bin/apt", "update"])
```
nrobinaubertin, line 507: Is it correct to use `apt` here instead of `apt-get` ? The former is intended as an end user interface (and is more subject to changes).  
This comment applies for the other invocations of `apt`

src/runner.py
```diff
@@ -499,3 +494,31 @@ def _execute(self, cmd: list[str], cwd: Optional[str] = None, **kwargs) -> str:
...
+
+        for pkg in packages:
+            logger.info("Installing %s via APT...", pkg)
+            self._execute(["/usr/bin/apt", "install", "-yq", pkg])
```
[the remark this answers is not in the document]
yhaliaw: I was debating over this, but decided this way it would be cleaner in the log which installation failed when there is a failure.

src/runner.py
yhaliaw: Updated the code to change `apt` to `apt-get`.

weiiwang01 (approved):

jdkandersson (approved):

src/runner.py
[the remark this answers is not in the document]
gregory-schiano: Ok let's keep it like this

gregory-schiano (approved):
LGTM

PR #45, opened 2023-03-27
Fix removal of offline runners.

yhaliaw:
Fixing a bug causing offline runners to be not removed from GitHub.

### DIFF
--- a/src/runner_manager.py
+++ b/src/runner_manager.py
@@ -71,7 +71,8 @@ class RunnerManager:
 
     def __init__(
         self,
-        instance_name: str,
+        app_name: str,
+        unit: int,
         runner_manager_config: RunnerManagerConfig,
         image: str = "jammy",
         proxies: ProxySetting = ProxySetting(),
@@ -79,11 +80,13 @@ class RunnerManager:
         """Construct RunnerManager object for creating and managing runners.
 
         Args:
-            instance_name: An name for the set of runners.
+            app_name: An name for the set of runners.
+            unit: Unit number of the set of runners.
             image: Image to use for the runner LXD instances.
             proxies: HTTP proxy settings.
         """
-        self.instance_name = instance_name
+        self.app_name = app_name
+        self.instance_name = f"{app_name}-{unit}-"
         self.config = runner_manager_config
         self.image = image
         self.proxies = proxies
@@ -287,7 +290,7 @@ class RunnerManager:
             logger.info("Adding %i additional runner(s).", delta)
             for _ in range(delta):
                 config = RunnerConfig(
-                    self.instance_name,
+                    self.app_name,
                     self.config.path,
                     self.proxies,
                     self._generate_runner_name(),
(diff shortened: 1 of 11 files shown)


src/runner_manager.py
```diff
@@ -422,3 +437,43 @@ def create_runner_info(
...
+        if isinstance(self.config.path, GitHubRepo):
+            token = self._clients.github.actions.create_registration_token_for_repo(
+                owner=self.config.path.owner, repo=self.config.path.repo
+            )
+        elif isinstance(self.config.path, GitHubOrg):
+            token = self._clients.github.actions.create_registration_token_for_org(
+                org=self.config.path.org
+            )
+        else:
+            assert_never(token)
```
jdkandersson, line 457: What does this do out of curiosity?

tests/unit/mock.py
```diff
@@ -156,10 +159,24 @@ def list_runner_applications_for_org(self, org: str):
         return self._list_runner_applications()
 
     def create_registration_token_for_repo(self, owner: str, repo: str):
-        return RegisterToken({"token": "test registration token"})
+        return RegistrationToken(
+            {"token": "test registration token", "expires_at": "2020-01-22T12:13:35.123-08:00"}
```
jdkandersson, line 163: Perhaps use secrets.hex_token?

tests/unit/test_runner.py
```diff
@@ -131,7 +129,7 @@ def test_remove(
     """
 
     runner.create("test_image", vm_resources, binary_path, "test_token")
```
jdkandersson, line 131: perhaps use secrets.hex_token for test tokens? Also applies for other tokens

src/runner_manager.py
yhaliaw: It assert to the type system that this branch in the if-else statement can never be reached.

This is a new feature in Python 3.11 or typing-extensions 4.1:
https://docs.python.org/3/library/typing.html#typing.assert_never

The link includes an example usage:
https://typing.readthedocs.io/en/latest/source/unreachable.html

tests/unit/mock.py
```diff
@@ -156,10 +159,24 @@ def list_runner_applications_for_org(self, org: str):
...
+
+    def create_remove_token_for_repo(self, owner: str, repo: str):
+        return RemoveToken(
+            {"token": "test registration token", "expires_at": "2020-01-22T12:13:35.123-08:00"}
+        )
+
+    def create_remove_token_for_org(self, org: str):
+        return RemoveToken(
+            {"token": "test registration token", "expires_at": "2020-01-22T12:13:35.123-08:00"}
+        )
```
yanksyoon: Would it be simpler to have a mock remove token as a constant defined somewhere? :) It seems like it is getting used in a few places.

yanksyoon (approved):
Minor suggestions :) LGTM!

tests/unit/mock.py
yhaliaw: I updated to use a test fixture that generates the token with `secrets.token_hex`.

tests/unit/mock.py
yhaliaw: Updated

weiiwang01 (approved):

PR #48, opened 2023-03-31
Added remome-offline-runners script

tingdahl:
Removes offline runners caused by bugs/mishaps in the charm. Most of it not written by me, but by Canonical colleague. My modifications are:

* page_size=100 on query call
* while True loop to run the process until there are no more runners to remove

### DIFF
--- a/script/remove_offline_runners.py
+++ b/script/remove_offline_runners.py
@@ -24,6 +24,7 @@ def get_runners():
                 "Authorization": "Bearer " + TOKEN,
                 "Accept": "application/vnd.github+json",
             },
+            timeout=60,
         )
 
         response.raise_for_status()
@@ -59,6 +60,7 @@ def delete_runner(runner):
                 "Authorization": "Bearer " + TOKEN,
                 "Accept": "application/vnd.github+json",
             },
+            timeout = 60,
         )
 
         response.raise_for_status()
@@ -69,11 +71,7 @@ def delete_runner(runner):
 
 
 if __name__ == "__main__":
-    while True:
-        runners = get_runners()
-        offline_runners = filter_offline_runners(runners)
-        if len(offline_runners)<=0:
-            break
+    while offline_runners := filter_offline_runners(get_runners()) > 0:
 
         for runner in offline_runners:
             delete_runner(runner)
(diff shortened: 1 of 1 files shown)


script/remove_offline_runners.py
```diff
@@ -0,0 +1,79 @@
...
+
+        response.raise_for_status()
+    except requests.HTTPError as http_err:
+        sys.exit(f"HTTP error occurred: {http_err}")
+    except Exception as err:
+        sys.exit(f"Other error occurred: {err}")
+
+
+if __name__ == "__main__":
+    while True:
```
jdkandersson, line 72: How about this?

`while offline_runners := filter_offline_runners(get_runners()) > 0:`

This is less prone to infinite loop bugs since the while loop has the exit condition rather than relying on break. Any thoughts?

script/remove_offline_runners.py
```diff
@@ -0,0 +1,79 @@
...
+TOKEN = "<your token>"
+
+
+logging.basicConfig(level=logging.INFO)
+logger = logging.getLogger(__name__)
+
+
+def get_runners():
+    try:
+        response = requests.get(
```
weiiwang01, line 19: requests does not have a default timeout, without timeout requests might wait indefinitely.

jdkandersson (dismissed):

script/remove_offline_runners.py
```diff
@@ -0,0 +1,77 @@
+# Copyright 2022 Canonical
+# See LICENSE file for licensing details.
+
+import sys
+import logging
+
+import requests
+
+ORG = "<your org>"
+TOKEN = "<your token>"
```
yanksyoon, line 10: I guess this was intended to be a quick script to be run as simple as possible, I think we could turn this in to arguments with the argparse library later :) Out of the scope of the PR, please feel free to close this comment

yanksyoon (dismissed):
Minor suggestions, LGTM!

jdkandersson (approved):

merkata (approved):
Looks great!

jdkandersson:
The test failure I'm guessing is because the runners weren't available in staging, @yhaliaw should be able to get the to pass when he is back

yhaliaw:
I have rerun the e2e-test.
This PR is ready for merging.

PR #50, opened 2023-04-13
Fix microk8s proxy settings

arturo-seijas:
Fix microk8s proxy settings by addint the proxy env vars to `/etc/environment`


merkata:
Looks great so far, I would think about the naming, that _ConfigConfig_.

PR #51, opened 2023-04-14
Load /etc/environment file to runner

yhaliaw:
Load the /etc/environment file with contents from a jinja template.


.github/workflows/end_to_end_test.yaml
```diff
@@ -12,6 +12,8 @@ jobs:
     steps:
       - name: Echo hello world
         run: echo "hello world"
+      - name: Proxy set in /etc/environment
```
jdkandersson, line 15: What about using microk8s? E.g., installing it?

templates/environment.j2
```diff
@@ -0,0 +1,7 @@
+PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/snap/bin
+HTTP_PROXY={{proxies['http']}}
+HTTPS_PROXY={{proxies['https']}}
+NO_PROXY={{proxies['no_proxy']}}
+http_proxy={{proxies['http']}}
+https_proxy={{proxies['https']}}
+no_proxy={{proxies['no_proxy']}}
```
jdkandersson, line 7: Final new line?

.github/workflows/end_to_end_test.yaml
jdkandersson: And deploy maybe a hello world container and check it works?

.github/workflows/end_to_end_test.yaml
```diff
@@ -12,6 +12,14 @@ jobs:
...
+        run: cat /etc/environment | grep PROXY
+      - name: Install microk8s
+        run: sudo snap install microk8s --classic
+      - name: Wait for microk8s
+        run: sudo microk8s status --wait-ready
+      - name: Deploy nginx for testing
+        run: sudo microk8s kubectl create deployment nginx --image=nginx
```
jdkandersson, line 22: Should we also include a check that the deployment actually worked? Like a curl or something?

weiiwang01, line 22: Should we check that nginx is actually deployed? For example `kubectl rollout status deployment/nginx --timeout=5m`

yhaliaw: Good idea

jdkandersson (dismissed):

arturo-seijas (approved):
LGTM

gregory-schiano (approved):
LGTM

PR #52, opened 2023-04-17
Lxd interface

yhaliaw:
Add an interface layer for the LXD client.
This is to allow separation of the underlying LXD client implementation and interface to be called by the charm code.


src/lxd.py
```diff
@@ -0,0 +1,321 @@
...
+
+"""LXD client interface.
+
+The Lxd class is intend to be layer of abstraction to isolate the underlying implementation of LXD.
+"""
+from __future__ import annotations
+
+import io
+import tempfile
+from subprocess import CalledProcessError  # nosec B404
```
jdkandersson, line 12: For these and any other disables we have a standard for how to document disables: https://github.com/canonical/is-charms-contributing-guide/blob/main/CONTRIBUTING.md#static-code-analysis, is it applicable here?

src/lxd.py
```diff
@@ -0,0 +1,321 @@
...
+
+        if mode:
+            lxc_cmd += ["--mode", str(mode)]
+
+        try:
+            execute_command(lxc_cmd)
+        except CalledProcessError as err:
+            raise LxdError(f"Unable to push file into LXD instance {self.instance.name}") from err
+
+    def put(self, filepath: str, content: Union[str, bytes], mode: Optional[str] = None) -> None:
```
jdkandersson, line 72: What is the difference between `put` and `push`? I think one pushes an existing fle and the other just based on a string, are both needed?

src/lxd.py
```diff
@@ -0,0 +1,321 @@
...
+        ]
+
+        try:
+            execute_command(lxc_cmd)
+        except CalledProcessError as err:
+            raise LxdError(
+                f"Unable to pull file {str(source)} from LXD instance {self.instance.name}"
+            ) from err
+
+    def get(self, filepath: str) -> str:
```
jdkandersson, line 117: Similar question as for the other one, do we need both of the `pull` and `get`? Also, any thoughts about naming? Perhaps this one is more `get_content` and the other one is `pull_file`?

src/lxd.py
```diff
@@ -0,0 +1,321 @@
...
+        lxc_cmd = [
+            "/snap/bin/lxc",
+            "file",
+            "push",
+            source,
+            f"{self.instance.name}/{destination.lstrip('/')}",
+        ]
+
+        if mode:
+            lxc_cmd += ["--mode", str(mode)]
```
arturo-seijas, line 65: Why casting `str(mode)`. It is already a string

src/lxd.py
```diff
@@ -0,0 +1,321 @@
...
+    @property
+    def status(self) -> str:
+        """Status of the LXD instance.
+
+        Returns:
+            Status of the LXD instance.
+        """
+        return self._pylxd_instance.status
+
+    def start(self, timeout: int = 30, force: bool = True, wait: bool = False) -> None:
```
jdkandersson, line 163: Do we need all these arguments? Or could this be simplified?

src/lxd.py
```diff
@@ -0,0 +1,321 @@
...
+
+        Raises:
+            LxdException: Unable to stop the LXD instance.
+        """
+        try:
+            self._pylxd_instance.stop(timeout, force, wait)
+        except pylxd.exceptions.LXDAPIException as err:
+            raise LxdError(f"Unable to stop LXD instance {self.name}") from err
+
+    def execute(self, cmd: list[str], cwd: Optional[str] = None) -> Tuple[int, IO, IO]:
```
jdkandersson, line 195: SOmething to think about: consider whether there should be a higher level abstraction that is the actual commands that should be executed rather than just re-exposing the execute interface. E.g., think about what the charm would need to execute on the instance and make those Python functions which are then converted to commands and the output parsed etc. That way, the messy details of writing commands is abstracted from most of the code. What do you think?

src/lxd.py
```diff
@@ -0,0 +1,321 @@
...
+
+    def __init__(self, pylxd_client: pylxd.Client):
+        """Construct the LXD instance manager.
+
+        Args:
+            pylxd_client: Instance of pylxd.Client.
+        """
+        self._pylxd_client = pylxd_client
+
+    def all(self) -> list[LxdInstance]:
```
jdkandersson, line 229: Should this return a tuple instead of a list?

src/lxd.py
jdkandersson, line 1: Some higher level thoughts: the wrapper looks fairly thin at the moment, i.e., it mainly re-exposes the underlying interface. Something to consider is to go a bit higher level and simplify the external interfaces to exactly what the user needs from LXD. These kind of interfaces are more useful and simplify testing etc a lot more if they expose functions that do what the user needs rather than expose the implementation details

jdkandersson (commented):
Mainly reviewed the lxd interface, let's look at the feedback on that as I suspect that the code might need to change a bit

src/lxd.py
yanksyoon, line 65:
```suggestion
            lxc_cmd += ["--mode", mode]
```
I think str conversion can be omitted :)

src/lxd.py
```diff
@@ -0,0 +1,321 @@
...
+            "pull",
+            f"{self.instance.name}/{source.lstrip('/')}",
+            destination,
+        ]
+
+        try:
+            execute_command(lxc_cmd)
+        except CalledProcessError as err:
+            raise LxdError(
+                f"Unable to pull file {str(source)} from LXD instance {self.instance.name}"
```
yanksyoon, line 114:
```suggestion
                f"Unable to pull file {source} from LXD instance {self.instance.name}"
```
The str conversion can be omitted :)

src/lxd.py
```diff
@@ -0,0 +1,321 @@
...
+        files (LxdInstanceFiles): Manager for the files on the LXD instance.
+        status (str): Status of the LXD instance.
+    """
+
+    def __init__(self, name: str, pylxd_instance: pylxd.models.Instance):
+        """Construct the LXD instance representation.
+
+        Args:
+            pylxd_instance: Instance of pylxd.models.Instance for the LXD instance.
+        """
```
yanksyoon, line 149: The name argument seems to be missed :)

src/lxd_type.py
```diff
@@ -0,0 +1,49 @@
...
+ResourceProfileConfig.__doc__ = "Represent LXD profile configuration."
+
+
+class ResourceProfileDevicesRoot(TypedDict):
+    """Represents LXD device profile."""
+
+    path: str
+    pool: str
+    type: str
+    size: str
```
yanksyoon, line 23: I think it might be helpful if the attributes could be explained with docstrings, for instance `size: str`: `size in integer with unit. E.g. 10G`

yanksyoon (dismissed):
Looks great! Added some minor comments!

src/lxd.py
yanksyoon: There is already a comment addressing this, resolving, sorry about the duplicated issue!

src/lxd.py
yanksyoon: There is already a comment addressing this, resolving, sorry about the duplicated issue!

src/lxd.py
yhaliaw: One is for pushing file and one is for pushing string.

The runner application is a tar file downloaded. So it is pushed by the file, rather than reading the bytes of the file and then convert to a temp file to be pushed.
Other configuration files are put into the instance with by the string, as they are a string generated from a template.

src/lxd.py
yhaliaw: For imports, from my testing the formatter with insert a blank line above the comments in the import section. This will mess up the import sections.

I think a while back in a PR for GitHub runner charm. This came up in a PR, I remember I decide not comment on the import and rather comment when the import is used.
This has the advantage of:
- Not messing up the formatter for the import section.
- Justify each instance of the usage, rather then the whole file.

Should I update the contributing guideline with this?

src/lxd.py
yhaliaw: Good catch. I change the data type for mode from an earlier version, and did not caught this.

src/lxd.py
yhaliaw: Currently, pull is not used. But I added it to mirror put and push.

I think the naming style of `get_content` and `pull_file` is clearer. I will be updating it  names.

src/lxd.py
yhaliaw: The `all` method returns a list of LXD instances, there could be zero instances or ten instances returned depend on what LXD instance exists at the monment.
Since it is variable length it should be list or iterator.

src/lxd.py
yhaliaw: Good idea.
There is a function that puts a file into the LXD instance and verify if the file content is the correctly loaded into the LXD instance in `runner.py`. That function might be better in the `lxd.py`. I will add more functions like this.

src/lxd.py
jdkandersson: Yes let's update the contributing guide

tests/unit/mock.py
```diff
@@ -100,13 +100,13 @@ def __init__(self):
     def mk_dir(self, path):
         pass
 
-    def push(self, source: str, destination: str, mode: Optional[str] = None):
+    def push_file(self, source: str, destination: str, mode: Optional[str] = None):
         self.files[destination] = "mock_content"
 
-    def put(self, filepath: str, data: Union[bytes, str], mode: Optional[str] = None):
+    def put_content(self, filepath: str, data: Union[bytes, str], mode: Optional[str] = None):
```
arturo-seijas, line 106: `write_content` seems more obvious to me

src/lxd.py
```diff
@@ -62,14 +61,16 @@ def push(self, source: str, destination: str, mode: Optional[str] = None) -> Non
...
+            lxc_cmd += ["--mode", mode]
 
         try:
             execute_command(lxc_cmd)
-        except CalledProcessError as err:
+        except SubprocessError as err:
             raise LxdError(f"Unable to push file into LXD instance {self.instance.name}") from err
 
-    def put(self, filepath: str, content: Union[str, bytes], mode: Optional[str] = None) -> None:
+    def put_content(
```
arturo-seijas, line 71: write_content seems more obvious to me

arturo-seijas (approved):
LGTM

yanksyoon (approved):
LGTM! Thanks for the changes!

PR #57, opened 2023-05-12
Delete LXD instance if LXD failed to remove ephemeral instance

yhaliaw:
Fix a resource leak problem when LXD failed to remove an ephemeral instance on stop.

This PR includes:
- Delete LXD ephemeral instance that LXD failed to remove.
- Installs pylxd from source on GitHub to fix a bug in version on PyPI. [See this comment](https://github.com/lxc/pylxd/issues/538#issuecomment-1520142839).

### DIFF
--- a/charmcraft.yaml
+++ b/charmcraft.yaml
@@ -7,6 +7,7 @@ parts:
     charm-python-packages:
       - setuptools  # for jinja2
     build-packages:
+      - git  # for installing git source of pylxd 
       - libffi-dev  # for cffi
       - libssl-dev  # for cryptography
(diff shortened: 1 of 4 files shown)


jdkandersson (approved):

arturo-seijas (approved):

sirkadirov:
@yhaliaw Glad that my comment helped you 😊! And yeah, sometimes LXD fails to delete ephemeral instances... I admit, I didn't think that an additional call to the container removal function could improve the situation. Thanks for the idea, I'll try it out in my project.

PR #60, opened 2023-05-16
Add configuration needed by test.yaml in operator-workflow.

yhaliaw:
- Docker client proxy setting according to: https://docs.docker.com/network/proxy/#configure-the-docker-client. Should be the default proxy setting during run and build according to: https://docs.docker.com/network/proxy/#run-containers-with-a-proxy-configuration
- Set the `/usr/local/bin` permission to 777. This is the same as GitHub-hosted runners.
The GitHub-hosted runner has file permission set to 777:
```
ls -la /usr/local
...
drwxrwxrwx  2 root root  4096 May  7 22:37 bin
...
```
- Fix an issue with removal of runner on GitHub if runner failed to do so.
- Flush the runner on charm upgrade.

### DIFF
--- a/.github/workflows/end_to_end_test.yaml
+++ b/.github/workflows/end_to_end_test.yaml
@@ -12,8 +12,14 @@ jobs:
     steps:
       - name: Echo hello world
         run: echo "hello world"
+      - name: File permission for /usr/local/bin
+        run: ls -ld /usr/local/bin | grep drwxrwxrwx
       - name: Proxy set in /etc/environment
-        run: cat /etc/environment | grep PROXY
+        run: cat /etc/environment | grep HTTP_PROXY
+      - name: Proxy set in docker daemon 
+        run: cat /etc/systemd/system/docker.service.d/http-proxy.conf | grep HTTP_PROXY
+      - name: Proxy set in docker client
+        run: cat /home/ubuntu/.docker/config.json | grep httpProxy
       - name: Install microk8s
         run: sudo snap install microk8s --classic
       - name: Wait for microk8s
--- a/.github/workflows/end_to_end_test.yaml
+++ b/.github/workflows/end_to_end_test.yaml
@@ -1,6 +1,6 @@
 name: End to end tests
 
-on: 
+on:
   pull_request:
   workflow_call:
   workflow_dispatch:
@@ -14,10 +14,16 @@ jobs:
         run: echo "hello world"
       - name: File permission for /usr/local/bin
         run: ls -ld /usr/local/bin | grep drwxrwxrwx
+      - name: Test fille permission for /usr/local/bin
+        run: touch /usr/local/bin/test_file
+      # "Install microk8s" step will test if the proxies settings are correct.
       - name: Proxy set in /etc/environment
         run: cat /etc/environment | grep HTTP_PROXY
-      - name: Proxy set in docker daemon 
+      # "Update apt in python docker container" step will test docker default proxy settings due to
+      # pulling the python image.
+      - name: Proxy set in docker daemon
         run: sudo cat /etc/systemd/system/docker.service.d/http-proxy.conf | grep HTTP_PROXY
+      # "Update apt in python docker container" step will test docker client default proxy settings.
       - name: Proxy set in docker client
         run: cat /home/ubuntu/.docker/config.json | grep httpProxy
       - name: Install microk8s
@@ -28,6 +34,8 @@ jobs:
         run: sudo microk8s kubectl create deployment nginx --image=nginx
       - name: Wait for nginx to be ready
         run: sudo microk8s kubectl rollout status deployment/nginx --timeout=30m
+      - name: Update apt in python docker container
+        run: docker run python:3.10-slim apt update
   dep-test:
     # Test the dependencies in one job to avoid using too many runners at once.
     name: dependency test
@@ -50,4 +58,4 @@ jobs:
         run: python3 -m pip install check-jsonschema
       # Test program installed by pip. The directory `~/.local/bin` need to be added to PATH.
       - name: test check-jsonschema
-        run: check-jsonschema --version 
+        run: check-jsonschema --version
--- a/.github/workflows/end_to_end_test.yaml
+++ b/.github/workflows/end_to_end_test.yaml
@@ -14,7 +14,7 @@ jobs:
         run: echo "hello world"
       - name: File permission for /usr/local/bin
         run: ls -ld /usr/local/bin | grep drwxrwxrwx
-      - name: Test fille permission for /usr/local/bin
+      - name: Test file permission for /usr/local/bin
         run: touch /usr/local/bin/test_file
       # "Install microk8s" step will test if the proxies settings are correct.
       - name: Proxy set in /etc/environment
--- a/.github/workflows/end_to_end_test.yaml
+++ b/.github/workflows/end_to_end_test.yaml
@@ -17,7 +17,7 @@ jobs:
       - name: Proxy set in /etc/environment
         run: cat /etc/environment | grep HTTP_PROXY
       - name: Proxy set in docker daemon 
-        run: cat /etc/systemd/system/docker.service.d/http-proxy.conf | grep HTTP_PROXY
+        run: sudo cat /etc/systemd/system/docker.service.d/http-proxy.conf | grep HTTP_PROXY
       - name: Proxy set in docker client
         run: cat /home/ubuntu/.docker/config.json | grep httpProxy
       - name: Install microk8s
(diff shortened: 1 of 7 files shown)


src/runner.py
```diff
@@ -366,6 +367,9 @@ def _configure_runner(self) -> None:
         self.instance.execute(["/usr/bin/sudo", "chown", "ubuntu:ubuntu", str(self.runner_script)])
         self.instance.execute(["/usr/bin/sudo", "chmod", "u+x", str(self.runner_script)])
 
+        # Set permission to the same as GitHub-hosted runner for this directory.
+        self.instance.execute(["/usr/bin/sudo", "chmod", "777", "/usr/local/bin"])
```
jdkandersson, line 371: Should we have an end-to-end test that tries to interact with the folder to check that this permission is available to the ubuntu user?

yhaliaw: Ok, will be doing that.

.github/workflows/end_to_end_test.yaml
```diff
@@ -12,8 +12,14 @@ jobs:
     steps:
       - name: Echo hello world
         run: echo "hello world"
+      - name: File permission for /usr/local/bin
+        run: ls -ld /usr/local/bin | grep drwxrwxrwx
       - name: Proxy set in /etc/environment
-        run: cat /etc/environment | grep PROXY
+        run: cat /etc/environment | grep HTTP_PROXY
```
merkata, line 18: That wouldn't catch HTTPS_PROXY not being set, no? Do you know whether we rely on it being set, so perhaps add an additional check for it?

src/runner.py
merkata: I don't think this is nice - you mentioned `777` being the default on the github runners, do you know why and where it comes from? I'd really explore going with `755`, it might be a small attack vector, though really anyone can place a runnable code inside /usr/local/bin that way and that directory is almost guaranteed to be on $PATH and it can take precedence over /usr/bin and /bin on some installs, meaning I can overwrite an existing command with whatever I want to.

src/runner.py
```diff
@@ -417,8 +412,12 @@ def _configure_runner(self) -> None:
                 }
             }
             docker_client_proxy_content = json.dumps(docker_client_proxy)
-            for path in docker_config_paths:
-                self._put_file(str(path), docker_client_proxy_content)
+            # Configure the docker client for root user and ubuntu user.
+            self._put_file("/root/.docker/config.json", docker_client_proxy_content)
+            self._put_file("/home/ubuntu/.docker/config.json", docker_client_proxy_content)
+            self.instance.execute(
+                ["/usr/bin/chown", "-R", "ubuntu:ubuntu", "/home/ubuntu/.docker/config.json"]
```
arturo-seijas, line 419: You shouldn't need  to change ownership recursively `-R ` provided this is a single file

src/runner.py
jdkandersson: Keep in mind that this machine is essentially designed to allow the GitHub workflow to do anything, they can already sudo and mess with the entire machine. The machine gets thrown away and a new clean machine created for the next workflow run

.github/workflows/end_to_end_test.yaml
```diff
@@ -12,8 +12,14 @@ jobs:
     steps:
       - name: Echo hello world
         run: echo "hello world"
+      - name: File permission for /usr/local/bin
+        run: ls -ld /usr/local/bin | grep drwxrwxrwx
```
jdkandersson, line 16: Should we just do something in the directory that requires the permission?

.github/workflows/end_to_end_test.yaml
jdkandersson, line 18: What if we run some  sort of build process here that requires the proxy?

src/runner.py
yhaliaw: However, GitHub Action such as [SkyWalking Eyes](https://github.com/apache/skywalking-eyes) (used in operator-workflow for checking licences) requires write permission to /usr/local/bin as a user.
From testing with a GitHub-hosted runner, the /usr/local/bin does have `drwxrwxrwx`.
```
2023-05-16T06:26:52.8224628Z ##[group]Run ls -la /usr/local
2023-05-16T06:26:52.8224975Z  [36;1mls -la /usr/local [0m
2023-05-16T06:26:52.8281068Z shell: /usr/bin/bash -e {0}
2023-05-16T06:26:52.8281367Z ##[endgroup]
2023-05-16T06:26:52.8565654Z total 92
2023-05-16T06:26:52.8567120Z drwxr-xr-x 19 root root  4096 May  7 22:16 .
2023-05-16T06:26:52.8568273Z drwxr-xr-x 14 root root  4096 Apr 28 02:09 ..
2023-05-16T06:26:52.8569131Z drwxrwxrwx 10 root root  4096 May  7 21:49 .ghcup
2023-05-16T06:26:52.8570111Z drwxr-xr-x  3 root root  4096 May  7 21:27 aws-cli
2023-05-16T06:26:52.8570973Z drwxr-xr-x  3 root root  4096 May  7 21:27 aws-sam-cli
2023-05-16T06:26:52.8571891Z drwxrwxrwx  2 root root  4096 May  7 22:37 bin
2023-05-16T06:26:52.8572634Z drwxr-xr-x  3 root root  4096 Apr  4 19:10 doc
2023-05-16T06:26:52.8573722Z drwxr-xr-x  2 root root  4096 Apr 28 02:09 etc
2023-05-16T06:26:52.8574621Z drwxr-xr-x  2 root root  4096 Apr 28 02:09 games
2023-05-16T06:26:52.8575704Z drwxr-xr-x  3 root root  4096 May  7 21:58 include
2023-05-16T06:26:52.8576553Z drwxr-xr-x  8 root root  4096 May  7 22:16 julia1.8.5
2023-05-16T06:26:52.8577717Z drwxr-xr-x  8 root root  4096 May  7 22:18 lib
2023-05-16T06:26:52.8578477Z drwxr-xr-x  4 root root  4096 Apr  4 19:10 man
2023-05-16T06:26:52.8579463Z drwxr-xr-x  3 root root  4096 May  7 21:58 n
2023-05-16T06:26:52.8581003Z drwxr-xr-x  2 root root  4096 May  7 21:11 sbin
2023-05-16T06:26:52.8582187Z drwxr-xr-x  3 root root  4096 May  7 21:27 sessionmanagerplugin
2023-05-16T06:26:52.8583055Z drwxr-xr-x 23 root root  4096 May  7 22:40 share
2023-05-16T06:26:52.8584201Z drwxr-xr-x  2 root root 20480 May  7 21:57 sqlpackage
2023-05-16T06:26:52.8585631Z drwxr-xr-x  2 root root  4096 Apr 28 02:09 src
```

yhaliaw: While the GitHub Actions are run under a user, the user are given sudo access within the GitHub-hosted runner and self-hosted runner.
Any GitHub Actions could already overwrite binaries in /usr/local/bin or any other file.

src/runner.py
yhaliaw: This is fixed. The patch should be `/home/ubuntu/.docker`

.github/workflows/end_to_end_test.yaml
yhaliaw: Thanks. This is added.

.github/workflows/end_to_end_test.yaml
yhaliaw: I added a comment noting the Install microk8s would be testing proxy setting.
Including NO_PROXY as well.

yhaliaw: The "Install microk8s" step would be testing the HTTPS_PROXY and NO_PROXY settings.
I added a comment for it.

This line would be testing whether the write file succeed. Since all the proxy setting are written into the file in one operation.

jdkandersson (approved):

src/runner.py
```diff
@@ -155,16 +156,31 @@ def remove(self, remove_token: str) -> None:
...
-            self._clients.github.actions.delete_self_hosted_runner_from_repo(
-                owner=self.config.path.owner,
-                repo=self.config.path.repo,
-                runner_id=self.config.name,
-            )
-        if isinstance(self.config.path, GitHubOrg):
-            self._clients.github.actions.delete_self_hosted_runner_from_org(
-                org=self.config.path.org, runner_id=self.config.name
-            )
+        if self.status.runner_id is not None:
```
yanksyoon, line 159: Minor nitpick,
Would this
'''Python
if self.status.runner_id is None:
    Return
if isinstance(
...
'''
Help reduce indentation?

yanksyoon (approved):
LGTM!

src/runner.py
yhaliaw: Will be including this change in a future PR.

PR #64, opened 2023-06-01
Add config for disk IO limits.

yhaliaw:
- Add config for disk read IO limit and disk write IO limit.
- Bug fix for check-runner, fix the unit test for it as well. The bug was in both the code and unit test.

Pending discussion with SRE on the disk IO read and write limit.
The placeholder value 100 MiB read limit and 50 MiB write limit is used for now.

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -33,25 +33,26 @@ options:
     type: string
     default: 7GiB
     description: >
-      Amount of memory to allocate per virtual machine runner. Positive integers with GiB suffix.
+      Amount of memory to allocate per virtual machine runner. Positive integers followed by the unit.
+      The unit could be the follow KiB, MiB, GiB, TiB.
   vm-disk:
     type: string
     default: 10GiB
     description: >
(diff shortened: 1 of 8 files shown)


config.yaml
```diff
@@ -40,6 +40,18 @@ options:
     description: >
       Amount of disk space to allocate to root disk for virtual machine runner. Positive integers
       with GiB suffix.
+  vm-disk-read:
+    type: string
+    default: 100MiB
+    description: >
+      Limit of disk read IO per second of the root disk for virtual machine runner. Positive integers
```
arturo-seijas, line 47: Why MiB specifically? I don't see that limitation in the code. Also, the name of the config doesn't seem too obvious to me, I'd rename it to `vm-disk-read-limit` or something similar (same for the other config).

yhaliaw: I have updated the descriptions.

arturo-seijas (approved):
LGTM

yhaliaw:
The default installation of LXD no longer support disk IO limit. The setting does not work anymore.
Close this PR due to this.

PR #65, opened 2023-06-05
Fix check-runners juju action

yhaliaw:
Fix check-runners juju action due to a bug with using the enum value.


weiiwang01 (approved):

weiiwang01 (approved):

yanksyoon (approved):
LGTM!

PR #69, opened 2023-06-15
Using ramdisk as LXD storage for the runners

yhaliaw:


jdkandersson (approved):

yhaliaw:
From manual testing, it seems the memory that is used by the LXD storage is not getting free.
Attempting another method for now.

PR #70, opened 2023-06-15
Implement fully automated end-to-end CI testing workflow

weiiwang01:
Updated the `github-runner` charm to include a new 'test mode'. This functionality facilitates the charm's deployment on an LXD container cloud, using LXD nesting. Building upon this 'test mode', we have also developed a new e2e test that is fully automated, and operable on GitHub hosted runners. This also resolves the bootstrap problem that we had encountered with our previous e2e test configuration, which requires a self-hosted runner to test self-hosted runner.


.github/workflows/e2e_test.yaml
```diff
@@ -0,0 +1,191 @@
...
+            --config token=${{ secrets.E2E_TESTING_TOKEN }} \
+            --config virtual-machines=1 \
+            --config test-mode=insecure
+
+      - name: Watch github-runner
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        timeout-minutes: 30
+        run: |
+          juju debug-log --replay --tail &
```
gregory-schiano, line 119: I suppose these logs are there for debug purpose ?
The model is configured with level DEBUG, this can produce a lot of logs here

.gitignore
```diff
@@ -5,3 +5,4 @@ placeholders/
 *.charm
 build/
 .coverage
+lxd-profile.yaml
```
gregory-schiano, line 8: Missing line at the end

src/lxd.py
```diff
@@ -47,7 +47,22 @@ def mk_dir(self, dir_name: str) -> None:
...
+            "-p",
+            dir_name,
+        ]
+        try:
+            execute_command(lxc_command)
+        except SubprocessError as err:
+            logger.exception("Failed to create directory")
+            raise LxdError(
+                f"Unable to create directory in LXD instance {self.instance.name}"
+            ) from err
```
gregory-schiano, line 65: Why making this change ?

gregory-schiano (commented):
I'm not a big big fan of the "test-mode" config parameter just for the E2E tests, but to be honest couldn't think about an good alternative for now

src/lxd.py
weiiwang01: I think this is the residue of an early debug attempt, removed, thanks!

.gitignore
weiiwang01: Added, thanks!

.github/workflows/e2e_test.yaml
weiiwang01: Yes, it's intended. The debug messages are very helpful.

.github/workflows/e2e_test.yaml
```diff
@@ -0,0 +1,191 @@
...
+            kill $(jobs -p)
+            exit 1
+          fi
+
+  e2e-test:
+    name: End-to-End Test
+    needs: [ build-charm, run-id ]
+    runs-on: [ self-hosted, linux, x64, "${{ needs.run-id.outputs.run-id }}" ]
+    steps:
+      - name: Echo hello world
```
amandahla, line 144: Why is this echo here?

src/charm.py
```diff
@@ -122,7 +123,10 @@ def __init__(self, *args, **kargs) -> None:
                 class.
         """
         super().__init__(*args, **kargs)
-
+        if LXD_PROFILE_YAML.exists():
```
amandahla, line 126: Question: Couldn't this be checked on config_changed? So the user can change the test-mode value if this error happens.

.github/workflows/e2e_test.yaml
jdkandersson: This is how the runners are tested, so this is a fairly basic test where the runner only has to be able to connect and be able to echo

jdkandersson (dismissed):

src/charm.py
weiiwang01: The `test-mode` here is not intended for user to change the charm into test mode. This configuration is to prevent human error that enabling the test mode in the production environment.

By turning on the `test-mode` alone, the charm does nothing, the developer need to build a special test charm that contains certain files to truly enable test mode. But if someone accidentally uploaded the special test charm to charmhub, this configuration will prevent it from running in the production environment.

.github/workflows/e2e_test.yaml
amandahla: Ok, if possible would be nice to have a comment clarifying that 👍

yhaliaw (dismissed):

jdkandersson (approved):

.github/workflows/e2e_test.yaml
weiiwang01: Added, thanks!

yhaliaw (approved):

PR #73, opened 2023-06-27
Disable resizing of tmpfs

yhaliaw:
TODO: Still testing this.
Resizing of tmpfs while it is in use by LXD VM, might be causing LXD VM to crash.


yhaliaw:
This is unlikely the cause of crash.
Tried manually resize a tmpfs on unit with runner. Afterwards, the runner successfully executed test job that includes installing microk8s.

PR #74, opened 2023-06-27
Add firewall to limit network access for runners

weiiwang01:
Add a charm configuration `allowlist` to configure runners to only access necessary ports and IP addresses, minimizing the potential surface for attacks.


templates/github-runner-firewall.j2
```diff
@@ -0,0 +1,21 @@
...
+        iifname "tap*" ip daddr { {{ host_ip }}, 255.255.255.255 } udp dport 67 counter accept
+        iifname "tap*" ip daddr {{ host_ip }} tcp dport 8080 counter accept
+
+{% for entry in allowlist %}
+        iifname "tap*" ip daddr {{ entry.ip_range }} {{ "udp" if entry.is_udp else "tcp" }} dport {{ entry.port_range }} counter accept
+{% endfor %}
+
+        iifname "tap*" counter reject with icmpx type admin-prohibited
+    }
+}
```
jdkandersson, line 21: Should this include a final EOL?

jdkandersson:
Thanks for putting this together!

templates/github-runner-firewall.j2
weiiwang01: Added, thanks!

src/firewall.py
```diff
@@ -0,0 +1,124 @@
...
+                raise_format_error()
+            port_start, port_end = port_range_parts
+            check_valid_port(port_start)
+            check_valid_port(port_end)
+        else:
+            check_valid_port(port_range)
+        return cls(ip_range=ip_range, port_range=port_range, is_udp=is_udp)
+
+
+class Firewall:  # pylint: disable=too-few-public-methods
```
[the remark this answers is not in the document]
weiiwang01: I think the name of the check, `too-few-public-methods`, is self-explanatory enough in this situation.

jdkandersson (dismissed):

src/firewall.py
```diff
@@ -0,0 +1,124 @@
...
+                ipaddress.IPv4Network(network)
+            except ValueError:
+                raise_format_error()
+
+        def check_valid_port(port: str):
+            """Check if the given string is a valid port number."""
+            if not port.isdigit() or int(port) < 1 or int(port) > 65535:
+                raise_format_error()
+
+        parts = entry.split(":")
```
arturo-seijas, line 58: Is something like 0.0.0.0/0:1-65535:udp expected to fail?

tests/unit/test_firewall.py
```diff
@@ -0,0 +1,69 @@
...
+    ),
+    pytest.param("10.0.0.1/8:11-22:udp", None),
+    pytest.param("a.b.c.d:123", None),
+    pytest.param("", None),
+    pytest.param("1.2.3.4/a", None),
+)
+
+
+@pytest.mark.parametrize("entry,excepted_firewall_entry", TESTING_FIREWALL_ENTRIES)
+def test_parse_firewall_entry(entry: str, excepted_firewall_entry: typing.Optional[FirewallEntry]):
```
arturo-seijas, line 28: s/excepted_firewall_entry/expected_firewall_entry?

src/charm.py
```diff
@@ -549,7 +563,9 @@ def _install_deps(self) -> None:
             env["NO_PROXY"] = self.proxies["no_proxy"]
             env["no_proxy"] = self.proxies["no_proxy"]
 
-        execute_command(["/usr/bin/apt-get", "install", "-qy", "gunicorn", "python3-pip"])
+        execute_command(
+            ["/usr/bin/apt-get", "install", "-qy", "gunicorn", "python3-pip", "nftables"]
```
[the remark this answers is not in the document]
weiiwang01: It is used by repo-policy-compliance.

tests/unit/test_firewall.py
weiiwang01: Fixed, thanks!

src/charm.py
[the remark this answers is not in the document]
arturo-seijas: Can you add a comment in the code mentioning this? It isn't very obvious from the outside :)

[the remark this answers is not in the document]
weiiwang01: 👍 added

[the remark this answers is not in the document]
arturo-seijas: Thanks!

arturo-seijas (dismissed):
LGTM

arturo-seijas (dismissed):
LGTM

src/charm.py
```diff
@@ -577,8 +583,8 @@ def _install_deps(self) -> None:
         execute_command(["/usr/bin/snap", "refresh", "lxd", "--channel=latest/stable"])
         execute_command(["/snap/bin/lxd", "waitready"])
         execute_command(["/snap/bin/lxd", "init", "--auto"])
-        execute_command(["/usr/bin/chmod", "a+wr", "/var/snap/lxd/common/lxd/unix.socket"])
         execute_command(["/snap/bin/lxc", "network", "set", "lxdbr0", "ipv6.address", "none"])
+        execute_command(["/usr/bin/chmod", "a+wr", "/var/snap/lxd/common/lxd/unix.socket"])
```
amandahla, line 587: I believe it's out of the scope of this PR but why is this needed?

weiiwang01: I guess this is to call lxc command without sudo, but since the charm runs as root, I don't think this is technically necessary.

weiiwang01: I just removed that line and it seems to be fine.

amandahla (dismissed):
LGTM

config.yaml
```diff
@@ -58,3 +58,10 @@ options:
     description: >
       When set to 'insecure', the charm test mode is activated, which may deactivate some security
       hardening measures.
+  allowlist:
+    type: string
+    default: "0.0.0.0/0:1-65535, 0.0.0.0/0:1-65535:udp"
+    description: >
+      A list of IPv4 networks and tcp/udp ports that runners can access. Note that ICMP traffic 
+      and IPv6 traffic are always disallowed. Additionally, the runner will always have
+      access to essential services such as DHCP and DNS regardless of the allowlist configuration.
```
gtrkiller, line 67: What is the logic behind allowing everything in IPv4 by default from a security perspective?

weiiwang01: For the default value, we only have two options: allow all or allow none. But when the charm is set to allow none, the GitHub runner can't even start, so I opt to chose allow all as the default value.

gtrkiller: Makes sense, thanks!

.github/workflows/e2e_test.yaml
```diff
@@ -190,3 +198,7 @@ jobs:
       # Test program installed by pip. The directory `~/.local/bin` need to be added to PATH.
       - name: test check-jsonschema
         run: check-jsonschema --version
+      - name: Test Firewall
+        run: |
+          HOST_IP=$(ip route | grep default | cut -f 3 -d" ")
+          [ $((ping $HOST_IP -c 5 || :) | grep "Destination Port Unreachable" | wc -l) -eq 5 ]
```
jdkandersson, line 204: Final newline?

src/firewall.py
```diff
@@ -38,87 +35,118 @@ def decode(cls, entry: str) -> "FirewallEntry":
...
+            {
+                "action": "reject",
+                "destination": host_ip,
+                "protocol": "icmp4",
+                "state": "enabled",
+            },
+            {
+                "action": "reject",
+                "destination": "::/0",
+                "state": "enabled",
```
jdkandersson, line 128: What is the effect of this rule please?

weiiwang01: Reject all IPv6 traffics, since we don't provision a IPv6 network to the runner right now.

.github/workflows/e2e_test.yaml
weiiwang01: Updated, thanks!

src/firewall.py
jdkandersson: This feels a little Canonical specific, should that go to the denylist configuration rather than in our code? If someone else installs the charm on an IPv6 network I'm guessing it wouldn't work with this configuration?

weiiwang01: The LXD network inside the runner has IPv6 disabled as well, so I suspect it will work even with this rule removed.

jdkandersson: Ah ok

jdkandersson (approved):

src/firewall.py
```diff
@@ -0,0 +1,152 @@
...
+    """
+
+    ip_range: str
+
+    @classmethod
+    def decode(cls, entry: str) -> "FirewallEntry":
+        """Decode a firewall entry from a string.
+
+        Args:
+            entry: The firewall entry string, e.g. '192.168.0.1:80' or '192.168.0.0/24:80-90:udp'.
```
gregory-schiano, line 30: Shouldn't the description be adapted ? Only CIDR notation is supported

src/firewall.py
```diff
@@ -0,0 +1,152 @@
...
+                "protocol": "icmp4",
+                "state": "enabled",
+            },
+            {
+                "action": "reject",
+                "destination": "::/0",
+                "state": "enabled",
+            },
+        ]
+        host_network = ipaddress.IPv4Network(host_ip)
```
gregory-schiano, line 131: The variable name could be confusing, it's not the real host_network, but the host_ip represented as a /32 CIDR notation to be able to use `address_exclude` that requires an IPv4Network object

.github/workflows/e2e_test.yaml
```diff
@@ -109,6 +112,7 @@ jobs:
             --config path=${{ secrets.E2E_TESTING_REPO }} \
             --config token=${{ secrets.E2E_TESTING_TOKEN }} \
             --config virtual-machines=1 \
+            --config denylist=10.0.0.0/8 \
```
gregory-schiano, line 115: Does it cover host_ip which is included in the denylist range ?

.github/workflows/e2e_test.yaml
```diff
@@ -100,6 +100,9 @@ jobs:
         with:
           name: dangerous-test-only-github-runner_ubuntu-22.04-amd64.charm
 
+      - name: Enable br_netfilter
+        run: sudo modprobe br_netfilter
```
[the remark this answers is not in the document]
weiiwang01: Interesting, the `br_netfilter` is not necessary for the ACL firewall. It's only required by `security.ipv6_filtering=true`

src/firewall.py
weiiwang01: Fixed, thanks!

src/firewall.py
weiiwang01: Yes, the variable is a bad idea, I removed the variable and inlined the IPv4Network casting.

.github/workflows/e2e_test.yaml
weiiwang01: Yes, this is intended to test the capability of the charm to resolve firewall conflicts.

.github/workflows/e2e_test.yaml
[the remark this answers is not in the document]
gregory-schiano: Interesting... curious to understand how they manage the ACL then. Anyway unrelated to this PR, closing the comment

gregory-schiano (approved):
LGTM

PR #75, opened 2023-06-29
Handle unmount of tmpfs

yhaliaw:
If the juju unit is rebooted, the tmpfs would become unmounted.
This patch fixes the tmpfs issues arising from this.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -162,6 +162,7 @@ class GithubRunnerCharm(CharmBase):
         self.framework.observe(self.on.flush_runners_action, self._on_flush_runners_action)
         self.framework.observe(self.on.update_runner_bin_action, self._on_update_runner_bin)
 
+    @retry(tries=5, delay=15, max_delay=60, backoff=1.5)
     def _create_memory_storage(self, path: Path, size: int) -> None:
         """Create a tmpfs-based LVM volume group.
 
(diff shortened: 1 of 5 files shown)


src/charm.py
```diff
@@ -179,10 +180,20 @@ def _create_memory_storage(self, path: Path, size: int) -> None:
             else:
                 execute_command(["mount", "-o", f"remount,size={size}k", str(path)])
         except OSError as err:
-            logger.exception("Unable to create directory")
+            logger.exception("Unable to create storage directory")
+            # Ensure the path is not empty for next retry.
```
gregory-schiano, line 184: Shouldn't it be `Ensure the path is empty for the next retry.` ?

yhaliaw: You are right

src/charm.py
```diff
@@ -179,10 +180,20 @@ def _create_memory_storage(self, path: Path, size: int) -> None:
             else:
                 execute_command(["mount", "-o", f"remount,size={size}k", str(path)])
         except OSError as err:
-            logger.exception("Unable to create directory")
+            logger.exception("Unable to create storage directory")
+            # Ensure the path is empty for next retry.
+            if path.exists():
+                shutil.rmtree(path, ignore_errors=True)
+                path.rmdir()
+                logger.info("Cleaned up storage directory")
```
gregory-schiano, line 188: I wonder if a bit of factoring could help here, it would avoid repeating but also ensure that any change will be applied to the 2 parts where this is required

gregory-schiano (dismissed):

gregory-schiano (dismissed):
Great, thx for the last change :)

jdkandersson (dismissed):

jdkandersson (dismissed):

jdkandersson (approved):

yanksyoon (approved):
LGTM!

PR #76, opened 2023-07-04
Add upgrading of linux kernel

yhaliaw:
Automatically upgrade the kernel to avoid a possible bug in the 5.15 version of the kernel.
This should be a temporary fix as new version of Ubuntu would include newer kernel by default in the future.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -273,6 +273,12 @@ class GithubRunnerCharm(CharmBase):
         """
         self.unit.status = MaintenanceStatus("Installing packages")
 
+        # Temporary solution: Upgrade the kernel due to a kernel bug in 5.15. Kernel upgrade
+        # not needed for container-based end-to-end tests.
+        if not LXD_PROFILE_YAML.exists():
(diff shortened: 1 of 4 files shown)


jdkandersson (dismissed):

src/charm.py
```diff
@@ -37,7 +37,7 @@
 from utilities import bytes_with_unit_to_kib, execute_command, get_env_var, retry
 
 if TYPE_CHECKING:
-    from ops.model import JsonObject  # pragma: no cover
+    from ops.model import JsonObject  # type: ignore
```
yanksyoon, line 40: The latest ops library seems to have replaced JsonObject to typing.Any, I've implemented the changes in [yq PR](https://github.com/canonical/github-runner-operator/pull/77/files), :D just a heads up in case the linting fails

yanksyoon (approved):
LGTM!

jdkandersson (approved):

src/charm.py
yhaliaw: Ok. It should be fix once that PR is merged. Thanks.

PR #79, opened 2023-07-06
Add integration test

yhaliaw:
Add basic integration tests.

More tests will be added in a future PR.


tests/conftest.py
```diff
@@ -0,0 +1,18 @@
...
+"""Fixtures for github runner charm."""
+
+
+import argparse
+
+
+def pytest_addoption(parser: argparse.ArgumentParser):
+    """Add options to pytest parser."""
+    # mypy cannot find the `addoption` attr.
+    parser.addoption("--path", action="store")  # type: ignore
```
amandahla, line 13: Shouldnt "from pytest import Parser" be used?

tests/integration/test_charm.py
```diff
@@ -1,26 +1,159 @@
...
+
+    # The old runner should be flushed out.
+    action = await app.units[0].run_action("check-runners")
+    await action.wait()
+
+    assert action.results["online"] == "0"
+    assert action.results["offline"] == "0"
+    assert action.results["unknown"] == "0"
 
+    action = await app.units[0].run("cat /etc/systemd/system/repo-policy-compliance.service")
```
amandahla, line 156: I was about to comment that recently @jdkandersson added a review on a Synapse PR about checking if the action succeed besides the result but then I realized that this is being done here, right? I was just confused by the name of the variable "action".

tests/integration/conftest.py
```diff
@@ -1,32 +1,133 @@
...
+def path(pytestconfig: pytest.Config) -> str:
+    path = pytestconfig.getoption("--path")
+    assert path is not None, "Please specify the --path command line option"
+    return path
+
+
+@pytest.fixture(scope="module")
+def token_one(pytestconfig: pytest.Config) -> str:
+    token = pytestconfig.getoption("--token-one")
+    assert token is not None, "Please specify the --token-one command line option"
```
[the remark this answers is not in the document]
yhaliaw: Good suggestion, since empty string is never a valid token.

tests/conftest.py
yhaliaw: Updated

tests/integration/conftest.py
```diff
@@ -1,32 +1,133 @@
...
+    kmsg:
+        path: /dev/kmsg
+        source: /dev/kmsg
+        type: unix-char
+"""
+        )
+
+    charm = await ops_test.build_charm(".")
+
+    os.remove(lxd_profile_path)
```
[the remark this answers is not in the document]
yhaliaw: The e2e test uses a version of charm that running in containers. If this file exists while `charmcraft pack` the charm built would be the version used for testing.

The integration test will be using the same version of the charm, and should clean up this file after building the test version of the charm.

tests/integration/test_charm.py
yhaliaw: I added checking of `action.status == "completed"` to each instance of `juju.action.Action` object usage.

jdkandersson: The point is more that we should get a second opinion on whether the action succeeded, e.g., in this case check that there are no more lxd machines or something like that

tox.ini
```diff
@@ -105,6 +105,7 @@ deps =
     juju
     pytest-operator
     pytest-asyncio
+    pytest-dependency
```
jdkandersson, line 108: We should discuss this one

.github/workflows/e2e_test.yaml
```diff
@@ -2,6 +2,7 @@ name: End-to-End Test
 
 on:
   pull_request:
+  workflow_call:
```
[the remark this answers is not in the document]
yhaliaw: This is removed

tests/integration/test_charm.py
```diff
@@ -1,26 +1,169 @@
...
+@pytest.mark.dependency()
+async def test_missing_config(model: Model, app: Application) -> None:
+    """
+    arrange: Deploy an application without token configuration.
+    act: Check the status the application.
+    assert: The application is in blocked status.
+    """
+    await model.wait_for_idle()
+    # mypy can not find type of `name` attribute.
+    assert app.status == BlockedStatus.name  # type: ignore
```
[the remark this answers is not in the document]
benhoyt: Indeed. This should do the trick: https://github.com/canonical/operator/pull/980

.github/workflows/integration_test.yaml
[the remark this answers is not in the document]
yhaliaw: Due to the need of passing GitHub secrets to the integration test, it seems `operator-workflows` cannot be used right now.
E.g.: https://github.com/canonical/wordpress-k8s-operator/blob/ad73df54f65dfd8648d4273390108e1dd5d19a8c/.github/workflows/integration_test.yaml#L26

src/charm.py
```diff
@@ -172,6 +169,9 @@ def _create_memory_storage(self, path: Path, size: int) -> None:
         Raises:
             RunnerError: Unable to setup storage for runner.
         """
+        if size <= 0:
+            return
+
```
yanksyoon, line 174: May I ask if this should be an exception? Or maybe a validation error upstream since size being 0 or less seems like not a sensible value. :D

.github/workflows/integration_test.yaml
```diff
@@ -0,0 +1,25 @@
...
+          provider: lxd
+      - name: Enable br_netfilter
+        run: sudo modprobe br_netfilter
+      - name: Run Integration tests
+        run: |
+          tox -e integration -- \
+            --keep-models \
+            --path ${{secrets.E2E_TESTING_REPO}} \
+            --token-one ${{secrets.E2E_TESTING_TOKEN}} \
+            --token-two ${{secrets.E2E_TESTING_TOKEN_ALT}
```
weiiwang01, line 25: One of the `}` is missing in the end?

tests/integration/conftest.py
```diff
@@ -1,32 +1,151 @@
...
-@pytest.fixture
-def application(model, metadata):
-    charm_name = metadata["name"]
-    app = model.applications[charm_name]
-    return app
+@pytest_asyncio.fixture(scope="module")
+async def lxd_profile() -> AsyncIterator[Path]:
+    lxd_profile_path = Path("lxd-profile.yaml")
+
+    with open(lxd_profile_path, "w") as profile_file:
```
yanksyoon, line 83: Minor nitpick, I think tempfile could also be used so that the following unlink can be removed.

tests/integration/conftest.py
```diff
@@ -1,32 +1,151 @@
...
+    ops_test: OpsTest,
+    model: Model,
+    app_name: str,
+    path: str,
+    http_proxy: str,
+    https_proxy: str,
+    no_proxy: str,
+    lxd_profile: Path,
+) -> AsyncIterator[Application]:
+    charm = await ops_test.build_charm(".")
```
yanksyoon, line 117: I think the new operator workflows injects the built charm via [`--charm-file` argument](https://github.com/canonical/operator-workflows/blob/33a71e5b6fccdc9af17cd4cd4c1a92fdf3171e51/.github/workflows/integration_test_run.yaml#L226) which can be used to reduce a bit of build time :D

tests/integration/conftest.py
```diff
@@ -78,16 +77,9 @@ def model(ops_test: OpsTest) -> Model:
...
-    app_name: str,
-    path: str,
-    http_proxy: str,
-    https_proxy: str,
-    no_proxy: str,
-) -> AsyncIterator[Application]:
+async def lxd_profile() -> AsyncIterator[Path]:
     lxd_profile_path = Path("lxd-profile.yaml")
+
     with open(lxd_profile_path, "w") as profile_file:
```
jdkandersson, line 83: What about `lxd_profile_path.write_text` https://docs.python.org/3/library/pathlib.html#pathlib.Path.write_text

yanksyoon (dismissed):
LGTM, minor comments, thank you!

src/charm.py
yhaliaw: If there is 0 runners, the allocation size will be 0.
But there should be not case where size is under 0.

tests/integration/conftest.py
yhaliaw: This file needs to be have this exact name under this path. I think it is not possible with the `tempfile` module.

tests/integration/conftest.py
yhaliaw: `operator-workflow` is not used as GitHub secrets needs to be passed to the charm.

weiiwang01 (approved):

jdkandersson (approved):

yanksyoon (approved):
Thank you for the explanation, LGTM!

PR #81, opened 2023-07-07
Add workflow_dispatch github event handling and upgrade rpc

devspyrosv:
Requirements: 
 - A self-hosted runner should handle the workflow_dispatch event.
 - The repo-policy-compliance (rpc for short) should be updated to the latest version on charm upgrade.

This PR modifies pre-job.j2 allowing it to differentiate between different github events, handles workflow_dispatch and logs useful variables to syslog.
Also the repo-policy-compliance package is upgraded on upgrade_charm.

### DIFF
--- a/.github/workflows/workflow_dispatch_test.yaml
+++ b/.github/workflows/workflow_dispatch_test.yaml
@@ -1,0 +1,15 @@
+name: Workflow Dispatch Tests
+
+on:
+  # Manually dispatched workflow action
+  workflow_dispatch:
+    inputs:
+      runner:
+        description: 'Self hosted gh runner'
+        required: true
+jobs:
+  workflow-dispatch-tests:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Run a one-line script
+        run: echo Hello, world of workflow dispatches!
--- a/src/charm.py
+++ b/src/charm.py
@@ -576,6 +576,7 @@ class GithubRunnerCharm(CharmBase):
         """Install dependencies."""
         logger.info("Installing charm dependencies.")
 
+        # Snap and Apt will use any proxies configured in the Juju model.
         # Binding for snap, apt, and lxd init commands are not available so subprocess.run used.
         execute_command(["/usr/bin/apt-get", "update"])
         # install dependencies used by repo-policy-compliance and the firewall
@@ -583,7 +584,7 @@ class GithubRunnerCharm(CharmBase):
             ["/usr/bin/apt-get", "install", "-qy", "gunicorn", "python3-pip", "nftables"]
         )
 
-        # Install repo-policy-compliance package.
+        # Prepare environment for pip subprocess
         env = {}
         if "http" in self.proxies:
             env["HTTP_PROXY"] = self.proxies["http"]
@@ -595,6 +596,7 @@ class GithubRunnerCharm(CharmBase):
             env["NO_PROXY"] = self.proxies["no_proxy"]
             env["no_proxy"] = self.proxies["no_proxy"]
 
+        # Install repo-policy-compliance package
         execute_command(
             [
                 "/usr/bin/pip",
--- a/src/charm.py
+++ b/src/charm.py
@@ -332,7 +332,7 @@ class GithubRunnerCharm(CharmBase):
             event: Event of charm upgrade.
         """
         logger.info("Reinstalling dependencies...")
-        self._install_deps(upgrade_compliance=True)
+        self._install_deps()
         self._start_services()
         self._refresh_firewall()
 
@@ -572,14 +572,8 @@ class GithubRunnerCharm(CharmBase):
             return {"delta": {"virtual-machines": 0}}
 
     @retry(tries=10, delay=15, max_delay=60, backoff=1.5, local_logger=logger)
-    def _install_deps(self, upgrade_compliance: bool = False) -> None:
-        """Install dependencies.
-
-        Args:
-            upgrade_compliance: Upgrade (True) or install (False)
-                the repo-policy-compliance package.
-
-        """
+    def _install_deps(self) -> None:
+        """Install dependencies."""
         logger.info("Installing charm dependencies.")
 
         # Binding for snap, apt, and lxd init commands are not available so subprocess.run used.
@@ -601,17 +595,13 @@ class GithubRunnerCharm(CharmBase):
             env["NO_PROXY"] = self.proxies["no_proxy"]
             env["no_proxy"] = self.proxies["no_proxy"]
 
-        command = [
-            "/usr/bin/pip",
-            "install",
-            "git+https://github.com/canonical/repo-policy-compliance@main",
-        ]
-
-        if upgrade_compliance is True:
-            command.insert(2, "--upgrade")
-
         execute_command(
-            command,
+            [
+                "/usr/bin/pip",
+                "install",
+                "--upgrade",
+                "git+https://github.com/canonical/repo-policy-compliance@main",
+            ],
             env=env,
         )
 
--- a/src/charm.py
+++ b/src/charm.py
@@ -332,7 +332,7 @@ class GithubRunnerCharm(CharmBase):
             event: Event of charm upgrade.
         """
         logger.info("Reinstalling dependencies...")
-        self._install_deps()
+        self._install_deps(upgrade_compliance=True)
         self._start_services()
         self._refresh_firewall()
 
@@ -572,11 +572,24 @@ class GithubRunnerCharm(CharmBase):
             return {"delta": {"virtual-machines": 0}}
 
     @retry(tries=10, delay=15, max_delay=60, backoff=1.5, local_logger=logger)
-    def _install_deps(self) -> None:
-        """Install dependencies."""
+    def _install_deps(self, upgrade_compliance: bool = False) -> None:
+        """Install dependencies.
+
+        Args:
+            upgrade_compliance: Upgrade (True) or install (False)
+                the repo-policy-compliance package.
+
+        """
         logger.info("Installing charm dependencies.")
 
         # Binding for snap, apt, and lxd init commands are not available so subprocess.run used.
+        execute_command(["/usr/bin/apt-get", "update"])
+        # install dependencies used by repo-policy-compliance and the firewall
+        execute_command(
+            ["/usr/bin/apt-get", "install", "-qy", "gunicorn", "python3-pip", "nftables"]
+        )
+
+        # Install repo-policy-compliance package.
         env = {}
         if "http" in self.proxies:
             env["HTTP_PROXY"] = self.proxies["http"]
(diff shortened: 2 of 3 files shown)


devspyrosv:
Maybe this is relevant too https://github.com/canonical/github-runner-operator/pull/82

src/charm.py
```diff
@@ -572,11 +572,24 @@ def _reconcile_runners(self, runner_manager: RunnerManager) -> Dict[str, Any]:
...
         logger.info("Installing charm dependencies.")
 
         # Binding for snap, apt, and lxd init commands are not available so subprocess.run used.
+        execute_command(["/usr/bin/apt-get", "update"])
+        # install dependencies used by repo-policy-compliance and the firewall
+        execute_command(
+            ["/usr/bin/apt-get", "install", "-qy", "gunicorn", "python3-pip", "nftables"]
+        )
+
+        # Install repo-policy-compliance package.
```
weiiwang01, line 592: It's possible that relocating the installation process before configuring the proxy environment variables could cause some issues.

devspyrosv: Got it but isn't 
```
execute_command(
   ["/usr/bin/apt-get", "install", "-qy", "gunicorn", "python3-pip", "nftables"]
)
``` 
ignoring/not using the `env` though?

I think only having it under the `env` definition and not adding the `env` as a kwarg to the command does not allow the command to access the `env`. Right?

weiiwang01: I think the `execute_command` uses `subprocess.run` internally, and `subprocess.run` by default inherits the parent process's environment variables. So setting the `http_proxy` environment in the charm process (parent process) will affect the environment variables of the child process (`apt-get`).

jdkandersson: I also think that apt-get doesn't require the proxy

src/charm.py
```diff
@@ -588,17 +601,17 @@ def _install_deps(self) -> None:
...
-            ["/usr/bin/apt-get", "install", "-qy", "gunicorn", "python3-pip", "nftables"]
-        )
+        command = [
+            "/usr/bin/pip",
+            "install",
+            "git+https://github.com/canonical/repo-policy-compliance@main",
+        ]
+
+        if upgrade_compliance is True:
+            command.insert(2, "--upgrade")
```
jdkandersson, line 611: Perhaps just always upgrade?

templates/pre-job.j2
jdkandersson, line 1: Perhaps this is better as a follow up PR, as this is growing in complexity, should this be a Python script as well? conditionality etc is much easier in a Python script

templates/pre-job.j2
```diff
@@ -1,11 +1,44 @@
...
+  curl --noproxy '*' \
+      --fail-with-body \
+      -H 'Authorization: Bearer {{one_time_token}}' \
+      -H 'Content-Type: application/json' \
+      -d "{\"repository_name\": \"${GITHUB_REPOSITORY}\", \"source_repository_name\": \"${GITHUB_SOURCE_REPOSITORY}\", \"target_branch_name\": \"${GITHUB_BASE_REF}\", \"source_branch_name\": \"${GITHUB_HEAD_REF}\", \"commit_sha\": \"${GITHUB_SHA}\"}" \
+      http://{{host_ip}}:8080/check-run
+
+else
+
+  logger -p user.error -s "${GITHUB_EVENT_NAME} is not handled."
```
jdkandersson, line 40: Perhaps `is not supported yet, please request it to be added on <link to create issue on the GitHub runner repo>`?

jdkandersson (commented):
Nice work!

jdkandersson:
I didn't see the triggering of the workflow dispatch during the e2e tests?

src/charm.py
weiiwang01: @devspyrosv found out that the code didn't set the environment variables of the charm process, instead, it just create an `env` variable and passed to the `execute_command` function. So my statement is wrong in the last comment. 

I think the `env` variable in missing in the `execute_command(["/usr/bin/apt-get", "install", "-qy", "gunicorn", "python3-pip", "nftables"])`, since `apt-get` requires http proxies as well. Would you please add that, thanks!

devspyrosv:
> I didn't see the triggering of the workflow dispatch during the e2e tests?

Yes it's not triggered as I was not sure if it should be added there. On it!

src/charm.py
weiiwang01: > I also think that apt-get doesn't require the proxy

It doesn't require proxy in our production and staging environment, but it does in a general environment.

templates/pre-job.j2
devspyrosv: Yes, I agree, a follow up PR that moves pre-job from bash to python might indeed make everything more clear. 

In the meantime while the `GITHUB_EVENT_NAME` is the only conditional used, it can be considered ok-ish to use bash, I think.

src/charm.py
devspyrosv: Added a commit that is always adding the `--upgrade` option.

src/charm.py
devspyrosv: Preparing commit that passes env to all the `apt` and `pip` commands in that method. Wouldn't `snap` commands below them also need this?

gregory-schiano: Note that Juju has specific model config for APT proxies:
https://juju.is/docs/juju/offline-mode-strategies#heading--apt-proxy
https://juju.is/docs/juju/configure-juju-for-offline-usage#heading--offline-configuration-methods

templates/pre-job.j2
```diff
@@ -1,11 +1,44 @@
...
+  GITHUB_BASE_REF: ${GITHUB_BASE_REF}, \
+  GITHUB_HEAD_REF: ${GITHUB_HEAD_REF}, \
+  GITHUB_SHA: ${GITHUB_SHA}"
+
+  curl --noproxy '*' \
+      --fail-with-body \
+      -H 'Authorization: Bearer {{one_time_token}}' \
+      -H 'Content-Type: application/json' \
+      -d "{\"repository_name\": \"${GITHUB_REPOSITORY}\", \"source_repository_name\": \"${GITHUB_SOURCE_REPOSITORY}\", \"target_branch_name\": \"${GITHUB_BASE_REF}\", \"source_branch_name\": \"${GITHUB_HEAD_REF}\", \"commit_sha\": \"${GITHUB_SHA}\"}" \
+      http://{{host_ip}}:8080/check-run
```
gregory-schiano, line 36: Preferably use `/pull_request/check-run` as `/check-run` is kept for backward compatibility and should be removed

templates/pre-job.j2
```diff
@@ -1,11 +1,44 @@
...
-    -H 'Content-Type: application/json' \
-    -d "{\"repository_name\": \"${GITHUB_REPOSITORY}\", \"source_repository_name\": \"${GITHUB_SOURCE_REPOSITORY}\", \"target_branch_name\": \"${GITHUB_BASE_REF}\", \"source_branch_name\": \"${GITHUB_HEAD_REF}\", \"commit_sha\": \"${GITHUB_SHA}\"}" \
-    http://{{host_ip}}:8080/check-run
+# Workflow dispatch - Request repo-policy-compliance service check:
+if [[ "${GITHUB_EVENT_NAME}" == "workflow_dispatch" ]]; then
+
+  logger -s "GITHUB_EVENT_NAME: ${GITHUB_EVENT_NAME}, \
+  GITHUB_REPOSITORY: ${GITHUB_REPOSITORY}, \
+  GITHUB_REF_NAME: ${GITHUB_REF_NAME}, \
+  GITHUB_SHA: ${GITHUB_SHA}"
```
gregory-schiano, line 9: I think that `GITHUB_EVENT_NAME`, `GITHUB_REPOSITORY` and `GITHUB_SHA` can be logged outside the if to avoid duplication.
That will also log it for non "supported" GH events, but it's not such a big deal

templates/pre-job.j2
```diff
@@ -1,11 +1,44 @@
...
+
+  curl --noproxy '*' \
+      --fail-with-body \
+      -H 'Authorization: Bearer {{one_time_token}}' \
+      -H 'Content-Type: application/json' \
```
gregory-schiano, line 14: You should probably put these arguments in a variable outside the if, and use it in both curl calls to avoid duplication

templates/pre-job.j2
devspyrosv: I saw that but was not sure if it was kept back for other reasons. This was the pre-existing command.

devspyrosv: fixed

templates/pre-job.j2
devspyrosv: Yes, indeed it's a nice improvement. Added it in a commit.

templates/pre-job.j2
devspyrosv: Added a the common args in a variable. --needs-testing

src/charm.py
devspyrosv: @gregory-schiano Do you mean that environment vars like `http_proxy` and `no_proxy` should not be used inside the charm code and that proxying should be handled at the model level? Could you please elaborate a bit on the above?
Thx!

devspyrosv: I also think that `apt` will not honor the `http[s]_proxy` vars. Doesn't it need a modification in an etc conf file?

So I think that proxying in general needs rework and a discussion before it's fixed/improved. @jdkandersson should it be part of this PR?

devspyrosv: The Juju model can be configured to make units use proxies for apt and snap.

I believe we can consider this discussion complete.

gregory-schiano (dismissed):
LGTM

.github/workflows/workflow_dispatch_test.yaml
jdkandersson, line 1: I'll put a comment here just as a note, please modify the e2e tests to trigger workflow dispatch and check that it worked

templates/pre-job.j2
```diff
@@ -1,11 +1,44 @@
...
+--fail-with-body \
+-H 'Authorization: Bearer {{one_time_token}}' \
+-H 'Content-Type: application/json'"
+
+# Workflow dispatch - Request repo-policy-compliance service check:
+if [[ "${GITHUB_EVENT_NAME}" == "workflow_dispatch" ]]; then
+
+  logger -s "GITHUB_REF_NAME: ${GITHUB_REF_NAME}"
+
+  curl "${CURL_AUTH_ARGS}" \
```
amandahla, line 19: Since this is being changed, what do you think about adding a timeout parameter to curl (-m)?

devspyrosv: It would be good to add it. Thank you!

amandahla (dismissed):
LGTM :-)

.github/workflows/workflow_dispatch_test.yaml
weiiwang01: Hi, I just talked with David, would you like to merge the current workflow_dispatch feature as it has been tested manually already. And I will revive an old discarded e2e test that include the workflow_dispatch test in another pull request. In that case we don't need to repeat implementing the same test twice.

devspyrosv: Ok. I will clean up the commits first. I will remove those I used to attempt to make the tests work.

devspyrosv: @weiiwang01 done!

devspyrosv:
This PR was "moved" to #84 as it could not use the secrets when running the workflows. 

The reason was that this was an external (read: from my own account) one. 

Closing

PR #88, opened 2023-07-12
Improve detection of runner health

yhaliaw:
Uses whether the runner script process is  running as the health check for the runner.


gregory-schiano (dismissed):
LGTM

src/utilities.py
```diff
@@ -105,12 +107,17 @@ def secure_run_subprocess(cmd: Sequence[str], **kwargs) -> subprocess.CompletedP
...
     Args:
         cmd: Command in a list.
+        hide_cmd: Hide logging of cmd.
         kwargs: Additional keyword arguments for the `subprocess.run` call.
 
     Returns:
         Object representing the completed process. The outputs subprocess can accessed.
     """
-    logger.info("Executing command %s", cmd)
+    if not hide_cmd:
```
amandahla, line 116: I think I'm missing something here. Is the hide_cmd parameter used only to show the log message that a sensitive command is being executed?

yhaliaw: Yes, current usage of `hide_cmd` is too hide sensitive commands.
It can also be used for hiding overly long commands, or any other reasons.

src/runner_manager.py
```diff
@@ -398,11 +401,14 @@ def _get_runner_github_info(self) -> Dict[str, SelfHostedRunner]:
         remote_runners_list: list[SelfHostedRunner] = []
         if isinstance(self.config.path, GitHubRepo):
             remote_runners_list = self._clients.github.actions.list_self_hosted_runners_for_repo(
-                owner=self.config.path.owner, repo=self.config.path.repo
+                owner=self.config.path.owner,
+                repo=self.config.path.repo,
+                per_page=100,
```
jdkandersson, line 406: Aren't we still limited to 100 runners? We're planning to go to 162 pretty soon

yhaliaw: The `per_page` has a 100 max size.
Should we migrate to PyGitHub?

jdkandersson: Is this just for logging or does it have a functional impact?

yhaliaw: Will update to use `paged` generator provided by ghapi to handle pagination.

jdkandersson (approved):

yanksyoon (approved):
LGTM!

PR #91, opened 2023-07-24
Fix GitHub token config update

yhaliaw:
When the github token config is updated, the token used by policy-service-compliance is not updated. This PR fixes this issue.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -130,6 +130,7 @@ class GithubRunnerCharm(CharmBase):
 
         self._stored.set_default(
             path=self.config["path"],  # for detecting changes
+            token=self.config["token"],  # for detecting changes
             runner_bin_url=None,
         )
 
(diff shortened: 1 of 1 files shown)


jdkandersson (dismissed):

src/charm.py
```diff
@@ -674,7 +674,8 @@ def _start_services(self) -> None:
         )
         self.repo_check_systemd_service.write_text(service_content, encoding="utf-8")
 
-        execute_command(["/usr/bin/systemctl", "start", "repo-policy-compliance"])
+        execute_command(["/usr/bin/systemctl", "daemon-reload"])
+        execute_command(["/usr/bin/systemctl", "restart", "repo-policy-compliance"])
```
jdkandersson, line 678: I think this means that any existing tokens for spawned runners will be invalid, so we need to make sure to remove existing runners as well (perhaps this is already being done as a part of changing the token)

jdkandersson (dismissed):

jdkandersson (dismissed):

jdkandersson (approved):

arturo-seijas (approved):
LGTM

mthaddon (commented):
We should really avoid using stored state for this (or anything else if possible). Is there some other way we can determine if the token has changed by inspecting the live state of the service, for instance?

yhaliaw:
> We should really avoid using stored state for this (or anything else if possible). Is there some other way we can determine if the token has changed by inspecting the live state of the service, for instance?

I think this is planned as a part of the charm review and refactor.

PR #92, opened 2023-07-27
Add integration test for charm configuration

yhaliaw:


tests/integration/helpers.py
```diff
@@ -0,0 +1,88 @@
...
+    await action.wait()
+    assert action.results["return-code"] == 0
+    profile_content = yaml.safe_load(action.results["stdout"])
+    assert f"{cpu}" == profile_content["config"]["limits.cpu"]
+    assert mem == profile_content["config"]["limits.memory"]
+    assert disk == profile_content["devices"]["root"]["size"]
+
+
+@retry(tries=30, delay=30)
+async def check_runner_instance(unit: Unit, num: int) -> None:
```
yanksyoon, line 63: Minor suggestion, would this be better named something like
`assert_n_runners` so that it's clear that there's an assert for the tests?

yanksyoon (dismissed):
LGTM!

tests/integration/helpers.py
yhaliaw: I have updated the names of these helper function that make assertions.

arturo-seijas (approved):
LGTM

yanksyoon (approved):
LGTM!

PR #93, opened 2023-08-01
Auto-update repo-policy-compliance service

yhaliaw:
* Modify the update-runner-binary event to also update the repo-policy-compliance service.
* Add `python` as alias for `python3`


src/charm.py
```diff
@@ -427,10 +427,11 @@ def _check_and_update_dependencies(self) -> bool:
             self.unit.status = MaintenanceStatus(f"Failed to check for runner updates: {err}")
             return False
 
+        # TODO: Remove debugging.
+        logger.info("%s %s", runner_info.download_url, self._stored.runner_bin_url)
```
merkata: Would it be possible to leave debugging on and only trigger it conditionally, I recall bitbucket pipelines let's you define variables on runs, so you might pass debug=on and it defaults to debug=off for instance..

yhaliaw: This is converted to debug-level logging.

src/charm.py
```diff
@@ -606,7 +651,15 @@ def _install_deps(self) -> None:
             env["NO_PROXY"] = self.proxies["no_proxy"]
             env["no_proxy"] = self.proxies["no_proxy"]
 
-        # Install repo-policy-compliance package
+        old_version = execute_command(
+            [
+                "/usr/bin/pip",
```
merkata, line 656: The check for pip in the GH actions invokes `python3 -m pip`, perhaps it can be invoked like this here as well?

src/runner.py
```diff
@@ -536,6 +536,9 @@ def _configure_runner(self) -> None:
                 ["/usr/bin/chown", "-R", "ubuntu:ubuntu", "/home/ubuntu/.docker"]
             )
 
+        # Make python an alias of python3
+        self.instance.execute(["ln", "-s", "/usr/bin/python3", "/usr/bin/python"])
```
merkata, line 540: If `/usr/bin/python` exists this command will fail, is this expected?

tests/integration/helpers.py
```diff
@@ -14,8 +14,43 @@
...
 
-async def remove_runner_bin(unit: Unit):
-    """Remove runner binary."""
+async def get_repo_policy_compliance_pip_info(unit: Unit) -> None | str:
+    """Get pip info for repo-policy-compliance.
+
+    Args:
+        source: The git source to install the package. If none the package is removed.
+    """
+    action = await unit.run("pip show repo-policy-compliance")
```
merkata, line 23: Similar to above comments, GH action invokes `python3 -m pip`, the charm invokes `pip` with a full path, here it relies on pip being in `$PATH` - perhaps there can be one way of invocation?

merkata (dismissed):
Looks good, left some minor nit picks.

src/runner.py
yhaliaw: This is run on start a runner LXD instance. Since the LXD instance is created from a ubuntu image, `/usr/bin/python` should be empty.

src/charm.py
yhaliaw: Updated the code.

tests/integration/helpers.py
yhaliaw: Switched to `python3 -m pip`

src/runner.py
```diff
@@ -536,6 +536,10 @@ def _configure_runner(self) -> None:
                 ["/usr/bin/chown", "-R", "ubuntu:ubuntu", "/home/ubuntu/.docker"]
             )
 
+        # Make python an alias of python3
+        self.instance.execute(["rm", "/usr/bin/python3"])
```
jdkandersson, line 540: Should this `rm` `/usr/bin/python` rather than `/usr/bin/python3`?

jdkandersson (approved):

src/runner.py
yhaliaw: Thanks this is fixed.

weiiwang01 (approved):

PR #95, opened 2023-08-09
Add integration test for repo-policy-compliance check failure 

yhaliaw:
Added test for repo-policy-compliance failure on job start. This is to ensure the runner correctly stops jobs that does not pass repo-policy-compliance check.

The end-to-end test will check for success case of the repo-policy-compliance check.

### DIFF
--- a/.github/workflows/integration_tests.yaml
+++ b/.github/workflows/integration_tests.yaml
@@ -75,6 +75,6 @@ jobs:
     timeout-minutes: 30
     steps:
       - run: |
-          [ '${{ needs.integration-test-charm-no-runner.result }}' = 'success' ] &&
-          [ '${{ needs.integration-test-charm.result }}' = 'success' ] &&
-          [ '${{ needs.integration-test-repo-policy-compliance.result }}' = 'success' ] || (echo integration-test failed && false)
+          [ '${{ needs.integration-test-charm-no-runner.result }}' = 'success' ] || (echo integration-test-charm-no-runner failed && false)
+          [ '${{ needs.integration-test-charm.result }}' = 'success' ] || (echo integration-test-charm failed && false)
+          [ '${{ needs.integration-test-repo-policy-compliance.result }}' = 'success' ] || (echo integration-test-repo-policy-compliance failed && false)
--- a/.github/workflows/integration_tests.yaml
+++ b/.github/workflows/integration_tests.yaml
@@ -1,4 +1,4 @@
-name: Integration test
+name: Integration tests
 
 on:
   pull_request:
--- a/tests/integration/helpers.py
+++ b/tests/integration/helpers.py
@@ -18,7 +18,7 @@ async def get_repo_policy_compliance_pip_info(unit: Unit) -> None | str:
     """Get pip info for repo-policy-compliance.
 
     Args:
-        source: The git source to install the package. If none the package is removed.
+        unit: Unit instance to check for the LXD profile.
     """
     action = await unit.run("python3 -m pip show repo-policy-compliance")
     await action.wait()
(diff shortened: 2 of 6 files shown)


jdkandersson (approved):
Looks good!

tests/integration/helpers.py
```diff
@@ -94,21 +95,38 @@ async def assert_resource_lxd_profile(unit: Unit, configs: dict[str, Any]) -> No
...
+    Args:
+        unit: Unit instance to check for the LXD profile.
+
+    Returns:
+        Tuple of runner names.
+    """
+    action = await unit.run("lxc list --format json")
+    await action.wait()
+
+    assert action.results["return-code"] == 0
```
amandahla, line 110: Just something to mind:
https://github.com/canonical/hardware-observer-operator/issues/36#issue-1795890406

```
# with juju 2
action = await unit.run(command, timeout=timeout)
results = action.results
return results # {'Code': '3', 'Stdout': 'inactive\n'}

# with juju 3
action = await unit.run(command, timeout=timeout)
await action.wait()
results = action.results
return results # {'return-code': 3, 'stdout': 'inactive\n'}

File: tests/integration/helpers.py [hunk 2]
```diff
@@ -94,21 +95,38 @@ async def assert_resource_lxd_profile(unit: Unit, configs: dict[str, Any]) -> No
...
+    """Get names of the runners in LXD.
+
+    Args:
+        unit: Unit instance to check for the LXD profile.
+
+    Returns:
+        Tuple of runner names.
+    """
+    action = await unit.run("lxc list --format json")
+    await action.wait()
```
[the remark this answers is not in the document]
yhaliaw: As per your comment below, this is needed for juju 3.

tests/integration/helpers.py
yhaliaw: Good catch.
The integration test currently uses juju 3.1.
Since it is likely we need to switch to juju 3.X in the future as juju 2.X goes out of support at some point, I think the current code is fine.

.github/workflows/integration_tests.yaml
```diff
@@ -0,0 +1,80 @@
...
+      - name: Enable br_netfilter
+        run: sudo modprobe br_netfilter
+      - name: Run Integration tests
+        run: |
+          tox -e integration -- \
+            --keep-models \
+            --path ${{secrets.E2E_TESTING_REPO}} \
+            --token ${{secrets.E2E_TESTING_TOKEN}} \
+            --token-alt ${{secrets.E2E_TESTING_TOKEN_ALT}} \
+            tests/integration/test_charm_no_runner.py
```
yanksyoon, line 26: Minor nitpick no need change: would this be better if it was `tox -e integration -- -k test_charm_no_runner`?

yanksyoon: Same for other tests

tests/integration/test_repo_policy_compliance.py
```diff
@@ -0,0 +1,182 @@
...
+
+    await app_no_runner.set_config(
+        {"virtual-machines": "1", "path": forked_github_repository.full_name}
+    )
+    action = await unit.run_action("reconcile-runners")
+    await action.wait()
+    await model.wait_for_idle(status=ACTIVE_STATUS_NAME)
+
+    # Wait until there is one runner.
+    await assesrt_num_of_runners(unit, 1)
```
yanksyoon, line 138: I think there is a typo here :D `assesrt` -> `assert`

yanksyoon (approved):
LGTM, minor nitpicks!

.github/workflows/integration_tests.yaml
yhaliaw: I feel using the filename should be better as it would be an exact match. Using patterns has the risk of matching something unintended.

tests/integration/test_repo_policy_compliance.py
yhaliaw: Good catch

yhaliaw: This typo is already in the main branch. I will fix this in the next PR.

PR #105, opened 2023-08-22
Upgrade kernel on charm upgrade

yhaliaw:
Applicable spec: NA

### Overview

Combine the update_dependencies event and the reconcile event.

Update to dependencies and kernel occurs with the reconcile event.

Reboot of the charm is done with `juju-reboot` command.

### Rationale

By combine the update dependencies/kernel with reconcile, the runner spawned should have the latest dependencies and kernel. This should increase the security as the latest version of the dependencies are used.

### Juju Events Changes

The update_dependencies custom event is removed.
The reconcile_runners custom event will update the dependencies and kernel as well.

The reconcile_runner_actions will update the dependencies but not the kernel. This is due to juju-reboot is not supported for juju actions:
```
$ juju help-tool juju-reboot
Usage: juju-reboot [options]

Summary:
Reboot the host machine

Options:
--now  (= false)
    reboot immediately, killing the invoking process

Details:
juju-reboot causes the host machine to reboot, after stopping all containers
	hosted on the machine.

	An invocation without arguments will allow the current hook to complete, and
	will only cause a reboot if the hook completes successfully.

	If the --now flag is passed, the current hook will terminate immediately, and
	be restarted from scratch after reboot. This allows charm authors to write
	hooks that need to reboot more than once in the course of installing software.

	The --now flag cannot terminate a debug-hooks session; hooks using --now should
	be sure to terminate on unexpected errors, so as to guarantee expected behaviour
	in all situations.

	juju-reboot is not supported when running actions.
```

### Checklist


src/charm.py
```diff
@@ -462,37 +478,24 @@ def _check_and_update_dependencies(self) -> bool:
...
         """
-        self.unit.status = MaintenanceStatus("Reconciling runners")
+        runner_manager = self._get_runner_manager()
 
-        if not RunnerManager.runner_bin_path.is_file():
-            logger.warning("Unable to reconcile due to missing runner binary")
-            return
+        runner_info = runner_manager.get_github_info()
+        if all(not info.busy for info in runner_info):
+            self._update_kernel(now=True)
```
jdkandersson, line 496: I'm guessing `juju-reboot` will only be triggered after the event handling is done? Otherwise, the check and apply updates should probably happen before the reboot

jdkandersson (dismissed):

src/charm.py
yhaliaw: I have update the code here.
The `juju-reboot` has two possible option. One is to reboot after the event handling is done, the other is to reboot immediately. If the `now` arg is true, then it is rebooted immediately.

jdkandersson (dismissed):

src/runner_manager.py
```diff
@@ -264,7 +265,10 @@ def get_github_info(self) -> Iterator[RunnerInfo]:
             List of information from GitHub on runners.
         """
         remote_runners = self._get_runner_github_info()
-        return iter(RunnerInfo(runner.name, runner.status) for runner in remote_runners.values())
+        return iter(
+            RunnerInfo(runner.name, runner.status, runner.busy)
+            for runner in remote_runners.values()
+        )
```
yanksyoon, line 271: May I ask what the `iter` wrapper is for please?

yanksyoon (dismissed):
LGTM, thx!!

src/runner_manager.py
jdkandersson: it should be possible to drop the `iter` and just return with `(...)` which creates a generator

src/runner_manager.py
```diff
@@ -258,14 +258,14 @@ def update_runner_bin(self, binary: RunnerApplication) -> None:
 
         logger.info("Validated newly downloaded runner binary and enabled it.")
 
-    def get_github_info(self) -> Iterator[RunnerInfo]:
+    def get_github_info(self) -> Generator[RunnerInfo, None, None]:
```
jdkandersson, line 261: I think `Iterator` is still a valid type, not 100% on that though. Up to you

jdkandersson (dismissed):

yanksyoon (dismissed):
LGTM!

jdkandersson (approved):

yanksyoon (approved):

PR #107, opened 2023-08-24
Allow traffic for docker user

yhaliaw:
Run command within self-hosted runner instance to allow docker user to access network.
Docker is installed within the self-hosted runner by default, however this configuration change was missing.

### DIFF
--- a/.github/workflows/e2e_test.yaml
+++ b/.github/workflows/e2e_test.yaml
@@ -22,7 +22,6 @@ jobs:
             security.nesting: true
             security.privileged: true
             raw.lxc: |
-              lxc.apparmor.profile=unconfined
               lxc.mount.auto=proc:rw sys:rw cgroup:rw
               lxc.cgroup.devices.allow=a
               lxc.cap.drop=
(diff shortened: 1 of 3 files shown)


weiiwang01 (dismissed):

arturo-seijas (dismissed):
LGTM

arturo-seijas (commented):
Can you change the PR title and description? Next time, consider opening different PRs so that the can be managed in a simpler way

merkata (dismissed):

.github/workflows/e2e_test.yaml
```diff
@@ -270,7 +270,9 @@ jobs:
           [[ -z "${http_proxy}" && -z "${HTTP_PROXY}" ]] \
             || cat /home/ubuntu/.docker/config.json | grep httpProxy
       - name: Install microk8s
-        run: sudo snap install microk8s --classic
+        # TMP: Pin version of microk8s to older version due to issues with running microk8s in LXD
```
merkata, line 273: Is this about running unconfined lxd containers or something else? I suggest adding a link to an issue or a clarifying comment what the problem is (although I think there should be an open issue we can link to).

amandahla:
Could add to the description why is this needed?

amandahla (approved):

weiiwang01 (approved):

PR #108, opened 2023-08-25
Automatically add LXD instance name to NO_PROXY

yhaliaw:
### Overview

Append needed no_proxy variable if http_proxy or https_proxy are used.

The appended variables are:
* The LXD instance name this is used by microk8s to refer to the localhost.
* `.svc` used by k8s/microk8s local services.
The above are added since they are specific to k8s/microk8s and the self-hosted runner are configured to support them if they are installed.

The following are not added as normally, these would be set by the user of the proxy.
* Local IPv4 address (`10.0.0.0/8,172.16.0.0/12,192.168.0.0/16`)
* Localhost address (`localhost,127.0.0.1,::1`)

### Rationale

This is needed for microk8s to run on the self-hosted runners.

### Checklist

- [ ] The documentation is generated using `src-docs`

The `src-docs` for this repo is still WIP. There is an WIP PR for it.

### DIFF
--- a/src/runner.py
+++ b/src/runner.py
@@ -94,8 +94,12 @@ class Runner:
         self.instance = instance
 
         # If the proxy setting are set, then add NO_PROXY local variables.
-        if self.config.proxies["no_proxy"]:
-            self.config.proxies["no_proxy"] += f",{self.config.name},.svc"
+        if self.config.proxies.get("http") or self.config.proxies.get("https"):
+            if self.config.proxies.get("no_proxy"):
+                self.config.proxies["no_proxy"] += ","
+            self.config.proxies[
+                "no_proxy"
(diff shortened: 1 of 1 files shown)


src/runner.py
```diff
@@ -93,6 +93,12 @@ def __init__(
         self.status = runner_status
         self.instance = instance
 
+        # If the proxy setting are set, then add NO_PROXY local variables.
+        if self.config.proxies.get("http") or self.config.proxies.get("https"):
```
arturo-seijas, line 97: Is it an issue if we add the no proxy variable even when the proxy is not configured?

arturo-seijas: Makes sense

arturo-seijas (approved):
LGTM

weiiwang01 (approved):

PR #112, opened 2023-08-30
Shutdown runner on repo-policy-compliance failure

yhaliaw:
### Overview

On repo-policy-compliance failure in pre-job script, wait a short while before shutdown the LXD instance. This will cause the runner to be lost and unable to execute the rest of the job.

The integration test with the repo-policy-compliance is changed to allow parallel tests, and a test case for this issue is added.

### Rationale

On pre-job script failure, the rest of the job can still be executed. See: https://github.com/actions/runner/issues/2811
An example would be: https://github.com/canonical/github-runner-operator/actions/runs/6019884975/job/16330344718
The pre-job script failed, but the `Run echo world` step still ran.

This PR intends to stop the runner application after the repo-policy-compliance check fails. When the runner application is  stopped the logs are uploaded to GitHub. If the LXD instance is directly poweroff the logs are lost, and no error message on the repo-policy-compliance can be found by the users.

A sample repo-policy-compliance check fail logs is as below:
https://github.com/canonical/github-runner-operator/actions/runs/6035086506/job/16374981798

### Checklist

- [ ] The documentation is generated using `src-docs`

The `src-docs` for this repo is still WIP. There is an WIP PR for it.


yhaliaw:
The following will sleep for 180 seconds before shutting down.
The logs of the pre-job step is still lost.
https://github.com/canonical/github-runner-operator/actions/runs/6022358785/job/16337198269

templates/pre-job.j2
```diff
@@ -22,6 +27,7 @@ if [[ "${GITHUB_EVENT_NAME}" == "workflow_dispatch" ]] || [[ "${GITHUB_EVENT_NAM
   curl "${CURL_ARGS[@]}" \
       -d "{\"repository_name\": \"${GITHUB_REPOSITORY}\", \"branch_name\": \"${GITHUB_REF_NAME}\", \"commit_sha\": \"${GITHUB_SHA}\"}" \
       http://{{host_ip}}:8080/${GITHUB_EVENT_NAME}/check-run
+  REPO_CHECK=$?
```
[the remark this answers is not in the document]
yhaliaw: I have thought about it, but I think that way makes the code more brittle. In the future, someone might add a new command at the end of one of the if-else block without realizing the exit code of the new command will be captured instead.

templates/pre-job.j2
```diff
@@ -36,11 +42,29 @@ elif [[ "${GITHUB_EVENT_NAME}" ==  "pull_request" ]]; then
...
+
+fi
+
+if [[ $REPO_CHECK -ne 0 ]]; then
+  echo "The repo-policy-compliance check failed with $REPO_CHECK exit code."
+  echo "Stopping execution of jobs due to repo-policy-compliance failure."
+  echo "This runner will be stopped or lost, fix the repo policy compliance then rerun this job."
+
+  # Killing the runner.Listener process to stop the runner application.
+  pkill -2 Runner.Listener
```
arturo-seijas, line 61: Do you want to send an interceptable signal for graceful termination?

yhaliaw: This is way recommended by someone working on GitHub Actions. https://github.com/actions/runner/issues/2190#issuecomment-1273302389

Is there any concerns with this approach?

arturo-seijas: No, not really. I was curious :)

arturo-seijas (dismissed):
LGTM

templates/pre-job.j2
```diff
@@ -36,11 +45,29 @@ elif [[ "${GITHUB_EVENT_NAME}" ==  "pull_request" ]]; then
...
 
   logger -p user.error -s "${GITHUB_EVENT_NAME} is not supported yet. Please request it to be added on https://github.com/canonical/github-runner-operator/issues/new/choose"
 
-  return 1
+  REPO_CHECK=1
 
 fi
+
+if [[ $REPO_CHECK -ne 0 ]]; then
+  echo "The repo-policy-compliance check failed with $REPO_CHECK exit code."
```
cbartz, line 59: In the other parts of the script, `logger` is used instead of `echo`. I think it might make sense to have a consistent approach throughout the script.

yhaliaw: Good catch. Let me update the script.

templates/pre-job.j2
```diff
@@ -36,11 +45,29 @@ elif [[ "${GITHUB_EVENT_NAME}" ==  "pull_request" ]]; then
...
 
-  return 1
+  REPO_CHECK=1
 
 fi
+
+if [[ $REPO_CHECK -ne 0 ]]; then
+  logger -p user.error -s "The repo-policy-compliance check failed with $REPO_CHECK exit code."
+  logger -p user.error -s "Stopping execution of jobs due to repo-policy-compliance failure."
+  logger -p user.error -s "This runner will be stopped or lost, fix the repo policy compliance then rerun this job."
```
jdkandersson, line 61: Perhaps something like `This runner will be stopped or lost, please fix the setup of the repository, then rerun this job.`

templates/pre-job.j2
```diff
@@ -36,11 +45,29 @@ elif [[ "${GITHUB_EVENT_NAME}" ==  "pull_request" ]]; then
...
+
+  # Killing the runner.Listener process to stop the runner application.
+  pkill -2 Runner.Listener
```
jdkandersson, line 64: Does this ensure that no further step from the job is processed?

templates/pre-job.j2
```diff
@@ -36,11 +45,29 @@ elif [[ "${GITHUB_EVENT_NAME}" ==  "pull_request" ]]; then
...
+
+  # Shutdown the instance as a safe guard.
+  bash -c "sleep 10; sudo systemctl poweroff -i" &
```
jdkandersson, line 67: Just to double check, does this give a 10 second window where a further step might get executed?

templates/pre-job.j2
```diff
@@ -36,11 +45,29 @@ elif [[ "${GITHUB_EVENT_NAME}" ==  "pull_request" ]]; then
...
+
+  exit 1
+
+fi
+
+logger -s "The repo-policy-compliance check has passed, proceeding to execute jobs"
```
jdkandersson, line 73: `The repository setup check has passed, proceeding to execute jobs` - generally we shouldn't refer to `repo-policy-compliance` since our users probably don't know what that means

templates/pre-job.j2
yhaliaw: In this job run, I have tested with `if: always()` and `if: failure()`:
https://github.com/canonical/github-runner-operator/actions/runs/6033194405/job/16369693342

Both of step was not executed.

templates/pre-job.j2
yhaliaw: The `pkill` should already prevent further step from getting executed (see above comment chain).

I have tested without the 10 sec delay, but the logs would not get uploaded.

templates/pre-job.j2
yhaliaw: I have updated the code comments and message for users.

templates/pre-job.j2
yhaliaw: Updated

jdkandersson (dismissed):

cbartz (dismissed):

jdkandersson (approved):

arturo-seijas (approved):

merkata (approved):

PR #117, opened 2023-09-08
Add support for schedule event

cbartz:
Applicable spec: n/a

### Overview

The GitHub schedule event handling is added to the pre-job script and an additional end-to-end test is added. 
The [schedule event](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#schedule) can only be triggered on the main branch, so to implement the end-to-end test, another repo is dynamically created in `secrets.E2E_TESTING_TOKEN_ORG` and the schedule event tests are run there.

### Rationale

A self-hosted runner should handle the schedule event.

### Checklist

### DIFF
--- a/.github/workflows/e2e_test.yaml
+++ b/.github/workflows/e2e_test.yaml
@@ -345,8 +345,8 @@ jobs:
               output=$(gh run list \
                         -R ${TESTING_REPO} \
                         -L 100 \
-                        --json headBranch,status \
-                        --jq '[.[] | select(.headBranch=="main")]')
+                        --json headBranch,status,createdAt \
+                        --jq '[.[] | select(.headBranch=="main")] | sort_by(.createdAt)'
 
               # Workflows that have not started have no status
               if [ $(echo "$output" | jq 'length') -eq 0 ]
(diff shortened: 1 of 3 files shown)


cbartz:
@yhaliaw , could you please have a look if you are happy with the schedule test approach. This would currently mean that a schedule workflow would run every 5 minutes.
 We may need an additional cleanup mechanism (deleting old workflow runs that did not run the did not trigger the schedule event test job due to lack of corresponding end-to-end tests branches).
 
An alternative approach would be to test the main branch only once a day or so, but then we would not be immediately alerted when a pr breaks schedule event support, which may be fine as we do not immediately promote releases to stable.

jdkandersson (dismissed):

yhaliaw:
> @yhaliaw , could you please have a look if you are happy with the schedule test approach. This would currently mean that a schedule workflow would run every 5 minutes. We may need an additional cleanup mechanism (deleting old workflow runs that did not run the did not trigger the schedule event test job due to lack of corresponding end-to-end tests branches).
> 
> An alternative approach would be to test the main branch only once a day or so, but then we would not be immediately alerted when a pr breaks schedule event support, which may be fine as we do not immediately promote releases to stable.

A test every 5 mins seems quite frequent. Is there a reason for it? Would a test every hour cause any problems?

cbartz:
> > @yhaliaw , could you please have a look if you are happy with the schedule test approach. This would currently mean that a schedule workflow would run every 5 minutes. We may need an additional cleanup mechanism (deleting old workflow runs that did not run the did not trigger the schedule event test job due to lack of corresponding end-to-end tests branches).
> > An alternative approach would be to test the main branch only once a day or so, but then we would not be immediately alerted when a pr breaks schedule event support, which may be fine as we do not immediately promote releases to stable.
> 
> A test every 5 mins seems quite frequent. Is there a reason for it? Would a test every hour cause any problems?

The end-to-end test (not included in this draft PR) triggered by a PR will have to wait for the schedule. So if we set it to one hour, the end-to-end test would take a very long time.

yhaliaw (dismissed):
LGTM

cbartz:
The end-to-end test is now implemented by using a different repo for the schedule test. The PR description has been updated.

jdkandersson (approved):

yhaliaw (approved):
LGTM

PR #122, opened 2023-09-14
Ensure kernel modules are loaded after reboot

yhaliaw:
Applicable spec: NA

### Overview

Ensure kernel modules are loaded after reboot.

### Rationale

The charm will reboots on upgrade to kernel or some dependencies. After reboot if the kernel modules are not loaded, it can cause crashes in LXD.

A deployment is running multiple jobs without any crashes:
https://github.com/canonical/github-runner-operator/actions/runs/6181781894
The failure from job 3 is due to unsigned commits on the `do-not-merge` branch as I am testing something else.

### Checklist

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -679,7 +679,7 @@ class GithubRunnerCharm(CharmBase):
         )
         return old_version != new_version
 
-    def enable_kernel_modules(self) -> None:
+    def _enable_kernel_modules(self) -> None:
         """Enable kernel modules needed by the charm."""
         execute_command(["/usr/sbin/modprobe", "br_netfilter"])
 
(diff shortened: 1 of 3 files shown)


src/charm.py
```diff
@@ -679,6 +679,13 @@ def _install_repo_policy_compliance(self) -> bool:
         )
         return old_version != new_version
 
+    def enable_kernel_modules(self) -> None:
+        """Enable kernel modules needed by the charm."""
+        execute_command(["/usr/sbin/modprobe", "br_netfilter"])
+
+        with open("/etc/modules", "a", encoding="utf-8") as modules_file:
+            modules_file.write("br_netfilter\n")
```
jdkandersson, line 687: what about using: https://docs.python.org/3/library/pathlib.html#pathlib.Path.write_text?

yhaliaw: That function overwrite the file, and there does not seems to be a pathlib function for appending.
This current approach is trying to append to the end of the file rather than overwriting the file.

src/charm.py
```diff
@@ -679,6 +679,13 @@ def _install_repo_policy_compliance(self) -> bool:
         )
         return old_version != new_version
 
+    def enable_kernel_modules(self) -> None:
```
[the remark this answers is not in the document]
yhaliaw: Good idea. I have made the change.

jdkandersson (dismissed):

jdkandersson (approved):

cbartz (approved):

PR #123, opened 2023-09-14
Bug/missing repo check error message

yhaliaw:
Applicable spec: NA

### Overview

Only half the error message from repo check is printed out. Fix this issue.
It was found that stopping the runner application prevent cURL from printing out the full body of response. Anything after a newline is cut off.

The fix is using the `logger` command to print out the content as `logger` functions normally under the scenario.

The integration test is modified to check for the reason of failure, e.g., `commit the job is running on is not signed`, to catch this issue if it pops up again.

### Rationale

The user are unable to see the reason for the repo check failure without this change.

### Checklist

### DIFF
--- a/tests/integration/test_charm_fork_repo.py
+++ b/tests/integration/test_charm_fork_repo.py
@@ -70,7 +70,20 @@ def forked_github_branch(forked_github_repository: Repository) -> Iterator[Branc
     branch_ref = forked_github_repository.create_git_ref(
         ref=f"refs/heads/{branch_name}", sha=main_branch.commit.sha
     )
-    branch = forked_github_repository.get_branch(branch_name)
+
+    for _ in range(10):
+        try:
+            branch = forked_github_repository.get_branch(branch_name)
+            break
+        except GithubException as err:
+            if err.status == 404:
+                sleep(5)
+                continue
+            raise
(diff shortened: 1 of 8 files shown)


tests/integration/test_charm_fork_repo.py
```diff
@@ -188,10 +188,7 @@ async def test_dispatch_workflow_failure(
             in logs
         ):
             assert run.jobs()[0].conclusion == "failure"
-            assert (
-                "Stopping execution of jobs due to repository setup is not compliant with policies"
-                in logs
-            )
```
cbartz, line 194: I think this message could be valuable and we could still check it.

To make the test goals more explicit, we could also change the docstring (L159) from
`assert: The workflow that was dispatched failed.` to `assert: The workflow that was dispatched failed and the reason is logged.`

jdkandersson (dismissed):
Looks good with the change that Chritopher proposed

tests/integration/test_charm_fork_repo.py
yhaliaw: Updated the code

jdkandersson (approved):

cbartz (approved):
LGTM

PR #127, opened 2023-09-19
Adapt charm to support COS integration

cbartz:
Applicable spec: [ISD-075](https://discourse.charmhub.io/t/specification-isd075-github-runner-cos-integration/12084)

### Overview

The PR adapts the charm to send metric logs to Loki by using  the [cos-agent](https://charmhub.io/grafana-agent/integrations) integration. 

- The metrics are logged into a file and then picked up by the [Grafana agent](https://charmhub.io/grafana-agent) and transferred to Loki.
- A simple metric event (`RunnerInstalled`) is fired to test the integration.
- A Grafana dashboard has also been added, which can be transferred to Grafana using the [cos-integration-k8s](https://charmhub.io/cos-configuration-k8s) charm.

image

### Rationale

To implement the COS integration as specified in the spec, the GitHub Runner charm needs to be adapted to send the metrics to Loki.

### Library Changes

New libraries fetched:

 * `charms/grafana_agent/v0/cos_agent.py`

### Checklist


cbartz:
Marked the PR as complex due to significant new feature added. @canonical/is-devops-leadership please provide feedback.

src/promtail.py
```diff
@@ -0,0 +1,24 @@
...
+def install() -> None:
+    """Install Promtail."""
+
+
+def config(loki_endpoint: str) -> None:
+    """Configure Promtail.
+
+    Args:
+        loki_endpoint: The Loki endpoint to send logs to.
+    """
```
jdkandersson, line 16: Do you think `install` and `config` could be combined, event, for example, with `start` as well? Or do you think we would potentially want to `stop` and `start` promtail multiple times?

jdkandersson (commented):
This looks good, thanks for putting this together!

jdkandersson:
Do you think any other modifications to public interfaces are required? I'm thinking about potentially:
* creating the logs temporary file system/ directory
* Reading the logs after the runner finished
* Submitting the logs to promtail

src/promtail.py
cbartz: Updated accordingly.

src/metrics.py
```diff
@@ -0,0 +1,32 @@
...
+
+class RunnerInstalled(Event):
+    """Metric event for when a runner is installed."""
+
+    name: str = "runner_installed"
+    """The name of the event."""
+    flavor: str
+    """The flavor of the runner."""
+    duration: conint(ge=0)
+    """The duration of the installation in seconds."""
```
jdkandersson, line 24: What are the docstrings below the attributes? Are the intended to be attribute descriptions? Usually those are done like this in the class docstring:

```python
class RunnerInstalled(Event):
    """
    Metric event for when a runner is installed.

    Attrs:
        name: The name of the event.
        ....
```

src/metrics.py
```diff
@@ -0,0 +1,32 @@
...
+
+    timestamp: conint(ge=0)
+    """The UNIX timestamp of the event."""
+
+
+class RunnerInstalled(Event):
+    """Metric event for when a runner is installed."""
+
+    name: str = "runner_installed"
+    """The name of the event."""
```
jdkandersson, line 20: It would be good to describe attributes using something other than the name of it, e.g., Short description of the event, what a flavour is meant to be, possible example values or something like that. This means that the description adds extra information

jdkandersson (commented):
I think the interfaces look good now, minor comments

src/metrics.py
cbartz: Updated. Removed the name attribute, as the name can also be retrieved by the class name. This would be used before sending the final log to Prometheus.

src/cos.py
```diff
@@ -22,21 +87,154 @@ def __init__(self, charm: ops.CharmBase, state: State):
...
+                "No Promtail binary information for amd64 architecture found."
+            )
+
+    def metrics_logging_available(self) -> bool:
+        """Check that the metrics logging integration is set up correctly.
+
+        Returns:
+            True if the integration is established, False otherwise.
+        """
+        return bool(self._retrieve_loki_integration_data().endpoints)
```
jdkandersson, line 159: Should this be included in `CharmState` rather than as a public function here? One of the ways that can be done is to pass the `requires` instances to `CharmState` and to the observer. The reason that this is relevant information about the charm state that needs to be used for the workload

src/metrics.py
```diff
@@ -2,26 +2,84 @@
...
 
-from pydantic import BaseModel, conint
+import requests
+from pydantic import BaseModel, NonNegativeInt
+
+import promtail
+
+PROMTAIL_PUSH_API_URL = "http://localhost:3100/loki/api/v1/push"
+
+requests_session = None  # pylint: disable=invalid-name
```
jdkandersson, line 16: What is the reason for needing this please? It looks a bit like a global variable? Also, for any disabled linting messages, please include a reason why it is reasonable to disable the rule in this case (see standard here: https://github.com/canonical/is-charms-contributing-guide/blob/main/CONTRIBUTING.md#static-code-analysis)

src/metrics.py
```diff
@@ -2,26 +2,84 @@
...
+
+def _get_session() -> requests.Session:
+    """Get a requests session.
+
+    Returns:
+        A requests session.
+    """
+    # We try to avoid capsuling the session in an object to be passed to issue_event,
+    # therefore we use the singleton pattern with a global variable.
+    global requests_session  # pylint: disable=global-statement
```
jdkandersson, line 51: Are there any alternatives here? Needing to maintain a session can be a reason to move to a class rather than a collection of functions. Global variables are usually not recommended

jdkandersson (commented):
Didn't have time to finish before my break, up to `promtail.py`

src/metrics.py
cbartz: I have removed the use of session here. Its not required, it could give some performance improvement but most likely not really felt, and I believe its not worth it if it increases implementation complexity (e.g. moving to a class or using a singleton/factory pattern to retrieve the session).

src/metrics.py
cbartz: I have removed it.

requirements.txt
```diff
@@ -6,3 +6,4 @@ requests
 typing-extensions
 # Newer version does not work with default OpenSSL version on jammy.
 cryptography <= 38.0.4
+pydantic>=1,<2
```
arturo-seijas, line 9: Would you mind pinning this?

src/cos.py
arturo-seijas: What I'm wondering if we should contribute this to the charm library

src/cos.py
```diff
@@ -0,0 +1,240 @@
...
+from event_timer import EventTimer
+
+METRICS_LOGGING_INTEGRATION_NAME = "metrics-logging"
+
+PROMTAIL_HEALTH_CHECK_INTERVAL_MINUTES = 5
+
+logger = logging.getLogger(__name__)
+
+
+class PromtailBinary(BaseModel):
```
arturo-seijas, line 28: IMO, these whould be in the loki lib

cbartz: @merkata (cc @arturo-seijas) Maybe you can address this in today's COS meeting: Would it be possible for the lib to provide dataclasses/pydantic models with appropriate validation? So in principle this would mean that the 3 dataclasses here and the method `_retrieve_loki_integration_data` would be in the `loki_push_api` lib.

requirements.txt
cbartz: I have pinned it more specifically, but not to an exact version. I am a bit hesitant, as the other packages in this file are also not specifically pinned, and it seems that we also need to agree on pinning as a team, taking into account @jdkandersson 's blog post https://jdkandersson.com/2023/09/18/to-pin-or-not-to-pin-the-version/ . cc @yhaliaw

requirements.txt
```diff
@@ -6,4 +6,4 @@ requests
 typing-extensions
 # Newer version does not work with default OpenSSL version on jammy.
 cryptography <= 38.0.4
-pydantic>=1,<2
\ No newline at end of file
+pydantic >= 1.10.13,<2
```
arturo-seijas, line 9: Sorry, I meant to the exact version, so that we know what's being pulled and let renovate handle the updates

cbartz: Yes, I assumed that, I added a comment above https://github.com/canonical/github-runner-operator/pull/127#discussion_r1339971396. As it would be the first package to be pinned in this charm, I asked @yhaliaw for input as it may be that the packages are not pinned for some reason/design choice.

src/cos.py
merkata: > @merkata (cc @arturo-seijas) Maybe you can address this in today's COS meeting: Would it be possible for the lib to provide dataclasses/pydantic models with appropriate validation? So in principle this would mean that the 3 dataclasses here and the method `_retrieve_loki_integration_data` would be in the `loki_push_api` lib.

defo, thanks for raising @cbartz  (and @arturo-seijas )!

merkata: > > @merkata (cc @arturo-seijas) Maybe you can address this in today's COS meeting: Would it be possible for the lib to provide dataclasses/pydantic models with appropriate validation? So in principle this would mean that the 3 dataclasses here and the method `_retrieve_loki_integration_data` would be in the `loki_push_api` lib.
> 
> defo, thanks for raising @cbartz (and @arturo-seijas )!

@sed-i  @0x12b fyi

arturo-seijas (dismissed):
LGTM

gregory-schiano:
Can you add a screenshot of the dashboard for reference ?

requirements.txt
gregory-schiano: I'd say consistant in the pinning approach, but we should definitely have a global discussion within the team on pinning

src/runner_manager.py
```diff
@@ -290,6 +295,40 @@ def _get_runner_health_states(self) -> RunnerByHealth:
...
+                metrics.issue_event(
+                    metrics.RunnerInstalled(
+                        timestamp=ts_after,
+                        flavor=self.app_name,
+                        duration=ts_after - ts_now,
+                    )
+                )
+            # We explicitly catch all exceptions here as we do not want to fail the
+            # runner creation process due to metrics issuing failure.
+            except Exception:  # pylint: disable=broad-except
```
gregory-schiano, line 329: I'm not 100% convinced, this could be the sign of really big problems on the charm, in such cases that's better to let the exception go through and lead to error status

But if such critical issues occurs on the charm, it's likely that some other parts will also raise exception

src/promtail.py
```diff
@@ -0,0 +1,208 @@
...
+import requests
+from charms.loki_k8s.v0.loki_push_api import PROMTAIL_BASE_URL
+
+from charm_state import ProxyConfig
+from utilities import execute_command
+
+SYSTEMCTL_PATH_STR = "/usr/bin/systemctl"
+
+PROMTAIL_BINARY_PATH = Path("/usr/local/bin/promtail-linux-amd64")
+PROMTAIL_BINARY_SHA_PATH = Path("/usr/local/bin/promtail-linux-amd64.sha256")
```
gregory-schiano, line 22: Same comments for architecture as previously

src/promtail.py
```diff
@@ -0,0 +1,208 @@
...
+    if not promtail_download_info.url.startswith(PROMTAIL_BASE_URL):
+        raise PromtailInstallationError(
+            f"Unknown Promtail download URL {promtail_download_info.url}. "
+            f"Must start with {PROMTAIL_BASE_URL}"
+        )
+    response = requests.get(promtail_download_info.url, timeout=300)
+    response.raise_for_status()
+
+    with open(PROMTAIL_BINARY_PATH.with_suffix(".gz"), "wb") as file:
+        logger.info("Writing Promtail binary zip to %s", PROMTAIL_BINARY_PATH.with_suffix(".gz"))
```
gregory-schiano, line 123: Picky comment: s/binary zip/binary gzip/

src/promtail.py
```diff
@@ -0,0 +1,208 @@
...
+        file.write(response.content)
+    if not _sha256sums_matches(
+        PROMTAIL_BINARY_PATH.with_suffix(".gz"), promtail_download_info.zip_sha256
+    ):
+        raise PromtailInstallationError(
+            f"Promtail zip file sha256sum mismatch, expected: {promtail_download_info.zip_sha256}"
+        )
+    with gzip.open(PROMTAIL_BINARY_PATH.with_suffix(".gz"), "rb") as file:
```
gregory-schiano, line 131: `PROMTAIL_BINARY_PATH.with_suffix(".gz")` is repeated multiple times, worthwhile storing it in a variable

src/promtail.py
```diff
@@ -0,0 +1,208 @@
...
+        logger.info("Writing Promtail binary to %s", PROMTAIL_BINARY_PATH)
+        PROMTAIL_BINARY_PATH.write_bytes(file.read())
+        PROMTAIL_BINARY_PATH.chmod(0o551)
```
gregory-schiano, line 134: Shouldn't we store it as a constant ?

gregory-schiano (commented):
Have to make another pass on the tests, but here's my review so far

src/cos.py
```diff
@@ -0,0 +1,240 @@
...
+
+        Raises:
+            LokiIntegrationDataNotComplete: If the integration data is not complete.
+        """
+        if not loki_integration_data.endpoints:
+            raise LokiIntegrationDataNotCompleteError("No Loki endpoint found.")
+        if not loki_integration_data.promtail_binaries.get("amd64"):
+            raise LokiIntegrationDataNotCompleteError(
+                "No Promtail binary information for amd64 architecture found."
+            )
```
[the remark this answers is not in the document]
cbartz: The architecture is now determined dynamically, in the same way as the lib: https://github.com/canonical/github-runner-operator/pull/127/files#diff-ad7051b5ca68befcdb77d638fcfa686fa902ae8089c8e72cea3c222df01a4597R1797-R1798

src/cos.py
```diff
@@ -0,0 +1,240 @@
...
+            event: The event object containing the Loki endpoint.
+        """
+        loki_integration_data = self._retrieve_loki_integration_data()
+        try:
+            self._validate_for_start(loki_integration_data)
+        except LokiIntegrationDataNotCompleteError as exc:
+            logger.info("Loki integration data not complete: %s Will not start Promtail", exc.msg)
+            return
+
+        self._start_promtail(loki_integration_data)
```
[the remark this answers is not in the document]
cbartz: I renamed the method.

templates/promtail.service.j2
```diff
@@ -0,0 +1,15 @@
+[Unit]
+Description=Promtail Service collecting metric logs
+After=network.target
+
+[Service]
+User=ubuntu
+Group=www-data
```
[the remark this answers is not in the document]
cbartz: Good catch, I just copied the group from `templates/repo-policy-compliance.service.j2`.

src/runner_manager.py
```diff
@@ -290,6 +295,40 @@ def _get_runner_health_states(self) -> RunnerByHealth:
...
+            runner: Runner to be created.
+        """
+        ts_now = time.time()
+        runner.create(
+            self.config.image,
+            resources,
+            RunnerManager.runner_bin_path,
+            registration_token,
+        )
+        ts_after = time.time()
```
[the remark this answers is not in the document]
cbartz: Have changed it, although my original thought was not to increase code complexity, I think the capture time is not relevant to performance.

src/runner_manager.py
cbartz: I narrowed the exception catch. My original thought was to obey the spec, which indicates that no error related to metrics should let the charm or runners stop working.

requirements.txt
cbartz: I have now pinned it.

src/cos.py
```diff
@@ -0,0 +1,240 @@
...
+        """Handle the PromtailHealthCheckEvent event.
+
+        Restarts Promtail if it is not running.
+
+        Args:
+            event: The event object
+        """
+        if not promtail.is_running():
+            logger.error("Promtail is not running, restarting")
+            promtail.restart()
```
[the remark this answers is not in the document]
cbartz: I have added a retry mechanism that leads to a blocked state if not resolvable.

cbartz:
I removed Promtail to send the logs directly to Loki. The PR description has also been updated.

src/cos.py
cbartz: The observer module is no longer required, and this method is now a property of the charm state.

src/cos.py
cbartz: I have removed the use of Promtail. Will resolve the conversation.

cbartz:
After feedback from the Observability team, I changed the implementation to use the `cos-agent` integration. I have updated the PR description.

metadata.yaml
gregory-schiano, line 1: EOL is missing

requirements.txt
```diff
@@ -6,4 +6,5 @@ requests
 typing-extensions
 # Newer version does not work with default OpenSSL version on jammy.
 cryptography <= 38.0.4
-pydantic == 1.10.3
\ No newline at end of file
+pydantic == 1.10.3
+cosl
```
gregory-schiano, line 10: Shouldn't we also pin cosl version ?

src/charm_state.py
```diff
@@ -49,16 +19,25 @@ class State:
 
     Attrs:
         proxy_config: Proxy configuration.
+        _charm: The charm instance.
     """
 
-    proxy_config: Optional[ProxyConfig]
+    _charm: CharmBase
```
gregory-schiano, line 25: I'm not a big fan of keeping the whole charm into the state, you should narrow down either to the relation or directly to the property (`is_metrics_logging_available`)

src/metrics.py
```diff
@@ -68,32 +73,61 @@ def _get_event_name(event: Event) -> str:
...
+
+
+def setup_logrotate():
+    """Configure logrotate for the metrics log.
+
+    Raises:
+        SubprocessError: If the logrotate.timer cannot be enabled.
+    """
+    _enable_logrotate()
+    _configure_logrotate()
```
gregory-schiano, line 133: Shouldn't handle to potential exception raised here ?

cbartz: An expected exception would be a `SubprocessError` which is passed through and handled by the caller in `charm.py`. 
I prefer not to catch it here in order to avoid that logrotate will never run if there is a systematic issue (e.g. `logrotate` is not installed). What do you think?

tests/integration/test_charm_metrics.py
```diff
@@ -0,0 +1,66 @@
...
+    Args:
+        app: The GitHub Runner Charm app to create the runner for.
+        model: The marchine charm model.
+    """
+    await app.set_config({"virtual-machines": "1"})
+    unit = app.units[0]
+    action = await unit.run_action("reconcile-runners")
+    await action.wait()
+    await model.wait_for_idle(apps=[app.name], status=ACTIVE_STATUS_NAME)
+    await wait_till_num_of_runners(unit, 1)
```
jdkandersson, line 29: Is there no fixture or helper function that does this already for another test? Just checking so we don't re-do something that is already done and I think I remember other tests that have runners in them

jdkandersson: I'm guessing that it is unique to this test since the relation is required before the runner is created. Perhaps see if this function could be used in the other cases when runners are being created?

tests/integration/test_charm_metrics.py
```diff
@@ -0,0 +1,66 @@
...
+    await model.wait_for_idle(apps=[grafana_agent.name])
+
+    await _create_runner(app=app, model=model)
+
+    metrics_log = await _get_metrics_log(app.units[0])
+    logging.info("Metric log: %s", metrics_log)
+    metric_log = json.loads(metrics_log)
+    assert metric_log.get("flavor") == app.name
+    assert metric_log.get("event") == "runner_installed"
+    assert metric_log.get("duration") >= 0
```
jdkandersson, line 66: Perhaps check that the logs are empty before the test just to double check that these entries were created during the test? That way any leakage between tests is detected

src-docs/runner_manager.py.md
```diff
@@ -32,8 +32,7 @@ __init__(
     app_name: str,
     unit: int,
     runner_manager_config: RunnerManagerConfig,
-    proxies: ProxySetting = {},
-    issue_metrics: bool = False
+    proxies: ProxySetting = {}
```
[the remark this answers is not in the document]
cbartz: Its origin is here https://github.com/canonical/github-runner-operator/blob/512796511a339f4de162eb6937c151fea32d1bce/src/runner_manager.py#L92 (ProxySetting is a TypedDict). This  has not been introduced in this PR but I have cleaned it up.

tests/integration/test_charm_metrics.py
cbartz: I moved it to `helpers.py` and reused it in a few places.

src/metrics.py
cbartz: Updated the code to throw a custom exception.

jdkandersson (approved):
Thank you!

gregory-schiano (approved):
Great, thx

PR #136, opened 2023-10-11
COS integration: Extract runner metrics

cbartz:
Applicable spec: [ISD-075](https://discourse.charmhub.io/t/specification-isd075-github-runner-cos-integration/12084)

### Overview

PR to adapt the charm to be able to extract the metrics from the runners.

Specifically

- Create a shared filesystem and place a timestamp file in it. We use lxd with loop device for this.
- Place the information for calculating the metrics in this shared file system.
- Clean up shared filesystems for all offline runners.
- To test the feature, a simple metrics event (RunnerStart) is issued (note that all events to be issued as per the spec will follow in a future pr) and the Grafana dashboard has been updated.

Screenshot from 2023-10-10 17-31-01

### Rationale

To implement the COS integration as specified in the spec, the GitHub Runner charm needs to be adapted to retrieve metrics from the runners.

### Module Changes
Two new modules:
- `shared_fs` : Handles creation and listing of shared filesystems
- `runner_metrics` : Extracts metrics from shared_fs

Notable module changes:
- `runner` : Attaches shared filesystem to runner lxd instance and renders pre-job.sh and start.sh to create metric files.

### Checklist


src/metrics.py
```diff
@@ -73,7 +73,26 @@ class RunnerInstalled(Event):
...
-    duration: NonNegativeInt
+    duration: NonNegativeFloat
+
+
+class RunnerStart(Event):
+    """Metric event for when a runner is started.
+
+    Attrs:
+        flavor: Describes the characteristics of the runner.
+          The flavour could be for example "small".
```
merkata, line 84: very nitpicky, no need to resolve - you have flavor and flavour, so mixing the two spellings ;)

src/runner.py
```diff
@@ -118,13 +129,24 @@ def create(
...
+                    logger.exception(
+                        "Unable to create shared filesystem for runner %s. "
+                        "Will not create metrics for this runner.",
+                        self.config.name,
+                    )
+
             self.instance = self._create_instance(image, resources)
             self._start_instance()
             # Wait some initial time for the instance to boot up
             time.sleep(60)
```
merkata, line 145: Do we need this sleep considering there's a _wait_boot_up function below it - I see in the definition of it a retry decorator, so I assume it's doing the same thing?

src/runner_metrics.py
```diff
@@ -0,0 +1,198 @@
...
+    The metrics are extracted from the shared filesystem of given runners
+    and respective metric events are issued.
+    Orphan shared filesystems are cleaned up.
+
+    If corrupt data is found, an error is raised immediately, as this may indicate that a malicious
+    runner is trying to manipulate the shared file system.
+    In order to avoid DoS attacks, the file size is also checked.
+
+    Args:
+        flavor: The flavour of the runners to extract metrics from.
```
merkata, line 176: flavor flavour

merkata (dismissed):
LGTM, very minor nitpicks that don't need resolving.

src/runner.py
cbartz: This was not introduced by this PR, but I believe the sleep should prevent the logging of an error message (https://github.com/canonical/github-runner-operator/blob/589e139e843cce84b778bf1b513ed9a16c7b5ee6/src/utilities.py#L80) , which happens on every retry.

merkata: Sounds good, I also noticed it is not part of the PR, wanted to ask. Thanks!

merkata (dismissed):

src/runner_metrics.py
```diff
@@ -0,0 +1,198 @@
...
+    files = [
+        fs.path.joinpath(PRE_JOB_METRICS_FILE_NAME),
+        fs.path.joinpath(POST_JOB_METRICS_FILE_NAME),
+        fs.path.joinpath(RUNNER_INSTALLED_TS_FILE_NAME),
+    ]
+    for file in files:
+        if file.exists() and file.stat().st_size > FILE_SIZE_BYTES_LIMIT:
+            raise MetricFileSizeTooLargeError(
+                f"File {file} is too large. The limit is {FILE_SIZE_BYTES_LIMIT} bytes."
+            )
```
jdkandersson, line 87: These kind of functions are interesting on whether they should raise exceptions or return data for the caller to raise exceptions with. An exception is conceptually that something went wrong during the execution. For the `_inspect_file_sizes`, it isn't an error for the file size to be too large since it is just inspecting it, so it should just return perhaps a boolean whether the inspection passed or failed and the caller can decide what to do with the information

src/runner_metrics.py
```diff
@@ -0,0 +1,198 @@
...
+        post_job_metrics = json.loads(fs.path.joinpath(POST_JOB_METRICS_FILE_NAME).read_text())
+        logger.debug("Post-job metrics for runner %s: %s", fs.runner_name, post_job_metrics)
+    except FileNotFoundError:
+        logger.warning("%s not found for runner %s", POST_JOB_METRICS_FILE_NAME, fs.runner_name)
+        post_job_metrics = None
+
+    return RunnerMetrics(
+        installed_timestamp=installed_timestamp,
+        pre_job=PreJobMetrics(**pre_job_metrics),
+        post_job=PostJobMetrics(**post_job_metrics) if post_job_metrics else None,
```
jdkandersson, line 126: What about if the `pre_job_metrics` or `post_job_metrics` are valid JSON but not a dictionary?

src/runner_metrics.py
```diff
@@ -0,0 +1,198 @@
...
+        ignore_runners: The set of runners to ignore.
+
+    Raises:
+        CorruptDataError: If one of the files inside the shared filesystem is not valid.
+    """
+    for fs in shared_fs.list_all():
+        if fs.runner_name not in ignore_runners:
+            try:
+                metrics_from_fs = _extract_metrics_from_fs(fs)
+            except (JSONDecodeError, ValidationError, MetricFileSizeTooLargeError) as exc:
```
jdkandersson, line 186: It would be good for `_extract_metrics_from_fs` not to bubble up errors like `JSONDecodeError` or any others not explicitly raised in `_extract_metrics_from_fs` just so that it is clear what exceptions can be raised in each function and to avoid having implementation details leak to the caller

src/shared_fs.py
```diff
@@ -0,0 +1,117 @@
...
+
+    Returns:
+        The shared filesystem object.
+
+    Raises:
+        CreateSharedFilesystemError: If the command fails.
+    """
+    FILESYSTEM_BASE_PATH.mkdir(exist_ok=True)
+    runner_fs_path = FILESYSTEM_BASE_PATH / runner_name
+    runner_fs_path.mkdir()
```
jdkandersson, line 46: What if the directory already exists for some reason?

src/shared_fs.py
```diff
@@ -0,0 +1,117 @@
...
+        )
+        execute_command(["sudo", "chown", "ubuntu:ubuntu", str(runner_fs_path)], check_exit=True)
+    except SubprocessError as exc:
+        raise CreateSharedFilesystemError(
+            f"Failed to create shared filesystem for runner {runner_name}"
+        ) from exc
+    return SharedFilesystem(runner_fs_path, runner_name)
+
+
+def list_all() -> Generator[SharedFilesystem, None, None]:
```
jdkandersson, line 67: I think `iterator[SharedFilesystem]` should be ok here as well

src/shared_fs.py
```diff
@@ -0,0 +1,117 @@
...
+    """Delete the shared filesystem for the runner.
+
+    Args:
+        runner_name: The name of the runner.
+
+    Raises:
+        NotFoundError: If the shared filesystem is not found.
+    """
+    runner_fs = get(runner_name)
+    runner_image_path = FILESYSTEM_BASE_PATH / f"{runner_name}.img"
```
jdkandersson, line 91: I think this is also done elsewhere, should there be a helper function just to ensure that the path is always generated the same way?

jdkandersson (commented):
Have not looked at the tests yet, thanks for putting this together!

.github/workflows/integration_tests.yaml
```diff
@@ -116,13 +116,18 @@ jobs:
           provider: lxd
       - name: Enable br_netfilter
         run: sudo modprobe br_netfilter
+      - name: Find available loop device
+        id: loop-device
+        run: |
+          echo "device=$(sudo losetup -f)" >> $GITHUB_OUTPUT
```
amandahla, line 122: Cool, didn't know this command :)

src/runner_metrics.py
```diff
@@ -0,0 +1,198 @@
...
+    Raises:
+        FileSizeTooLargeError: If one of the files is too large.
+    """
+    files = [
+        fs.path.joinpath(PRE_JOB_METRICS_FILE_NAME),
+        fs.path.joinpath(POST_JOB_METRICS_FILE_NAME),
+        fs.path.joinpath(RUNNER_INSTALLED_TS_FILE_NAME),
+    ]
+    for file in files:
+        if file.exists() and file.stat().st_size > FILE_SIZE_BYTES_LIMIT:
```
amandahla, line 84: I believe it was not mentioned in the spec but do you think if an invalid file is found, it should be deleted?

amandahla (dismissed):
Very interesting PR

src/runner_metrics.py
cbartz: Thanks for the suggestion. I think we should not remove the files that are too large, as we may want to inspect those files, as it could also be a programming error on our part (since we are writing to the file).

amandahla: I see. I was just concerned that they might be left behind and somehow create a problem. Maybe compress and move somewhere else.

src/runner_metrics.py
```diff
@@ -0,0 +1,215 @@
...
+    except FileNotFoundError:
+        logger.exception("installed_timestamp not found for runner %s", fs.runner_name)
+        return None
+
+    logger.debug("Runner %s installed at %s", fs.runner_name, installed_timestamp)
+    try:
+        pre_job_metrics = json.loads(fs.path.joinpath(PRE_JOB_METRICS_FILE_NAME).read_text())
+        logger.debug("Pre-job metrics for runner %s: %s", fs.runner_name, pre_job_metrics)
+    except FileNotFoundError:
+        logger.warning("%s not found for runner %s.", PRE_JOB_METRICS_FILE_NAME, fs.runner_name)
```
amandahla, line 119: Minor/can be addressed in a follow-up PR: logger.exception instead?

Out of curiosity: whats the expected action/monitoring in case of metrics not found?

src/runner_metrics.py
```diff
@@ -0,0 +1,215 @@
...
+    """
+    event = metrics.RunnerStart(
+        timestamp=runner_metrics.pre_job.timestamp,
+        flavor=flavor,
+        workflow=runner_metrics.pre_job.workflow,
+        repo=runner_metrics.pre_job.repository,
+        github_event=runner_metrics.pre_job.event,
+        idle=runner_metrics.pre_job.timestamp - runner_metrics.installed_timestamp,
+    )
+    metrics.issue_event(event)
```
amandahla, line 170: Should OSError be catched?

src/runner_metrics.py
```diff
@@ -0,0 +1,215 @@
...
+
+    Remove all metric files and afterwards the shared filesystem.
+    Args:
+        fs: The shared filesystem for a specific runner.
+    """
+    fs.path.joinpath(PRE_JOB_METRICS_FILE_NAME).unlink(missing_ok=True)
+    fs.path.joinpath(POST_JOB_METRICS_FILE_NAME).unlink(missing_ok=True)
+    fs.path.joinpath(RUNNER_INSTALLED_TS_FILE_NAME).unlink(missing_ok=True)
+
+    shared_fs.delete(fs.runner_name)
```
amandahla, line 184: Same question for NotFoundError

src/runner_metrics.py
amandahla: ps.: This can be addressed in a follow-up PR.

src/runner_metrics.py
cbartz: This is not actually an error. This can happen if a runner never picks up a job (e.g. because the number of runners has been reduced).I think it is only necessary to take action if there are a lot of warnings, as this may indicate a programming error.

src/runner_metrics.py
cbartz: Good catch. I do not think so, as it may indicate a more serious error. I have also removed it from the docstring as the fct does not explicitly raise it either.

src/runner_metrics.py
cbartz: Good catch. I have updated the code for handling an error on delete.

src/runner_metrics.py
cbartz: Ah, I see. I think this should almost never happen, only if we really come across a malicious runner. 
The filesystem size is currently 1M, so it is unlikely to fill up the disk. I assume the operator would take immediate action if this happened, as the charm would go into error state.
@jdkandersson @yhaliaw What do you think?

src/runner_metrics.py
cbartz: Yes, the name of the method does not necessarily indicate that an error would be thrown. I have updated the code.

src/runner_metrics.py
cbartz: I thought about it. I saw that we already catch it in RunnerManager, and to be consistent it makes sense to catch it here too. I changed the exc to a custom one and it is now caught.

src/runner_metrics.py
```diff
@@ -0,0 +1,222 @@
...
+            f"The limit is {FILE_SIZE_BYTES_LIMIT} bytes."
+        )
+
+    try:
+        installed_timestamp = fs.path.joinpath(RUNNER_INSTALLED_TS_FILE_NAME).read_text()
+    except FileNotFoundError:
+        logger.exception("installed_timestamp not found for runner %s", fs.runner_name)
+        return None
+
+    logger.debug("Runner %s installed at %s", fs.runner_name, installed_timestamp)
```
gregory-schiano, line 115: I would move the log inside the try block similarly to other metrics log afterwards

src/runner_metrics.py
```diff
@@ -0,0 +1,222 @@
...
+    if not isinstance(post_job_metrics, dict) and post_job_metrics is not None:
+        raise CorruptMetricDataError(
+            f"Post-job metrics for runner {fs.runner_name} is not a JSON object."
+        )
+
+    try:
+        return RunnerMetrics(
+            installed_timestamp=installed_timestamp,
+            pre_job=PreJobMetrics(**pre_job_metrics),
+            post_job=PostJobMetrics(**post_job_metrics) if post_job_metrics else None,
```
gregory-schiano, line 148: When are the PostJobMetrics used ? I can't find any reference to it

src/shared_fs.py
```diff
@@ -0,0 +1,154 @@
...
+def _get_runner_image_path(runner_name: str) -> Path:
+    """Get the path of the runner image.
+
+    Args:
+        runner_name: The name of the runner.
+
+    Returns:
+        The path of the runner image.
+    """
+    return FILESYSTEM_BASE_PATH / f"{runner_name}.img"
```
gregory-schiano, line 46: So the image mounted using loopback is in the same directory as the mount point right ?
Well have something like:
```
runner_1             dir
runner_1.img      file
runner_2            dir
runner_2img      file
....
```
Shouldn't we have a dedicated folder for the images ?

src/runner_metrics.py
```diff
@@ -0,0 +1,222 @@
...
+def _clean_up_shared_fs(fs: shared_fs.SharedFilesystem) -> None:
+    """Clean up the shared filesystem.
+
+    Remove all metric files and afterwards the shared filesystem.
+    Args:
+        fs: The shared filesystem for a specific runner.
+    """
+    fs.path.joinpath(PRE_JOB_METRICS_FILE_NAME).unlink(missing_ok=True)
+    fs.path.joinpath(POST_JOB_METRICS_FILE_NAME).unlink(missing_ok=True)
+    fs.path.joinpath(RUNNER_INSTALLED_TS_FILE_NAME).unlink(missing_ok=True)
```
gregory-schiano, line 183: Isn't it the responsability of `shared_fs` to empty the fs directory ? I think it's anyway doing `rmtree` on the folder

src/runner_metrics.py
```diff
@@ -0,0 +1,222 @@
...
+    Args:
+        flavor: The flavor of the runners to extract metrics from.
+        ignore_runners: The set of runners to ignore.
+
+    Raises:
+        CorruptMetricDataError: If one of the files inside the shared filesystem is not valid.
+    """
+    for fs in shared_fs.list_all():
+        if fs.runner_name not in ignore_runners:
+            metrics_from_fs = _extract_metrics_from_fs(fs)
```
gregory-schiano, line 211: Shouldn't we catch exceptions here, because if any runner fail, all the other won't be treated

src/runner_manager.py
```diff
@@ -365,6 +379,9 @@ def reconcile(self, quantity: int, resources: VirtualMachineResources) -> int:
             len(runner_states.unhealthy),
         )
 
+        if self.config.charm_state.is_metrics_logging_available:
+            runner_metrics.extract(flavor=self.app_name, ignore_runners=set(runner_states.healthy))
```
gregory-schiano, line 383: Extract can raise many types of exception, shouldn't we manage them (some) here to avoid breaking the reconcile loop ?

gregory-schiano (commented):
First pass, still need to look at tests in details

src/runner_manager.py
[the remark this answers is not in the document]
cbartz: I think we should also extract metrics for the unhealthy runners before cleaning them up, as these may be crashed runners that already have a `pre-job` metrics file in the shared fs. So I would call extract before the `if` block.

src/runner_metrics.py
cbartz: Not used yet. This will be used in a follow-up PR where all metrics (including `RunnerStop`) will be logged.

src/runner_metrics.py
cbartz: The idea was to remove the metric files explicitly, and let an error pass through in order to avoid keeping the files and logging the same metric multiple times in the next reconciliation pass.

src/runner_metrics.py
cbartz: Hm, the only "expected" exception would be `CorruptMetricDataError`. I am not sure if a catch would be too quiet as this could be a security issue (malicious runner) or a programming error.

src/runner_manager.py
cbartz: Same as above, the only expected (not explicitly caught) exceptiond would be `CorruptMetricDataError` and an `OSError` if any of the metric files could not be removed and I hesitate to catch them for the reasons mentioned above.

src/runner_metrics.py
cbartz: Code has been updated, files are still explicitly unlinked but errors are caught.

src/runner_metrics.py
cbartz: @amandahla So, after talking to Greg, we decided to go your way (move to quarantine). ;-)

gregory-schiano (approved):
Great PR 👍  LGTM

amandahla (approved):
LGTM nice!

PR #147, opened 2023-11-14
Use aproxy and install yarn

cbartz:
Applicable spec: n/a

### Overview

- Install [aproxy](https://github.com/canonical/aproxy) by default and use it if specified by config option. This option should be considered experimental and may be removed, as aproxy has not been reviewed. Additional cleanup may also be considered in a follow-on PR if aproxy is to be fully used by default (e.g. remove explicit Docker proxy setup).
- Install yarn.

### Rationale

- To fix proxy-related problems with Github actions (when applications do not respect the http_proxy,... env variables), we use aproxy.
- yarn is missing in the Github self-hosted runners (but is installed by default in the Github hosted runners), causing workflows to fail.

### Juju Events Changes

n/a

### Module Changes

- `charm_state`: Validate aproxy config
- `runner`: Install aproxy and yarn

### Checklist

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -70,3 +70,13 @@ options:
       The private docker registry configured as dockerhub mirror to be used by the runners. If set 
       a message will be displayed prior to job execution on self-hosted runner informing users to 
       use the provided registry.
+  aproxy-proxy:
+    type: string
+    default: ""
+    description: |
+      (Experimental, may be removed) When set to a http proxy server socket address 
+      (format hostname:port, e.g. squid.internal:3128),
+      the charm (and the runners) will use aproxy (https://github.com/canonical/aproxy) 
+      to forward all traffic to the proxy server.
+      This is useful when the charm is deployed in a network that requires a proxy to access the 
+      internet.
--- a/config.yaml
+++ b/config.yaml
@@ -76,7 +76,7 @@ options:
     description: |
       (Experimental, may be removed) When set to a http proxy server socket address 
       (format hostname:port, e.g. squid.internal:3128),
-      the charm (and the runners) will use aproxy (https://github.com/canonical/aproxy) 
+      the charm (and the runners) will use [aproxy](https://github.com/canonical/aproxy) 
       to forward all traffic to the proxy server.
       This is useful when the charm is deployed in a network that requires a proxy to access the 
       internet.
--- a/config.yaml
+++ b/config.yaml
@@ -80,3 +80,5 @@ options:
       to forward all traffic to the proxy server.
       This is useful when the charm is deployed in a network that requires a proxy to access the 
       internet.
+      Note that you cannot use a proxy server listening on port 80 or 443, as all traffic to these
+      ports is relayed to aproxy, which would cause an infinite loop.
--- a/config.yaml
+++ b/config.yaml
@@ -70,7 +70,7 @@ options:
       The private docker registry configured as dockerhub mirror to be used by the runners. If set 
       a message will be displayed prior to job execution on self-hosted runner informing users to 
       use the provided registry.
-  aproxy-proxy:
+  aproxy-address:
     type: string
     default: ""
     description: |
--- a/config.yaml
+++ b/config.yaml
@@ -70,7 +70,7 @@ options:
       The private docker registry configured as dockerhub mirror to be used by the runners. If set 
       a message will be displayed prior to job execution on self-hosted runner informing users to 
       use the provided registry.
-  use-aproxy:
+  experimental-use-aproxy:
     type: boolean
     default: false
     description: |
--- a/config.yaml
+++ b/config.yaml
@@ -70,15 +70,15 @@ options:
       The private docker registry configured as dockerhub mirror to be used by the runners. If set 
       a message will be displayed prior to job execution on self-hosted runner informing users to 
       use the provided registry.
-  aproxy-address:
-    type: string
-    default: ""
+  use-aproxy:
+    type: boolean
+    default: false
     description: |
-      (Experimental, may be removed) When set to a http proxy server socket address 
-      (format hostname:port, e.g. squid.internal:3128),
-      the charm (and the runners) will use [aproxy](https://github.com/canonical/aproxy) 
-      to forward all traffic to the proxy server.
+      (Experimental, may be removed) When set to true, [aproxy](https://github.com/canonical/aproxy) 
+      will  be installed inside the runners and will forward all http/s traffic to a proxy server
+      configured by the juju model config `juju-http-proxy` 
+      (or if this is not set, `juju-https-proxy` will be used).
       This is useful when the charm is deployed in a network that requires a proxy to access the 
       internet.
-      Note that you cannot use a proxy server listening on port 80 or 443, as all traffic to these
-      ports is relayed to aproxy, which would cause an infinite loop.
+      Note that you should not specify a proxy server listening on port 80 or 443, as all traffic
+      to these ports is relayed to aproxy, which would cause an infinite loop.
--- a/scripts/pre-integration-test.sh
+++ b/scripts/pre-integration-test.sh
@@ -8,4 +8,13 @@ sudo modprobe br_netfilter
 
 # Find a loop-device
 loop_device=$(sudo losetup -f)
-echo "PYTEST_ADDOPTS=--loop-device=$loop_device" >> "$GITHUB_ENV"
+
+# Install squid proxy for proxy test
+default_ip=$(ip route get $(ip route show 0.0.0.0/0 | grep -oP 'via \K\S+') | grep -oP 'src \K\S+')
+sudo apt install squid
+sudo sed -i 's/http_access deny/# http_access deny/g' /etc/squid/squid.conf
+echo "http_access allow all" | sudo tee -a /etc/squid/squid.conf
+sudo systemctl restart squid
+
+# Output PYTEST_ADDOPTS to GitHub Actions environment
+echo "PYTEST_ADDOPTS=--loop-device=$loop_device --squid-proxy=http://$default_ip:3128" >> "$GITHUB_ENV"
--- a/scripts/pre-integration-test.sh
+++ b/scripts/pre-integration-test.sh
@@ -11,7 +11,7 @@ loop_device=$(sudo losetup -f)
 
 # Install squid proxy for proxy test
 gateway_ip="$(ip route show 0.0.0.0/0 | grep -oP 'via \K\S+')"
-default_ip="$(ip route get $gateway_ip | grep -oP 'src \K\S+')"
+default_ip="$(ip route get "$gateway_ip" | grep -oP 'src \K\S+')"
 sudo apt install squid
 sudo sed -i 's/http_access deny/# http_access deny/g' /etc/squid/squid.conf
 echo "http_access allow all" | sudo tee -a /etc/squid/squid.conf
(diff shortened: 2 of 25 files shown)


cbartz:
@weiiwang01 As you are the author of aproxy, could you please take a look?

config.yaml
```diff
@@ -70,3 +70,13 @@ options:
       The private docker registry configured as dockerhub mirror to be used by the runners. If set 
       a message will be displayed prior to job execution on self-hosted runner informing users to 
       use the provided registry.
+  aproxy-proxy:
+    type: string
+    default: ""
+    description: |
+      (Experimental, may be removed) When set to a http proxy server socket address 
+      (format hostname:port, e.g. squid.internal:3128),
+      the charm (and the runners) will use aproxy (https://github.com/canonical/aproxy)
```
arturo-seijas, line 79: Will this be formated when published to charmhub? If so, we can use a proper markdown link here

src/charm_state.py
```diff
@@ -5,32 +5,63 @@
...
 
 @dataclasses.dataclass(frozen=True)
 class State:
     """The charm state.
 
     Attributes:
-        proxy_config: Proxy configuration.
-        _charm: The charm instance.
+        is_metrics_logging_available: Whether the charm is able to issue metrics.
+        aproxy_proxy: The socket address of the proxy to configure aproxy with.
```
arturo-seijas, line 41: This name sounds a bit odd to me. How about `proxy_address`?

src/charm_state.py
```diff
@@ -5,32 +5,63 @@
...
         """Initialize the state from charm.
 
+        Args:
+            charm: The charm instance.
+
         Returns:
             Current state of the charm.
         """
+        if aproxy_proxy := charm.config.get("aproxy-proxy"):
+            if not HOSTNAME_PORT_PATTERN.match(aproxy_proxy):
```
arturo-seijas, line 58: Why not using pydantic to validate with a regex?

src/charm_state.py
cbartz: I want to include "aproxy" to avoid confusing it with the other proxy settings from juju (e.g. from `JUJU_CHARM_HTTP_PROXY`). Maybe `aproxy_address`?

src/charm_state.py
cbartz: I thought it was a bit simpler than introducing a new class inheriting from `BaseModel` or a pydantic field validator. But we'll probably introduce a ProxySetting class in `charm_state` to when we move the juju proxy setting to that place.

arturo-seijas: Up to you, then

arturo-seijas (dismissed):
LGTM

src/charm.py
```diff
@@ -419,6 +424,46 @@ def _on_upgrade_charm(self, _event: UpgradeCharmEvent) -> None:
...
+
+        execute_command(["snap", "set", "aproxy", f"proxy={aproxy_proxy}"])
+
+        stdout, _ = execute_command(
+            [
+                "/bin/bash",
+                "-c",
+                r"ip route get $(ip route show 0.0.0.0/0 | grep -oP 'via \K\S+') |"
+                r" grep -oP 'src \K\S+'",
+            ],
```
weiiwang01, line 443: This script is to get the ip address of the default nic. I think this is obtainable from the dnsmasq DHCP lease record on host `/var/snap/lxd/common/lxd/networks/lxdbr0/dnsmasq.leaseses`

src/charm.py
```diff
@@ -419,6 +424,46 @@ def _on_upgrade_charm(self, _event: UpgradeCharmEvent) -> None:
...
+            check_exit=True,
+        )
+        default_ip = stdout.strip()
+
+        nft_input = f"define default-ip = {default_ip}"
+        nft_input += r"""
```
weiiwang01, line 449:
```suggestion
        nft_input = f"""define default-ip = {default_ip}"
```

f-string and triple quotes can be used together.

cbartz: Yes, I know, but my intention  was to avoid f raw strings. But I have updated the code now.

src/charm.py
cbartz: Using your method, I get the IP of the runner vm, not the charm unit. But aproxy is installed in the charm unit.

src/charm.py
weiiwang01: Yes, indeed, if we have to use the f string, we need to escape all the curly braces. Sorry I missed that, up to you to use your previous approach or use the f string.

src/charm.py
```diff
@@ -768,6 +815,9 @@ def _install_deps(self) -> None:
                 "security.port_isolation=true",
             ]
         )
+        # install aproxy, currently only edge channel available
+        execute_command(["snap", "install", "aproxy", "--edge"])
```
weiiwang01, line 819: Wouldn't it be better to set up the aproxy within the runner machine, rather than within the charm unit? That would allow users to easily access and review the aproxy logs inside the GitHub workflows to identify any proxy issues, such as rejections caused by upstream proxy server ACL rules.

cbartz: Have updated the code accordingly

src/charm.py
cbartz: Have updated the code

yanksyoon: Sorry for my lack of understanding, would this and the following code be trying to:
route all non-local http/s traffic from/towards the charm unit to each of the spawned runner VMs to the dnat and hence we need the charm unit IP?

cbartz: So, I moved aproxy inside the runner vm with the latest commits as requested by weii through latest commits https://github.com/canonical/github-runner-operator/pull/147#discussion_r1395426026 .
The goal is to route all non-local http/s traffic from inside the runner to aproxy. We now need the runner IP, not the charm unit IP.

config.yaml
```diff
@@ -70,3 +70,13 @@ options:
       The private docker registry configured as dockerhub mirror to be used by the runners. If set 
       a message will be displayed prior to job execution on self-hosted runner informing users to 
       use the provided registry.
+  aproxy-address:
+    type: string
+    default: ""
+    description: |
+      (Experimental, may be removed) When set to a http proxy server socket address 
+      (format hostname:port, e.g. squid.internal:3128),
```
weiiwang01, line 78: The nftables rules we used is a generic one, which means that the proxy ip address can't not be a public ip with proxy port is 80 or 443 at the same time. Or else, the nftables rule will cause an infinite loop. Maybe we should include that in the document?

weiiwang01 (dismissed):

src/runner.py
```diff
@@ -548,6 +553,83 @@ def _should_render_templates_with_metrics(self) -> bool:
...
+        self.instance.execute(["snap", "set", "aproxy", f"proxy={aproxy_address}"])
+
+        stdout, _ = execute_command(
+            [
+                "/usr/bin/awk",
+                f'$4 == "{self.instance.name}" {{ print $3 }}',
+                LXDBR_DNSMASQ_LEASES_FILE,
+            ],
+            check_exit=True,
+        )
```
weiiwang01, line 579: When using `awk`, if the name doesn't exist in the file, `awk` will print an empty line and `exit 0`. Maybe we should check the return value here, or using Python to parse the file and raise an exception.

config.yaml
jdkandersson: Shouldn't we just use the model proxy configuration for this?

tests/integration/test_charm_one_runner.py
```diff
@@ -49,6 +50,35 @@ async def test_network_access(app: Application) -> None:
...
 
+@pytest.mark.asyncio
+@pytest.mark.abort_on_fail
+async def test_usage_of_aproxy(model: Model, app_no_runner: Application, squid_proxy: str) -> None:
+    """
+    arrange: A working application with one runner using aproxy configured for a proxy server.
+    act: Run curl in the runner.
+    assert: The aproxy log contains the request.
+    """
+    app = app_no_runner  # Rename to make it clear that the app will contain a runner.
```
jdkandersson: Perhaps create a different fixture? How would this runner be added in this case?

config.yaml
cbartz: Originally I thought it would be better to be independent of the juju proxy configuration, since aproxy was installed in the charm and could be used independently. With the current approach, we still need the model proxy configuration for the charm, so it makes sense to reuse that setting for aproxy. I have updated the code.

scripts/pre-integration-test.sh
```diff
@@ -8,4 +8,14 @@ sudo modprobe br_netfilter
...
+# Install squid proxy for proxy test
+gateway_ip="$(ip route show 0.0.0.0/0 | grep -oP 'via \K\S+')"
+default_ip="$(ip route get "$gateway_ip" | grep -oP 'src \K\S+')"
+sudo apt install squid
+sudo sed -i 's/http_access deny/# http_access deny/g' /etc/squid/squid.conf
+echo "http_access allow all" | sudo tee -a /etc/squid/squid.conf
+sudo systemctl restart squid
+
+# Output PYTEST_ADDOPTS to GitHub Actions environment
+echo "PYTEST_ADDOPTS=--loop-device=$loop_device --squid-proxy=$default_ip:3128" >> "$GITHUB_ENV"
```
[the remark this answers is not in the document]
cbartz: Have updated the code

weiiwang01 (dismissed):

src/runner.py
```diff
@@ -759,6 +828,22 @@ def _apt_install(self, packages: Iterable[str]) -> None:
...
+    def _snap_install(self, snaps: Iterable[Snap]) -> None:
+        """Installs the given snap packages.
+
+        This is a temporary solution to provide tools not offered by the base ubuntu image. Custom
+        images based on the GitHub action runner image will be used in the future.
+
+        Args:
+            snaps: snaps to be installed.
+        """
+        if self.instance is None:
```
arturo-seijas, line 840: Is this actually possible?

src/runner.py
```diff
@@ -549,6 +556,98 @@ def _should_render_templates_with_metrics(self) -> bool:
         """
         return self._shared_fs is not None
 
+    def _get_default_ip(self) -> Optional[str]:
+        """Get the default IP of the runner.
+
+        Returns:
+            The default IP of the runner or None if not found.
+        """
+        if self.instance is None:
```
arturo-seijas, line 565: This is certainly not reachable, as far as I can tell

arturo-seijas (dismissed):
LGTM. Some minor inline comments

src/runner.py
cbartz: This is mainly to make mypy happy. I followed the way it was done in the other methods.

src/runner.py
cbartz: This is mainly to make mypy happy. I followed the way it was done in the other methods.

src/charm.py
```diff
@@ -167,6 +172,8 @@ def __init__(self, *args, **kargs) -> None:
         no_proxy = get_env_var("JUJU_CHARM_NO_PROXY")
         if (https_proxy or http_proxy) and no_proxy:
             self.proxies["no_proxy"] = no_proxy
+        if self._state.aproxy_address:
+            self.proxies["aproxy_address"] = self._state.aproxy_address
```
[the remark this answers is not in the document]
benhoyt: No, not really. I suppose you'd just do them as charm config, perhaps with a special name/prefix to indicate it's a (short-lived) feature flag rather than stable config?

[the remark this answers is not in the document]
cbartz: Thank you, Ben, for suggesting this.

arturo-seijas (approved):

weiiwang01 (approved):

PR #156, opened 2023-11-29
Fix repo-policy-compliance in integration test

yhaliaw:
### Overview

Add test workflow that always failure repo-policy-compliance check. 

### Rationale

Changes to repo-policy-compliance causes some test to unable to fail the repo check.

### Checklist


.github/workflows/workflow_dispatch_failure_test.yaml
```diff
@@ -0,0 +1,23 @@
...
+jobs:
+  workflow-dispatch-tests:
+    runs-on: [self-hosted, linux, x64, "${{ inputs.runner }}"]
+    steps:
+      - name: Echo input variable and message
+        run: |
+          echo "Hello, runner: ${{ inputs.runner }}"
+      - name: Always echo a message
+        if: always()
+        run: echo "Should not echo if pre-job script failed"
```
jdkandersson, line 23: Missing new line please

jdkandersson (commented):
Looks good

.github/workflows/workflow_dispatch_failure_test.yaml
```diff
@@ -0,0 +1,23 @@
+name: Workflow Dispatch Failure Tests 2a34f8b1-41e4-4bcb-9bbf-7a74e6c482f7
```
yanksyoon, line 1: Would it be better to use a different name than UUID here?

tests/integration/conftest.py
```diff
@@ -311,3 +309,20 @@ def forked_github_branch(
...
+
+    Test should ensure it returns with the application in a good state and has
+    one runner.
+    """
+    app = app_no_runner  # alias for readability as the app will have a runner during the test
+
+    await app.set_config({"path": forked_github_repository.full_name})
+    await ensure_charm_has_runner(app=app, model=model)
+
+    yield app
```
yanksyoon, line 328: Since there's no cleanup code, i think it can return rather than yield :)

yanksyoon (dismissed):
LGTM, minor comments!

jdkandersson (dismissed):

.github/workflows/workflow_dispatch_failure_test.yaml
jdkandersson: The theory is that, whatever name we pick, may clash with a workflow name of a user and then that workflow always gets declined. Including a UUID reduces the chance of a clash essentially to zero which is why it is useful in this case

tests/integration/test_charm_fork_repo.py
```diff
@@ -75,7 +75,7 @@ async def test_dispatch_workflow_failure(
                 "Stopping execution of jobs due to repository setup is not compliant with policies"
                 in logs
             )
-            assert "Endpoint designed for testing that always fails" in logs
+            assert "The method is not allowed for the requested URL" in logs
```
jdkandersson, line 78: That doesn't seem right?

yhaliaw: The message from repo-policy-compliance is returning that. See:
https://github.com/canonical/github-runner-operator/actions/runs/7032231846/job/19139302478#step:14:167
After this change the integrations test are passing.

Since this is blocking other PR. Can we merge this PR first, then work on a fix on repo-policy-compliance and change this line back?

jdkandersson: The issue is that curl will automatically do a post if you send data, in this case no data is being sent. So `-X POST` should fix that issue

jdkandersson: So in this case curl issued a GET which resulted in a method now allowed error

jdkandersson (approved):

tests/integration/test_charm_fork_repo.py
```diff
@@ -8,114 +8,48 @@
...
     forked_github_repository: Repository,
-    branch_with_unsigned_commit: Branch,
+    forked_github_branch: Branch,
 ) -> None:
     """
     arrange:
-        1. A forked repository with unsigned commit in default branch.
+        1. A forked repository.
         2. A working application with one runner on the forked repository.
     act: Trigger a workflow dispatch on a branch in the forked repository.
```
cbartz, line 37:
```suggestion
    act: Trigger a workflow dispatch that fails the repo policy check on a branch in the forked repository.
```

cbartz (approved):
LGTM, small comment, but this can be fixed later.

PR #157, opened 2023-11-29
COS: Clean up dashboard

cbartz:
Applicable spec: n/a

### Overview

Clean up the dashboard :

1. Remove the `Active Runner` gauge and rename the graph to `Idle Runners after Reconciliation`.
2. Display only whole numbers (integers) for the `Idle Runners after Reconciliation` graph.
3. Adjust range aggregation in  `Idle Runners after Reconciliation` to `60m` instead of `10m` for better reconciliation event consideration.
4. Include offline but healthy runners in the `Idle Runners` count.
5. Showcase `Lifecycle Status` graph counters as monotonically increasing within the selected range, disregarding values outside it.

Screenshot from 2023-12-01 08-05-25

### Rationale

1. **Clearer Representation**:

   - Focused on observing overall trends of idle runners for user provisioning.
   - Active Runner display during reconciliation can mislead user interpretation.

2. **Enhanced Visualization**: Improved visual representation for better user comprehension.

3. **Optimized Time Range**: `10m` range leads to missing values due to our 30-minute reconciliation period in our deployment. Upgrading to `60m` better aligns with the latest reconciliation events for units in normal scenarios.

4. **Improved Idle Runner Detection**: Corrects issues where idle runners weren't accurately detected post-spawning due to API detection delays.

5. **Relevant Data Presentation**: Focus on presenting pertinent data within the selected range, catering to user interests.

### Module Changes

Modify the `runner_manager` module to calculate the metrics accordingly.

### Checklist


jdkandersson:
> @mthaddon @jdkandersson you may want to review this PR. I have decided not to detect active runners between reconciliation events, as this could lead to confusing visualisation (even if they are no longer active, they would still be shown as active in the panel until the next reconciliation event updates the status).

Isn't that also the case for active runners, i.e., the active runners could finish their job within seconds of us reading the metric and then they would be counted as active until the next reconciliation event even though that information is outdated quite quickly. That metric should be interpreted as the number of runners that were active within the reconciliation window

jdkandersson (approved):
Looks good, it would be good to get this into the `edge` deploy on Monday. Whether or not the active stays I'll leave up to you, there isn't a clear use case for it at this point and it looks like it would be easy to add back in

cbartz:
> Looks good, it would be good to get this into the edge deploy on Monday. Whether or not the active stays I'll leave up to you, there isn't a clear use case for it at this point and it looks like it would be easy to add back in

As there is no real use case (other than to see if active + idle adds up to the expected total or not), I fear that the metric will add more confusion than value to the dashboard viewer.

The panel is not about real-time data, and the user needs to understand that the data points refer to the sum of the data from the last reconciliation events on each of the units. I have added an explanation to the panel, but I think most dashboard users will simply ignore it.

mthaddon (approved):

PR #166, opened 2023-12-11
Remove race condition between restart repo-policy-compliance and request of the service

yhaliaw:
### Overview

On update of service, flush the runners, restart the services, then spawn runners to replace the flushed ones.
Prior to flushing the runners, wait until all running job ran at least 1 minute.

### Rationale

The race condition can cause the pre-job stage to fail from time to time.

### Checklist


src/runner_manager.py
```diff
@@ -522,15 +523,54 @@ def reconcile(self, quantity: int, resources: VirtualMachineResources) -> int:
...
+        Returns:
+            Whether there are runners that has taken a job and run for less than 1 minute.
+        """
+        now = datetime.now(timezone.utc)
+        busy_runners = [
+            runner for runner in self._get_runners() if runner.status.exist and runner.status.busy
+        ]
+        for runner in busy_runners:
+            exit_code, stdout, _ = runner.instance.execute(
+                ["/usr/bin/stat", "-c", "'%w'", "/home/ubuntu/github-runner/_work"]
```
jdkandersson, line 540: What does this do please? I'm guessing it reads our metrics logs?

src/runner_manager.py
```diff
@@ -522,15 +523,54 @@ def reconcile(self, quantity: int, resources: VirtualMachineResources) -> int:
...
         Args:
-            flush_busy: Whether to flush busy runners as well.
+            flush_busy: Whether to flush busy runners.
+            wait_repo_check: Whether to wait for busy runner to complete
+                repo-policy-compliance check before flushing the runners.
 
         Returns:
             Number of runners removed.
         """
+        if wait_repo_check:
```
jdkandersson, line 561: Should the non-busy runners be flushed first so that no new runners are spawned whilst we wait for existing runners to pass their repo-policy-check phase?

src/runner_manager.py
```diff
@@ -522,15 +523,54 @@ def reconcile(self, quantity: int, resources: VirtualMachineResources) -> int:
...
+            for _ in range(5):
```
jdkandersson, line 562: Theoretically this should at most take 1 minute, right? In the worst case where a runner takes a job just as we start the flush?

src/runner_manager.py
yhaliaw: When the runner takes a job, the `_work` directory is created.
This command checks the file creation date for that directory to determine when the job started.

I will add an code comment.

yhaliaw: This is done because it takes a while for GitHub API to update the status some times.

src/runner_manager.py
yhaliaw: Yes, currently the max should be slightly more than 1 minute as the query to repo-policy-compliance has a timeout of 60 secs. 
In theory, 90 secs (3 * 30) of wait should cover all cases right now. I set it as (5 * 30).

src/runner_manager.py
yhaliaw: Updated the comments.

src/runner_manager.py
yhaliaw: Updated to remove the idle runners first.

tests/integration/test_charm_one_runner.py
```diff
@@ -189,6 +197,70 @@ async def test_reconcile_runners_with_lxd_storage_pool_failure(
...
+            continue
+
+        if f"Job is about to start running on the runner: {app_runner.name}-" in logs:
+            assert run.jobs()[0].conclusion == "success"
+            assert (
+                "A private docker registry is setup as a dockerhub mirror for this self-hosted"
+                " runner."
+            ) in logs
+
+
```
yanksyoon, line 263: I think we can use functions defined in `helpers.py` to get the logs. 

https://github.com/canonical/github-runner-operator/blob/main/tests/integration/test_debug_ssh.py#L36-L56
Here's an example of fetching logs from a workflow dispatched run.

tests/integration/test_charm_one_runner.py
```diff
@@ -189,6 +197,70 @@ async def test_reconcile_runners_with_lxd_storage_pool_failure(
...
+    runner_manager_github_client: GithubClient,
+) -> None:
+    """
+    arrange: A working application with no runners.
+    act:
+        1. Set dockerhub-mirror config and spawn one runner.
+        2. Dispatch a workflow.
+    assert:
+        1. registry-mirrors is setup in /etc/docker/daemon.json of runner.
+        2. Message about dockerhub_mirror appears in logs.
```
yanksyoon, line 213: May I ask how this checks that the busy_runners have been waited?

yanksyoon (dismissed):
LGTM, a few questions, thanks!

src/charm.py
```diff
@@ -583,8 +583,8 @@ def _check_and_update_dependencies(self) -> bool:
 
             self.unit.status = MaintenanceStatus("Flushing runners due to updated deps")
 
+            runner_manager.flush(wait_repo_check=True)
```
cbartz, line 586: Do we really want to flush busy runners when the runner binary has been updated?

src/charm.py
```diff
@@ -485,7 +485,7 @@ def _on_upgrade_charm(self, _event: UpgradeCharmEvent) -> None:
         if not runner_manager:
             return
 
-        runner_manager.flush()
+        runner_manager.flush(wait_repo_check=True)
```
cbartz, line 488: Reading the code here without digging into the flush docstring, I would not expect flush to flush busy runners. Could we perhaps just do inside flush what the name of the kwarg suggests, i.e. just wait, and only flush if the `force_flush_busy` kwarg is additionally passed? This might be redundant, but would improve readability.

tests/integration/test_charm_one_runner.py
```diff
@@ -189,6 +197,70 @@ async def test_reconcile_runners_with_lxd_storage_pool_failure(
...
+    model: Model,
+    app_runner: Application,
+    github_repository: Repository,
+    runner_manager_github_client: GithubClient,
+) -> None:
+    """
+    arrange: A working application with no runners.
+    act:
+        1. Set dockerhub-mirror config and spawn one runner.
+        2. Dispatch a workflow.
```
cbartz, line 210: .It might be worth mentioning here that the workflow is designed to wait,

tests/integration/test_charm_one_runner.py
yhaliaw: Good idea. I will update the code.

src/charm.py
jdkandersson: Probably it is fine not to

tests/integration/test_charm_one_runner.py
yhaliaw: I have changed the test.
The test was outdated.

src/runner_manager_type.py
```diff
@@ -16,6 +17,17 @@
 from runner_type import GithubPath
 
 
+class FlushMode(Enum):
+    """Strategy for flushing runners."""
+
+    FLUSH_IDLE = 0
+    """Flush only idle runners."""
```
jdkandersson, line 24: Shouldn't this go into the docstring? I'm guessing that our linting will complain that these attributes are not explained in the class docstring

src/runner_manager_type.py
```diff
@@ -16,6 +17,17 @@
 from runner_type import GithubPath
 
 
+class FlushMode(Enum):
```
jdkandersson, line 20: This is a good change!

src/charm.py
yhaliaw: I have changed to use an enum as input. The enum covers different types of strategies.

src/charm.py
yhaliaw: Changed to not flush busy runners.

src/runner_manager_type.py
yhaliaw: Updated to the class docstring.

tests/integration/test_self_hosted_runner.py
```diff
@@ -92,3 +97,86 @@ async def test_dispatch_workflow_with_dockerhub_mirror(
...
+    for _ in range(30):
+        all_runners = runner_manager_github_client.get_runner_github_info(
+            GithubRepo(
+                owner=forked_github_repository.owner.login, repo=forked_github_repository.name
+            )
+        )
+        runners = [runner for runner in all_runners if runner.name == runner_to_be_used]
+
+        if not runners:
+            # if runner is not online yet.
```
cbartz, line 153: Should we add a `sleep` here?

tests/integration/test_charm_one_runner.py
```diff
@@ -187,8 +187,6 @@ async def test_reconcile_runners_with_lxd_storage_pool_failure(
     await wait_till_num_of_runners(unit, 1)
 
 
-@pytest.mark.asyncio
-@pytest.mark.abort_on_fail
```
cbartz, line 191: Why here in particular?

tests/integration/test_self_hosted_runner.py
```diff
@@ -92,3 +97,86 @@ async def test_dispatch_workflow_with_dockerhub_mirror(
                 "A private docker registry is setup as a dockerhub mirror for this self-hosted"
                 " runner."
             ) in logs
+
+
+@pytest.mark.asyncio
+@pytest.mark.abort_on_fail
+async def test_wait_on_busy_runner_repo_check(
```
cbartz, line 104: Isn't this a general test to flush out busy runners? This test would also make sense without the pre job check, right?
Otherwise, please add some more description in the test to make it more explicit for readers of the code.

cbartz (dismissed):
Looks good, but have some questions.

tests/integration/test_self_hosted_runner.py
yhaliaw: Added.

tests/integration/test_self_hosted_runner.py
yhaliaw: I was looking into a way to test the wait on pre-job. But reading the logs from GitHub does not work. As the logs retrieved by GitHub API are the logs upload to after the job completes, the flushed runner will not be doing that step.
So I have update the to reflect this test is for testing flushing busy runners.

cbartz (dismissed):

yanksyoon (dismissed):
Awesome flush modes!

src/runner.py
```diff
@@ -229,7 +230,12 @@ def remove(self, remove_token: str) -> None:
             self.status.runner_id,
             self.config.path.path(),
         )
-        self._clients.github.delete_runner(self.config.path, self.status.runner_id)
+        try:
+            self._clients.github.delete_runner(self.config.path, self.status.runner_id)
+        except HTTPError:
+            logger.exception("Unable the remove runner on GitHub: %s", self.config.name)
```
yanksyoon, line 236: Would it help to log the exception as well?
```suggestion
        except HTTPError as exc:
            logger.exception("Unable the remove runner on GitHub: %s, exception: %s", self.config.name, exc)
```

tests/integration/test_self_hosted_runner.py
```diff
@@ -92,3 +95,87 @@ async def test_dispatch_workflow_with_dockerhub_mirror(
...
+            GithubRepo(
+                owner=forked_github_repository.owner.login, repo=forked_github_repository.name
+            )
+        )
+        runners = [runner for runner in all_runners if runner.name == runner_to_be_used]
+
+        if not runners:
+            # if runner is not online yet.
+            sleep(30)
+            continue
```
yanksyoon, line 153: I'm wondering if this could benefit in readability by using `helpers.wait_for` method.

yanksyoon: The same with timeout errors below.

yanksyoon (dismissed):
Minor nitpicks, LGTM!

cbartz (dismissed):

src/runner.py
yhaliaw: The `logger.exception` would log the stack trace of the exception by default.

tests/integration/test_self_hosted_runner.py
yhaliaw: For this one, I feel it might be clearer like this.

The alternative is to have line 143 - 159 as a function the `wait_for` .
This would involve using `functools.partial` or defining a inner function in the scope to load the `runner_to_be_used` to that function.

I feel that might be over complexing things.

yanksyoon (dismissed):

cbartz (dismissed):

yanksyoon (approved):

cbartz (approved):

PR #174, opened 2023-12-18
Sync disk prior to publishing image

yhaliaw:
Applicable spec: <link>

### Overview

During building of runner image, run`/usr/bin/sync` prior to stopping the VM instance and publishing the VM as image.

### Rationale

Not running the `sync` can result in missing/broken file in the published image.

### Checklist

### DIFF
--- a/scripts/build-image.sh
+++ b/scripts/build-image.sh
@@ -96,13 +96,14 @@ else
     echo "Unsupported CPU architecture: $(uname -m)"
     return 1
 fi
-/usr/bin/wget "https://github.com/mikefarah/yq/releases/latest/download/yq_linux_$YQ_ARCH"
-/usr/bin/wget https://github.com/mikefarah/yq/releases/latest/download/checksums
-/usr/bin/wget https://github.com/mikefarah/yq/releases/latest/download/checksums_hashes_order
-/usr/bin/wget https://github.com/mikefarah/yq/releases/latest/download/extract-checksum.sh
+/usr/bin/wget "https://github.com/mikefarah/yq/releases/latest/download/yq_linux_$YQ_ARCH" -O "yq_linux_$YQ_ARCH"
+/usr/bin/wget https://github.com/mikefarah/yq/releases/latest/download/checksums -O checksums
+/usr/bin/wget https://github.com/mikefarah/yq/releases/latest/download/checksums_hashes_order -O checksums_hashes_order
+/usr/bin/wget https://github.com/mikefarah/yq/releases/latest/download/extract-checksum.sh -O extract-checksum.sh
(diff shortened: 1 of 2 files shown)


yhaliaw:
Tested in proxied environment and non-proxied environment, the runner created has the `yq` binary inside.

scripts/build-image.sh
```diff
@@ -96,13 +96,14 @@ else
     echo "Unsupported CPU architecture: $(uname -m)"
     return 1
 fi
-/usr/bin/wget "https://github.com/mikefarah/yq/releases/latest/download/yq_linux_$YQ_ARCH"
-/usr/bin/wget https://github.com/mikefarah/yq/releases/latest/download/checksums
-/usr/bin/wget https://github.com/mikefarah/yq/releases/latest/download/checksums_hashes_order
-/usr/bin/wget https://github.com/mikefarah/yq/releases/latest/download/extract-checksum.sh
+/usr/bin/wget "https://github.com/mikefarah/yq/releases/latest/download/yq_linux_$YQ_ARCH" -O "yq_linux_$YQ_ARCH"
```
yanksyoon, line 99: May I ask if this is for style consistency since this ends up producing the same output binary name? (same for below)

scripts/build-image.sh
```diff
@@ -96,13 +96,14 @@ else
...
+/usr/bin/wget https://github.com/mikefarah/yq/releases/latest/download/checksums -O checksums
+/usr/bin/wget https://github.com/mikefarah/yq/releases/latest/download/checksums_hashes_order -O checksums_hashes_order
+/usr/bin/wget https://github.com/mikefarah/yq/releases/latest/download/extract-checksum.sh -O extract-checksum.sh
 /usr/bin/bash extract-checksum.sh SHA-256 "yq_linux_$YQ_ARCH" | /usr/bin/awk '{print $2,$1}' | /usr/bin/sha256sum -c | /usr/bin/grep OK
 /snap/bin/lxc file push "yq_linux_$YQ_ARCH" builder/usr/bin/yq --mode 755
 
+/snap/bin/lxc exec builder -- /usr/bin/sync
```
yanksyoon, line 106: I'm wondering if it would be better to sync target folders and files rather than the whole filesystem :)

yanksyoon (approved):
LGTM, minor questions! thanks :D

scripts/build-image.sh
yhaliaw: I think it is better to ensure all file are sync, since the installation from `apt` and configuration change also made modification to files. Previously, this issue only pop up in `yq` binary probably due to it is a larger file and it is the last file being modified.

scripts/build-image.sh
yhaliaw: Specifying the output with the `-O` option (upper-case O), will make wget to overwrite the file if it is there. Without this option, wget would be appending `.1`, `.2`, ... to the filename after the first run to prevent overwriting existing files.

PR #175, opened 2023-12-19
Add integration test with external network disabled and aproxy setup

yhaliaw:
### Overview

For setup of a integration test, deploy a juju machine and disable external network access in it, then deploy the charm within the juju machine with aproxy setup.

Use tinyproxy instead of proxy.py.

### Rationale

This test whether the charm works in an environment with HTTP proxy.
There were previous bugs related to not all traffic is routed through the proxy.

Tinyproxy is written in C and much faster than proxy.py. Since the deployment (juju, downloading image, installing deps) will be using the proxy, switching the a faster proxy seems to reduce the chance of failure.

### Checklist


src/charm.py
```diff
@@ -324,8 +324,6 @@ def _on_install(self, _event: InstallEvent) -> None:
         """
         self.unit.status = MaintenanceStatus("Installing packages")
 
-        self._update_kernel()
```
yhaliaw, line 327: The reconcile in the `_on_start` will still be executing this method.
The `_on_install` is taking too long, therefore removing this call as it will get executed again in `_on_start`.

jdkandersson (dismissed):

cbartz (commented):
Looks good, one nitpicky comment

src/charm.py
```diff
@@ -672,6 +672,9 @@ def _reconcile_runners(self, runner_manager: RunnerManager) -> Dict[str, Any]:
         delta_virtual_machines = runner_manager.reconcile(
             virtual_machines, virtual_machines_resources
         )
+
+        self.unit.status = ActiveStatus()
+
```
yanksyoon, line 677: Should this be set by the controller function that calls `_reconcile_runners`?

tests/integration/test_charm_with_proxy.py
```diff
@@ -40,26 +52,87 @@ async def proxy_fixture() -> AsyncIterator[str]:
...
+    await machine.ssh("sudo iptables -I OUTPUT -d 192.168.0.0/16 -j ACCEPT")
+    await machine.ssh("sudo iptables -I OUTPUT -d 198.18.0.0/15 -j ACCEPT")
+    await machine.ssh("sudo iptables -I OUTPUT -d 198.51.100.0/24 -j ACCEPT")
+    await machine.ssh("sudo iptables -I OUTPUT -d 203.0.113.0/24 -j ACCEPT")
+    await machine.ssh("sudo iptables -I OUTPUT -d 224.0.0.0/4 -j ACCEPT")
+    await machine.ssh("sudo iptables -I OUTPUT -d 233.252.0.0/24 -j ACCEPT")
+    await machine.ssh("sudo iptables -I OUTPUT -d 240.0.0.0/4 -j ACCEPT")
+    await machine.ssh("sudo iptables -P OUTPUT DROP")
+    # Test the external network access is disabled.
+    await machine.ssh("ping -c1 canonical.com 2>&1 | grep '100% packet loss'")
```
yanksyoon, line 114: I'm wondering if the output for this should be `asserted`?

yanksyoon (approved):
LGTM, minor questions!

src/charm.py
yhaliaw: This function was already setting the MaintenanceStatus at line 664.
This change is to ensure status is set back to active.

When creating this function I was debating whether the controller function(top-level charm event handler functions) should be the one setting the status. However, this results in three place of the code having something like:
```
self.unit.status = MaintenanceStatus("Reconciling")
self._reconcile_runners()
```
So I decided that it might be better if helper function within `charm.py` was allow to set the status as well.

tests/integration/test_charm_with_proxy.py
yhaliaw: The `machine.ssh` will throw exception on non-zero exit code.
See `if process.returncode != 0:` in:
https://pythonlibjuju.readthedocs.io/en/latest/_modules/juju/machine.html#Machine.ssh

This is why the assertion logic is within the `ping -c1 canonical.com 2>&1 | grep '100% packet loss'` command.

jdkandersson (approved):

cbartz (approved):

PR #187, opened 2024-01-11
Add mount point checks for shared file systems

cbartz:
### Overview

Add mountpoint checks before accessing the shared file system.

### Rationale

On reboot, all shared file systems are unmounted. Make the code more robust against unmounted shared file systems and ensure that metrics from runners completed before the reboot are processed.

### Checklist


src/runner_manager.py
```diff
@@ -281,16 +281,26 @@ def _create_runner(
...
+                try:
+                    (fs.path / RUNNER_INSTALLED_TS_FILE_NAME).write_text(
+                        str(ts_after), encoding="utf-8"
+                    )
+                except FileNotFoundError:
+                    logger.exception(
+                        "Failed to write runner-installed.timestamp into shared filesystem "
+                        "for runner %s, will not be able to issue all metrics.",
+                        runner.config.name,
+                    )
```
jdkandersson, line 302: Is there anything we can do to recover in these cases? Like re-create the file system or something?

tests/unit/test_shared_fs.py
jdkandersson, line 1: Is there a reason to not use an integration test? Maybe it is difficult to have these situations come about in integration test? Perhaps it is worth documenting the scenarios we cannot easily reproduce during testing but that we encountered in production so we don't lose knowledge of them. Perhaps a good place to document that is somewhere near the unit tests?

cbartz: I added an integration test. There are already a lot of tests in `test_charm_metrics.py`, most of them using runners to execute a wf, which leads to a quite long execution time compared to the other integration test modules. We might want to split this module into several to take advantage of concurrency.

src/runner_manager.py
cbartz: We could try to recreate the fs, but we cannot reattach a disk to a running lxd instance, so we would have to stop and restart the lxd instance and re-register the runner. This also means increasing the complexity of the code, which I am not sure is worth it. Also, if creating the shared fs failed the first time, it is likely that a second attempt will fail as well.
Before this PR, catching the exception was not implemented at all, so we have not seen this issue in production yet.

merkata (approved):

jdkandersson (approved):

PR #191, opened 2024-01-16
Increase reconcile timeout

yhaliaw:
### Overview

Increase reconcile timeout length.

### Rationale

Reconcile event timeout can be longer than the default which was half of the event interval.

### Checklist

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -163,7 +163,6 @@ class GithubRunnerCharm(CharmBase):
             path=self.config["path"],  # for detecting changes
             token=self.config["token"],  # for detecting changes
             runner_bin_url=None,
-            runner_image_url=None,
         )
 
         self.proxies: ProxySetting = {}
--- a/src/charm.py
+++ b/src/charm.py
@@ -499,7 +499,9 @@ class GithubRunnerCharm(CharmBase):
         self._refresh_firewall()
         try:
             self._event_timer.ensure_event_timer(
-                "reconcile-runners", self.config["reconcile-interval"]
+                event_name="reconcile-runners",
+                interval=int(self.config["reconcile-interval"]),
+                timeout=int(self.config["reconcile-interval"]) - 1,
             )
         except TimerEnableError as ex:
             logger.exception("Failed to start the event timer")
--- a/src/event_timer.py
+++ b/src/event_timer.py
@@ -74,11 +74,16 @@ class EventTimer:
         Raises:
             TimerEnableError: Timer cannot be started. Events will be not emitted.
         """
+        if timeout is not None:
+            timeout_in_secs = timeout * 60
+        else:
+            timeout_in_secs = interval * 30
+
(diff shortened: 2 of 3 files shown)


src/charm.py
```diff
@@ -499,7 +499,9 @@ def _on_config_changed(self, _event: ConfigChangedEvent) -> None:
         self._refresh_firewall()
         try:
             self._event_timer.ensure_event_timer(
-                "reconcile-runners", self.config["reconcile-interval"]
+                event_name="reconcile-runners",
+                interval=int(self.config["reconcile-interval"]),
```
yhaliaw, line 503: The conversion to int and validation of the `reconicle-interval` configuration, will be done in a future PR where all configuration in the `charm.py` is refactor to `charm_state.py`.

src/charm.py
```diff
@@ -499,7 +499,9 @@ def _on_config_changed(self, _event: ConfigChangedEvent) -> None:
...
+                timeout=int(self.config["reconcile-interval"]) - 1,
```
cbartz, line 504: Is it really safe to assume that we do not need a pause of more than one second between two reconciliations?

cbartz (dismissed):
Left a question.

src/charm.py
yhaliaw: The unit is 1 minutes, which I assume is enough.

cbartz: For interval, its minutes, for timeout, its actual seconds. 

```
man timeout

DURATION is a floating point number with an optional suffix: 's' for seconds (the default), 'm' for minutes, 'h' for hours or 'd' for days.  A duration of 0 disables the associated timeout
```

yhaliaw: I update the code to be more consistent.
The `interval` was in minutes, but the `timeout` was in seconds.
Now both are in minutes.

jdkandersson (approved):

yanksyoon (approved):

src/event_timer.py
```diff
@@ -69,15 +69,21 @@ def ensure_event_timer(self, event_name: str, interval: int, timeout: Optional[i
...
 
         Raises:
             TimerEnableError: Timer cannot be started. Events will be not emitted.
         """
+        if timeout is not None:
+            timeout_in_secs = timeout * 60
+        else:
+            timeout_in_secs = interval * 30
+
         context: EventConfig = {
```
yanksyoon, line 82: A nitpicky comment, the `EventConfig` definition could have attributes docstring so the time units can be determined :)

yanksyoon: Pleaes do feel free to resolve this for now!

cbartz (approved):

src/event_timer.py
yhaliaw: I have added the attribute docstring in the refactoring charm_state branch.

PR #196, opened 2024-01-17
COS: Add dashboard for long-term metrics

cbartz:
### Overview

Add a dashboard with a panel that visualises the number of jobs per day.

Screenshot from 2024-01-17 15-51-13

Screenshot from 2024-01-17 15-34-03

The panel would look like this with real data:

image

### Rationale

A separate dashboard is added to allow querying for longer time periods (this is not suitable for the other visualisations as too many data points are queried, resulting in empty panels with errors or high latency before anything is displayed). Note that the current loki-k8s charm uses the default setting of loki, which does not allow querying more than 30 days in the past.

### Checklist


jdkandersson (approved):

arturo-seijas (approved):

yanksyoon (approved):

jdkandersson:
> @jdkandersson The `Draft publish docs` job fails because of conflicting content. What do you suggest to solve this problem in this particular case?

The way to do it is to go to discourse and update the content to resolve the conflict. For example, by updating it to the intended content. Just make sure that if there were any community contributions, that we explain why we're changing to docs

PR #199, opened 2024-01-18
Refactor configuration to charm state

yhaliaw:
### Overview

Move charm configuration to charm state.

Removal of StoredSet will be done in a future PR.

### Rationale

Refactor to comply with Managing Charm Complexity.

### Module Changes

Move charm configuration to charm_state.py

### Checklist


src/charm.py
```diff
@@ -88,15 +93,11 @@ def func_with_catch_errors(self, event: EventT) -> None:
         except TokenError as err:
             logger.exception("Issue with GitHub token")
             self.unit.status = BlockedStatus(str(err))
-        except MissingConfigurationError as err:
```
yhaliaw, line 91: The `MissingConfigurationError` is now removed. Missing configuration will be managed by CharmState. The exception will eventually become a `ConfigurationError` now.

src/charm.py
```diff
@@ -205,6 +225,26 @@ def __init__(self, *args, **kargs) -> None:
         )
         self.framework.observe(self.on.update_status, self._on_update_status)
 
+    def setup_state(self) -> None:
+        """Set up the charm state."""
+        try:
+            self.state = State.from_charm(self)
+        except CharmConfigInvalidError as exc:
+            self.unit.status = ops.BlockedStatus(exc.msg)
+            raise ConfigurationError(exc.msg) from exc
```
yanksyoon, line 234: I'm wondering if `BlockedStatus` would be overridden by `ConfigurationError` (`ErrorStatus`).

src/charm.py
```diff
@@ -205,6 +225,26 @@ def __init__(self, *args, **kargs) -> None:
...
+                self.proxies["http"] = str(http_proxy)
+            if https_proxy := proxy_config.https:
+                self.proxies["https"] = str(https_proxy)
+            # there's no need for no_proxy if there's no http_proxy or https_proxy
+            no_proxy = proxy_config.no_proxy
+            if (https_proxy or http_proxy) and no_proxy:
+                self.proxies["no_proxy"] = no_proxy
+            if proxy_config.use_aproxy:
+                self.proxies["aproxy_address"] = proxy_config.aproxy_address
+
```
yanksyoon, line 247: Just out of curiosity, may I ask why proxy is handled specially?

src/charm.py
```diff
@@ -286,61 +326,32 @@ def _ensure_service_health(self) -> None:
...
-
-        Args:
-            token: GitHub personal access token to manager the runners with.
-            path: GitHub repository path in the format '<org>/<repo>', or the GitHub organization
-                name.
+        """Get a RunnerManager instance.
 
         Returns:
             An instance of RunnerManager.
         """
```
yanksyoon, line 335: The argument docstring seems to be removed, was this intended?

src/charm_state.py
```diff
@@ -30,6 +32,66 @@
 
 OPENSTACK_CLOUDS_YAML_CONFIG_NAME = "experimental-openstack-clouds-yaml"
 
+OPENSTACK_CLOUDS_YAML_CONFIG_NAME = "experimental-openstack-clouds-yaml"
```
yanksyoon, line 35: These lines seem to be duplicated :D

src/charm_state.py
```diff
@@ -30,6 +32,66 @@
...
+        path = GithubOrg(org=path_str, group=runner_group)
+    return path
+
+
+class VirtualMachineResources(NamedTuple):
+    """Virtual machine resource configuration."""
+
+    cpu: int
+    memory: str
+    disk: str
```
yanksyoon, line 93: I'm wondering if adding a comment/docstring about units would help avoid confusion here :D

src/charm_state.py
```diff
@@ -65,16 +127,68 @@ def __init__(self, msg: str):
...
+
+    Args:
+        size: Storage size string.
+
+    Return:
+        Whether the string is valid.
+    """
+    # Checks whether the string confirms to using the KiB, MiB, GiB, TiB, PiB,
+    # EiB suffix for storage size as specified in config.yaml.
+    valid_suffixes = ["KiB", "MiB", "GiB", "TiB", "PiB", "EiB"]
```
yanksyoon, line 141: I think these can be tuples / defined as constants :D

src/charm_state.py
```diff
@@ -65,16 +127,68 @@ def __init__(self, msg: str):
...
+    valid = any(suffix == size[-3:] for suffix in valid_suffixes)
```
yanksyoon, line 142: How about
```suggestion
    return size[-3:] in valid_suffixes and size[:-3].isdigit()
```

src/charm_state.py
```diff
@@ -114,7 +240,109 @@ def from_charm(cls, charm: CharmBase) -> "CharmConfig":
...
+        return cls(
+            virtual_machines=virtual_machines,
+            virtual_machine_resources=virtual_machine_resources,
+            runner_storage=runner_storage,
+        )
+
+    @root_validator
+    @classmethod
+    def check_fields(cls, values: dict) -> dict:
+        """Validate the runner configuration."""
```
yanksyoon, line 322: I think an argument docstring can help here :D

yanksyoon (commented):
Reviewed until `[src/runner_type.py](https://github.com/canonical/github-runner-operator/pull/199/files#diff-c20b0e214fe8be1aedddb401ae09dcbb42478db284aed1b08b5d1bca6658956a)`, thanks for the refactor! It looks great so far!

src/charm_state.py
```diff
@@ -114,7 +240,109 @@ def from_charm(cls, charm: CharmBase) -> "CharmConfig":
...
+        virtual_machines = values.get("virtual_machines")
+        resources = values.get("virtual_machine_resources")
+        # By property definition, this cannot be None. Needed for mypy type check.
+        assert virtual_machines is not None, "Unreachable code"  # nosec for [B101:assert_used]
+        assert resources is not None, "Unreachable code"  # nosec for [B101:assert_used]
+
```
jdkandersson, line 328: what about a `cast` from `typing` if it is primarily for mypy?

tests/unit/factories.py
```diff
@@ -46,3 +47,60 @@ class Meta:
     port: int = factory.LazyAttribute(lambda n: random.randrange(1024, 65536))
     rsa_fingerprint: str = factory.fuzzy.FuzzyText(prefix="SHA256:")
     ed25519_fingerprint: str = factory.fuzzy.FuzzyText(prefix="SHA256:")
+
+
+class MockGithubRunnerCharmUnitFactory(factory.Factory):
+    class Meta:
+        model = MagicMock
```
jdkandersson, line 54: What is the reason for using `MagicMock` here please? Usually you would set it as the class being created, you can get the types to work as demonstrated on discourse-gatekeeper: https://github.com/canonical/discourse-gatekeeper/blob/5c6f9bedadf2137b9e6f93ed2b50fcdb58fca084/tests/factories.py#L22

src/charm.py
yhaliaw: The `ConfigurationError` would be caught by `catch_charm_errors` declarator and be in `BlockedStatus`.
The line 233 is a bit unnecessary. I have removed that.

src/charm.py
yhaliaw: The usage of `self.proxies` was removed during refactor, but I forgot to delete this section.

src/charm.py
yhaliaw: Re-added. This args was removed then added back during the refactor.

src/charm_state.py
yhaliaw: Good idea. The following lines needs to be deleted as well.

src/charm_state.py
```diff
@@ -30,6 +32,66 @@
...
+    org: str
+    group: str
+
+    def path(self) -> str:
+        """Return a string representing the path."""
+        return self.org
+
+
+GithubPath = Union[GithubOrg, GithubRepo]
```
cbartz, line 82:
```suggestion
GithubPath = GithubOrg | GithubRepo
```

src/charm_state.py
```diff
@@ -114,7 +240,109 @@ def from_charm(cls, charm: CharmBase) -> "CharmConfig":
...
+
+    @root_validator
+    @classmethod
+    def check_fields(cls, values: dict) -> dict:
+        """Validate the general charm configuration."""
+        reconcile_interval = values.get("reconcile_interval")
+        # By property definition, this cannot be None. Needed for mypy type check.
+        assert reconcile_interval is not None, "Unreachable code"  # nosec for [B101:assert_used]
+
+        if reconcile_interval < 2:
```
cbartz, line 296: Why 2, is it a heuristic? This should also be added to the config description.

src/charm_state.py
```diff
@@ -114,7 +240,109 @@ def from_charm(cls, charm: CharmBase) -> "CharmConfig":
...
+
+        return values
+
+
+class RunnerCharmConfig(BaseModel):
+    """Runner configurations for the charm.
+
+    Attributes:
+        virtual_machines: Number of virtual machine-based runner to spawn.
+        virtual_machine_resources: Hardware resource use by one virtual machine for a runner.
```
cbartz, line 310:
```suggestion
        virtual_machine_resources: Hardware resource used by one virtual machine for a runner.
```

src/charm_state.py
```diff
@@ -65,16 +127,68 @@ def __init__(self, msg: str):
...
+
+        entry_list = [entry.strip() for entry in denylist_str.split(",")]
+        denylist = [FirewallEntry.decode(entry) for entry in entry_list if entry]
+        return denylist
+
+    @classmethod
+    def _parse_dockerhub_mirror(cls, charm: CharmBase) -> str | None:
+        dockerhub_mirror = charm.config.get("dockerhub-mirror") or None
+
+        dockerhub_mirror_url = urlsplit(dockerhub_mirror)
```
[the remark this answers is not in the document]
yhaliaw: I have tested that case, it would just return a empty object.
```
>>> urlsplit(None)
SplitResultBytes(scheme=b'', netloc=b'', path=b'', query=b'', fragment=b'')
```

src/charm_state.py
```diff
@@ -114,7 +240,109 @@ def from_charm(cls, charm: CharmBase) -> "CharmConfig":
...
+            openstack_clouds_yaml=openstack_clouds_yaml,
+        )
+
+    @root_validator
+    @classmethod
+    def check_fields(cls, values: dict) -> dict:
+        """Validate the general charm configuration."""
+        reconcile_interval = values.get("reconcile_interval")
+        # By property definition, this cannot be None. Needed for mypy type check.
+        assert reconcile_interval is not None, "Unreachable code"  # nosec for [B101:assert_used]
```
cbartz, line 294:
```suggestion
        reconcile_interval = values["reconcile_interval"]
```
If it cannot be None, this should work and mypy should not complain (at least I was able to run `tox -e lint` without errors).

src/charm_state.py
```diff
@@ -114,7 +240,109 @@ def from_charm(cls, charm: CharmBase) -> "CharmConfig":
...
+
+    @root_validator
+    @classmethod
+    def check_fields(cls, values: dict) -> dict:
+        """Validate the runner configuration."""
+        virtual_machines = values.get("virtual_machines")
+        resources = values.get("virtual_machine_resources")
+        # By property definition, this cannot be None. Needed for mypy type check.
+        assert virtual_machines is not None, "Unreachable code"  # nosec for [B101:assert_used]
+        assert resources is not None, "Unreachable code"  # nosec for [B101:assert_used]
```
cbartz, line 370:
```suggestion
        virtual_machines = values["virtual_machines"]
        resources = values.get["virtual_machine_resources"]
```

src/charm.py
```diff
@@ -126,18 +127,41 @@ def func_with_catch_errors(self, event: ActionEvent) -> None:
...
+    Returns:
+        Decorated charm function with charm state setup.
+    """
+
+    @functools.wraps(func)
+    def func_with_state_setup(self: "GithubRunnerCharm", event: EventT) -> None:
+        self.setup_state()
+        func(self, event)
+
+    return func_with_state_setup
```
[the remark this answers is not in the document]
yhaliaw: Updated, can you check if this version is what you expect?

src/charm_state.py
yhaliaw: Thanks, it worked.

tests/integration/test_charm_one_runner.py
```diff
@@ -239,7 +239,11 @@ async def test_token_config_changed_insufficient_perms(
...
 
-    await app.set_config({"token": "", "virtual-machines": "0"})
+    await app.set_config({"token": "invalid-token", "virtual-machines": "0"})
     await model.wait_for_idle()
 
     await wait_till_num_of_runners(unit, num=0)
+
+    # Add the runner back for the next test using the `app` deployment. Not part of the testing.
+    await app.set_config({"token": token, "virtual-machines": "1"})
+    await model.wait_for_idle(status=ACTIVE, timeout=10 * 60)
```
cbartz, line 249: Imho, this should be part of a setup/teardown fixture, see for example `test_charm_with_proxy` which has a setup fixture called `app` to ensure that the number of runners is zero.

cbartz (commented):
Up to unit tests

src/charm.py
```diff
@@ -361,38 +330,40 @@ def _get_runner_manager(
...
     @catch_charm_errors
-    @_setup_github_runner_charm_state
     def _on_install(self, _event: InstallEvent) -> None:
         """Handle the installation of charm.
 
         Args:
             event: Event of installing the charm.
         """
-        self.unit.status = MaintenanceStatus("Installing packages")
+        state = self._setup_state()
```
cbartz, line 348: As mentioned in the roadmap meeting, letting the install hook go into BlockedStatus (which can happen with an invalid config) may not be desirable, as the install hook will not be retriggered.

This may also apply to start hook.

cbartz: We can also create a bug ticket and resolve it later, as it is not strictly related to this PR.

src-docs/charm_state.py.md
```diff
@@ -208,7 +346,6 @@ The charm state.
  - <b>`arch`</b>:  The underlying compute architecture, i.e. x86_64, amd64, arm64/aarch64. 
  - <b>`charm_config`</b>:  Configuration of the juju charm. 
  - <b>`is_metrics_logging_available`</b>:  Whether the charm is able to issue metrics. 
- - <b>`openstack_clouds_yaml`</b>:  The openstack clouds.yaml configuration.
```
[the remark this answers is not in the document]
yhaliaw: I think so. The `openstack_clouds_yaml` is an attr on `CharmConfig` and not `State`. I think it was moved at some point during development and this entry in the docstring was not removed.

src/charm_state.py
yhaliaw: The reason is `EventTimer` sets a timeout of `reocncile_interval - 1` mintues. Therefore, the interval must be at least 2 or the timeout will be 0 mins or less.

This configuration is generally needs to be much larger than 2 minutes. So I am unsure if we need to document this in the `config.yaml`.

src/charm.py
yhaliaw: I have created a ticket for this bug.

src/charm.py
```diff
@@ -509,7 +527,9 @@ def _on_config_changed(self, _event: ConfigChangedEvent) -> None:
             self._stored.token = None
 
         if self.config["path"] != self._stored.path:
-            prev_config_for_flush["path"] = str(self._stored.path)
+            prev_config_for_flush["path"] = parse_github_path(
+                self._stored.path, self.config["group"]
```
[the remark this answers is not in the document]
yhaliaw: This would be need be changed when we remove `StoredSet`.
The `group` is only needed due to `self._stored.path` is a string. The `parse_github_path` is not needed when we use `CharmState` for persistence as it would directly provide a `GithubPath` object.
This is temp code until we remove `StoredSet`.

src/runner_manager_type.py
```diff
@@ -56,7 +55,8 @@ class RunnerManagerClients:
 
 
 @dataclass
-class RunnerManagerConfig:
+# The instance attributes are all required.
```
jdkandersson, line 58: Could any be grouped together into a sub class?

yanksyoon (dismissed):
LGTM

src/runner_manager_type.py
yhaliaw: I revert to having `CharmState` under the `RunnerManagerConfig`.

tests/unit/test_runner_manager.py
```diff
@@ -271,7 +278,7 @@ def test_reconcile_issues_runner_installed_event(
     runner_manager: RunnerManager,
     monkeypatch: MonkeyPatch,
     issue_event_mock: MagicMock,
-    charm_state: MagicMock,
+    charm_state: CharmState,
```
cbartz, line 281: Mhh, this change does not seem to be correct. My IDE now also complains with an err msg
```
'CharmState' object attribute 'is_metrics_logging_available' is read-only 
```
What is the reasoning for this?

cbartz (commented):
Mostly minor things

src/charm_state.py
```diff
@@ -278,26 +552,26 @@ def from_charm(cls, charm: CharmBase) -> list["SSHDebugConnection"]:
...
+class CharmState:
     """The charm state.
 
     Attributes:
         arch: The underlying compute architecture, i.e. x86_64, amd64, arm64/aarch64.
         charm_config: Configuration of the juju charm.
         is_metrics_logging_available: Whether the charm is able to issue metrics.
-        openstack_clouds_yaml: The openstack clouds.yaml configuration.
         proxy_config: Proxy-related configuration.
         ssh_debug_connections: SSH debug connections configuration information.
```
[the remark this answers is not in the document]
yhaliaw: The lint is not working due to this line:
https://github.com/canonical/github-runner-operator/blob/3ef9b0e8e45d28e6dc4daee2cb50bdbc81bd018f/.flake8#L4
We have a planned future PR (next week) to address this.

tests/unit/test_runner_manager.py
yhaliaw: This was error due to multiple refactor that removed then add back this arg.
It should be fixed now.

yanksyoon (dismissed):
LGTM!

tests/integration/conftest.py
```diff
@@ -289,7 +289,7 @@ async def app_no_wait_fixture(
         app_name=app_name,
         path=path,
         token=token,
-        runner_storage="memory",
+        runner_storage="juju-storage",
```
cbartz, line 292: Why the change here? @yanksyoon

yanksyoon: @cbartz 
There seems to be a problem with the config option `memory`, see this [workflow run](https://github.com/canonical/github-runner-operator/actions/runs/8043111866/job/21964781492#step:13:191).
I'm trying to fix the `StopIteration` error first by using `tzutc` and having a bit of padding in start time when querying for workflow runs.
This should be tackled once the `StopIteration` issue is fixed :D

jdkandersson (dismissed):

cbartz (dismissed):

jdkandersson (dismissed):

cbartz (dismissed):

tests/integration/helpers.py
```diff
@@ -408,7 +408,7 @@ def get_workflow_runs(
         latest_job: WorkflowJob = run.jobs()[0]
         logs = get_job_logs(job=latest_job)
 
-        if JOB_LOG_START_MSG_TEMPLATE.format(runner_name=runner_name) in logs:
+        if runner_name in logs:
```
cbartz, line 411: I believe that `JOB_LOG_START_MSG_TEMPLATE` is now unused? We really need to have a look at the linting, it seems to be either not triggered or misconfigured.

I assume that there cannot be any false positives with this method (so that a job has not yet started although the runner name is in the logs), as the string previously explicitly contained `Job is about to start on the runner:`

yhaliaw: It seems the `JOB_LOG_START_MSG_TEMPLATE` is used in one more place, which is why the linting is not triggered.

Yanks has mention he is looking into why some test are passing when `Job is about to start on the runner:` is not in the message anymore.

I do think fixing those tests are outside the scope of this PR.

cbartz (approved):

yanksyoon (approved):

PR #201, opened 2024-01-19
COS: Add more long-term metrics

cbartz:
### Overview

Add more panels to the dashboard to make it look like this:

Screenshot from 2024-01-19 15-40-32

Also, the CharmHub badge URL is fixed.

### Checklist


docs/reference/cos.md
```diff
@@ -26,7 +26,11 @@ The "GitHub Self-Hosted Runner Metrics" metrics dashboard presents the following
 
 The "GitHub Self-Hosted Runner Metrics (Long-Term)" metrics dashboard displays the following rows:
 
-- General: Displays a panel showing the number of jobs per day.
+- General: Contains the following panels:
+  - Total Jobs
+  - Total unique repositories
+  - Timeseries chart displaying the number of jobs per day
+  - Percentage of jobs with low queue time (less than 60 seconds)
```
arturo-seijas, line 33: Can you specify this in the panel title too?

arturo-seijas (commented):
Jobs per day might be better visualized as a bar chart instead of a line chart, which seems more suitable for visualizing trends

gtrkiller (dismissed):
LGTM

docs/reference/cos.md
cbartz: Just so I understand you correctly, it is already in the panel description, you would prefer it to be in the panel title too?

cbartz:
> Jobs per day might be better visualized as a bar chart instead of a line chart, which seems more suitable for visualizing trends

Like this (thats real data) ?
image

docs/reference/cos.md
arturo-seijas: Yes, it will be clearer to an operator what "low" means

arturo-seijas:
> > Jobs per day might be better visualized as a bar chart instead of a line chart, which seems more suitable for visualizing trends
> 
> @arturo-seijas Like this (thats real data) ? image

Yes, that's exactly what I had in mind

docs/reference/cos.md
cbartz: Have updated

arturo-seijas (dismissed):

arturo-seijas (approved):

jdkandersson (approved):

PR #202, opened 2024-01-19
test: add secondary tokens

yanksyoon:
Applicable spec: N/A

### Overview

Adds an option to have secondary tokens (CSV of PATs) to reduce chance of having rate limiting issues.

### Rationale

To reduce test failures due to rate limiting.

### Juju Events Changes

None.

### Module Changes

None.

### Library Changes

None.

### Checklist

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -11,7 +11,7 @@ def pytest_addoption(parser: Parser):
     """Add options to pytest parser."""
     parser.addoption("--path", action="store")
     parser.addoption("--token", action="store")
-    # A comma separated GitHub Personal Accesss Tokens, to help reduce rate limiting.
+    # A comma separated GitHub Personal Access Tokens, to help reduce rate limiting.
     parser.addoption("--secondary-tokens", action="store")
     parser.addoption("--charm-file", action="store")
     parser.addoption("--token-alt", action="store")
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -11,8 +11,11 @@ def pytest_addoption(parser: Parser):
     """Add options to pytest parser."""
     parser.addoption("--path", action="store")
     parser.addoption("--token", action="store")
-    # A comma separated GitHub Personal Access Tokens, to help reduce rate limiting.
-    parser.addoption("--secondary-tokens", action="store")
+    parser.addoption(
+        "--secondary-tokens",
+        action="store",
+        help="A comma separated GitHub Personal Access Tokens, to help reduce rate limiting.",
+    )
     parser.addoption("--charm-file", action="store")
     parser.addoption("--token-alt", action="store")
     parser.addoption("--http-proxy", action="store")
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -10,12 +10,11 @@ from pytest import Parser
 def pytest_addoption(parser: Parser):
     """Add options to pytest parser."""
     parser.addoption("--path", action="store")
-    parser.addoption("--token", action="store")
     parser.addoption(
-        "--tokens",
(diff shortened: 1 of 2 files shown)


tests/integration/conftest.py
```diff
@@ -95,7 +96,10 @@ def token(pytestconfig: pytest.Config) -> str:
     """Configured token setting."""
     token = pytestconfig.getoption("--token")
     assert token, "Please specify the --token command line option"
-    return token
+    secondary_tokens: str = pytestconfig.getoption("--secondary-tokens", default="")
+    tokens = secondary_tokens.split(",").append(token)
```
arturo-seijas, line 100: AFAIK `append` doesn't return a list. Consider 
```
tokens = secondary_tokens.split(",")
tokens.append(token)
```
instead

yanksyoon: You're very right, thanks for catching that, i've added additional parsing logic!

tests/integration/conftest.py
```diff
@@ -97,7 +97,10 @@ def token(pytestconfig: pytest.Config) -> str:
     token = pytestconfig.getoption("--token")
     assert token, "Please specify the --token command line option"
     secondary_tokens: str = pytestconfig.getoption("--secondary-tokens", default="")
-    tokens = secondary_tokens.split(",").append(token)
+    tokens = [
+        token.strip() for token in (secondary_tokens.split(",") if secondary_tokens != "" else [])
```
arturo-seijas, line 101: "".split(",") will already return an empty array, so no need for an if else here

yanksyoon: `"".split(",")` returns a list with an item, i.e. [""] unfortunately :/

arturo-seijas: Oops, you're right. My bad

arturo-seijas (dismissed):

tests/conftest.py
```diff
@@ -11,6 +11,8 @@ def pytest_addoption(parser: Parser):
     """Add options to pytest parser."""
     parser.addoption("--path", action="store")
     parser.addoption("--token", action="store")
+    # A comma separated GitHub Personal Access Tokens, to help reduce rate limiting.
+    parser.addoption("--secondary-tokens", action="store")
```
jdkandersson, line 15: What about making the token optionally a comma separated list so that we could use `N` tokens? Just split it on `,` or something (not sure if that could be in the token) or something and that will mean that if 1 or more tokens can be supplied

yanksyoon: Yup, it currently can receive from 0 to N tokens :D
```python
tokens = [token.strip() for token in (secondary_tokens.split(",") if secondary_tokens else [])]
```

yanksyoon: Added another option `--tokens` so that we don't break tests in other PRs

yanksyoon: `--token` option is now optinally csv.

cbartz (commented):
I think we need to bind the token to the path or dynamically generate the path from the token. Currently we use a forked repo for many tests, which is in the @yhaliaw namespace. We cannot assume that our tokens have access to each other's namespaces.

EDIT: This is not necessary, as @yanksyoon  reminded me that the fork is created dynamically in the token owner's ns.

In addition, we may want to stop passing tokens via the `INTEGRATION_TEST_ARGS` secret, as this would mean sharing our tokens with each other, as someone has to fill in this value.

arturo-seijas (dismissed):

cbartz (dismissed):

jdkandersson (approved):

cbartz (approved):

PR #206, opened 2024-01-23
Use Github token in E2E tests

cbartz:
### Overview

Use the [GITHUB_TOKEN](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#about-the-github_token-secret) in E2E tests wherever possible.

### Rationale

Reduces the effects of the rate limit, as a [different rate limit](https://docs.github.com/en/rest/using-the-rest-api/rate-limits-for-the-rest-api?apiVersion=2022-11-28#primary-rate-limit-for-github_token-in-github-actions) applies to this token.

### Checklist

### DIFF
--- a/.github/workflows/e2e_test.yaml
+++ b/.github/workflows/e2e_test.yaml
@@ -131,7 +131,7 @@ jobs:
             ${{ steps.runner-name.outputs.name }} \
             --base ubuntu@22.04 \
             --config path=${{ secrets.E2E_TESTING_REPO }} \
-            --config token=${{ secrets.E2E_TESTING_TOKEN }} \
+            --config token=${{ secrets.GITHUB_TOKEN }} \
             --config virtual-machines=1 \
             --config denylist=10.0.0.0/8 \
             --config test-mode=insecure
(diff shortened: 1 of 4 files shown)


.github/workflows/e2e_test.yaml
```diff
@@ -96,6 +96,17 @@ jobs:
       - name: Install GitHub Cli
         run: which gh || sudo apt install gh -y
 
+      - name: Check rate limit
+        env:
+          GH_TOKEN: ${{ (matrix.event.name == 'issues' || matrix.event.name == 'schedule') && secrets.E2E_TESTING_TOKEN || secrets.GITHUB_TOKEN }}
```
yanksyoon, line 101: Im wondering if the GH_TOKEN should come before E2E_TESTING_TOKEN, so that it gets prioritised?

yanksyoon (approved):
LGTM, thanks!

.github/workflows/e2e_test.yaml
cbartz: Can you explain this? This code should ensure that the `E2E_TESTING_TOKEN` is selected for `issues` and `schedule` and otherwise the `GITHUB_TOKEN`.

weiiwang01 (approved):

.github/workflows/e2e_test.yaml
yanksyoon: Ah gotcha, thanks!

PR #209, opened 2024-01-26
feat: allow tmate ip addresses to firewall.

yanksyoon:
Applicable spec: N/A

### Overview

To allow tmate-ssh-server IP addresses from relationdata to be excluded from denylist by filtering them if from denylist if it belongs to one of the networks defined by the denylist.

### Rationale

Since many internal subnets can be blocked, including the one that the charm is related to (tmate-ssh-server), it is necessary to allow traffic between related units even though the subnet they belong in might be covered by the denylist.

### Juju Events Changes

On ssh-debug relation data change, the firewall is now updated.

### Module Changes

`firewall.py` now splits denylist if one of the ip ranges in allowlist is found within the range of denylist. (Allowlist has priority.)
`charm.py` now refreshes firewall on ssh-debug relation data changed.

### Library Changes

None.

### Checklist

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -910,8 +910,11 @@ class GithubRunnerCharm(CharmBase):
                 FirewallEntry.decode(entry.strip())
                 for entry in firewall_denylist_config.split(",")
             ]
+        allowlist = [
+            FirewallEntry.decode(str(entry.host)) for entry in self._state.ssh_debug_infos
+        ]
         firewall = Firewall("lxdbr0")
-        firewall.refresh_firewall(denylist)
+        firewall.refresh_firewall(denylist, allowlist)
         logger.debug(
             "firewall update, current firewall: %s",
             execute_command(["/usr/sbin/nft", "list", "ruleset"]),
@@ -930,6 +933,7 @@ class GithubRunnerCharm(CharmBase):
 
     def _on_debug_ssh_relation_changed(self, _: ops.RelationChangedEvent) -> None:
         """Handle debug ssh relation changed event."""
+        self._refresh_firewall()
         runner_manager = self._get_runner_manager()
         runner_manager.flush(flush_busy=False)
         self._reconcile_runners(runner_manager)
--- a/src/charm.py
+++ b/src/charm.py
@@ -232,16 +232,6 @@ class GithubRunnerCharm(CharmBase):
                 logger.info("Cleaned up storage directory")
             raise RunnerError("Failed to configure runner storage") from err
 
-    def _juju_storage_mounted(self) -> bool:
-        """Whether a juju storage is mounted on the runner-storage location.
-
-        Returns:
-            True if a juju storage is mounted on the runner-storage location.
-        """
-        # Use `mount` as Python does not have easy method to get the mount points.
-        stdout, exit_code = execute_command(["mountpoint", str(self.juju_storage_path)], False)
-        return exit_code == 0 and str(self.juju_storage_path) in stdout
-
     @retry(tries=5, delay=15, max_delay=60, backoff=1.5, local_logger=logger)
     def _ensure_runner_storage(self, size: int) -> Path:
         """Ensure the runner storage is setup.
@@ -258,15 +248,7 @@ class GithubRunnerCharm(CharmBase):
                 path = self.ram_pool_path
                 self._create_memory_storage(self.ram_pool_path, size)
             case RunnerStorage.JUJU_STORAGE:
-                logger.info("Verifying juju storage")
                 path = self.juju_storage_path
-                if not self._juju_storage_mounted():
-                    raise ConfigurationError(
-                        (
-                            "Non-root disk storage should be mount on the runner juju storage to "
-                            "be used as the disk for the runners"
-                        )
-                    )
 
         # tmpfs storage is not created if required size is 0.
         if size > 0:
--- a/src/charm.py
+++ b/src/charm.py
@@ -914,7 +914,7 @@ class GithubRunnerCharm(CharmBase):
             FirewallEntry.decode(str(entry.host)) for entry in self._state.ssh_debug_infos
         ]
         firewall = Firewall("lxdbr0")
-        firewall.refresh_firewall(denylist, allowlist)
+        firewall.refresh_firewall(denylist=denylist, allowlist=allowlist)
         logger.debug(
             "firewall update, current firewall: %s",
             execute_command(["/usr/sbin/nft", "list", "ruleset"]),
--- a/src/charm.py
+++ b/src/charm.py
@@ -911,7 +911,7 @@ class GithubRunnerCharm(CharmBase):
                 for entry in firewall_denylist_config.split(",")
             ]
         allowlist = [
-            FirewallEntry.decode(str(entry.host)) for entry in self._state.ssh_debug_infos
+            FirewallEntry.decode(str(entry.host)) for entry in self._state.ssh_debug_connections
         ]
         firewall = Firewall("lxdbr0")
         firewall.refresh_firewall(denylist=denylist, allowlist=allowlist)
--- a/src/firewall.py
+++ b/src/firewall.py
@@ -82,18 +82,18 @@ class Firewall:  # pylint: disable=too-few-public-methods
         Returns:
             The network pool without the network segments in excludes.
         """
-        total_excluded_networks = set(networks)
+        total_networks_without_excluded = set(networks)
 
         for exclude in excludes:
             scoped_excluded_networks: set[NetworkT] = set()
-            for net in total_excluded_networks:
+            for net in total_networks_without_excluded:
                 if net.overlaps(exclude):
                     scoped_excluded_networks.update(net.address_exclude(exclude))
                 else:
                     scoped_excluded_networks.add(net)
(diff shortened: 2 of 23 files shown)


src/charm.py
```diff
@@ -910,8 +910,11 @@ def _refresh_firewall(self):
                 FirewallEntry.decode(entry.strip())
                 for entry in firewall_denylist_config.split(",")
             ]
+        allowlist = [
+            FirewallEntry.decode(str(entry.host)) for entry in self._state.ssh_debug_infos
+        ]
         firewall = Firewall("lxdbr0")
-        firewall.refresh_firewall(denylist)
+        firewall.refresh_firewall(denylist, allowlist)
```
cbartz, line 917:
```suggestion
        firewall.refresh_firewall(denylist=denylist, allowlist=allowlist)
```
You might want to use kwargs as guards here to avoid using the wrong position ;-)

src/firewall.py
```diff
@@ -128,23 +146,24 @@ def refresh_firewall(self, denylist: typing.List[FirewallEntry]):
...
-                    }
-                    for ip in excluded
-                ]
-            )
+        allowed_ips = [
+            host_network,
+            *[ipaddress.IPv4Network(entry.ip_range) for entry in (allowlist or [])],
+        ]
+        ips_to_deny = [ipaddress.IPv4Network(entry.ip_range) for entry in denylist]
+        denied_ips = self._exclude_network(ips_to_deny, allowed_ips)
```
cbartz, line 156:
```suggestion
        denied_ips = self._exclude_network(networks=ips_to_deny, excludes=allowed_ips)
```

src/firewall.py
```diff
@@ -66,7 +66,25 @@ def get_host_ip(self) -> str:
...
+        excludes: typing.List[ipaddress.IPv4Network],
+    ) -> typing.Set[ipaddress.IPv4Network]:
+        total: typing.Set[ipaddress.IPv4Network] = set()
+        for exclude in excludes:
+            for network in networks:
+                try:
+                    total.update(network.address_exclude(exclude))
+                except ValueError:
+                    total.update([network])
+        return total  # this currently doesn't work
```
cbartz, line 81: Could you explain what you mean by that comment?

cbartz (dismissed):
LGTM, left some nitpicky comments.

src/charm.py
yanksyoon: fixed, thank you!

src/firewall.py
yanksyoon: Sorry it was a comment for me before I went to sleep - removed, thank you!

gtrkiller (commented):
Out of interest, is there a reason why this PR only addresses IPv4 and not IPv6? otherwise it LGTM

yanksyoon:
@gtrkiller Good point, added a generic type - but for concrete types - like host address, it's because we're using the command on L67: `"/snap/bin/lxc", "network", "get", self._network, "ipv4.address"` to get ipv4 addresses. Then everything has to be IPv4Network type to use the `overlaps` and `address_exclude` functions, since they can only be compared to network of same types. Thanks for the review! :D

gtrkiller (dismissed):

src/firewall.py
```diff
@@ -66,7 +68,38 @@ def get_host_ip(self) -> str:
...
+        """Excludes the network segment from a pool of networks.
+
+        Args:
+            networks: The network pool to apply.
+            excludes: The networks to exclude from the pool.
+
+        Returns:
+            The network pool without the network segments in excludes.
+        """
+        total_excluded_networks = set(networks)
```
cbartz, line 85: I would prefer something like `total_networks_without_excluded`, because one could think you would return the excluded networks.

cbartz (dismissed):
LGTM, one nitpicky comment

src/firewall.py
yanksyoon: That is indeed a much better var name choice, thank you!

cbartz (dismissed):

tests/unit/test_firewall.py
```diff
@@ -0,0 +1,97 @@
...
+            id="exclude a single IP and one different subnet",
+        ),
+        pytest.param(
+            [IPv4Network("198.18.0.0/15"), IPv4Network("172.16.0.0/12")],
+            [IPv4Network("10.136.12.36/32")],
+            {IPv4Network("198.18.0.0/15"), IPv4Network("172.16.0.0/12")},
+            id="no matching networks",
+        ),
+    ],
+)
```
yhaliaw, line 86: There can be tests with empty `excludes_ranges` and `domain_ranges`.

yhaliaw (dismissed):
Left a minor comment

src/charm.py
```diff
@@ -910,8 +910,11 @@ def _refresh_firewall(self):
                 FirewallEntry.decode(entry.strip())
                 for entry in firewall_denylist_config.split(",")
             ]
+        allowlist = [
+            FirewallEntry.decode(str(entry.host)) for entry in self._state.ssh_debug_infos
```
jdkandersson, line 914: minor point, information cannot be plural in english

src/firewall.py
```diff
@@ -66,7 +68,38 @@ def get_host_ip(self) -> str:
         )
         return str(ipaddress.IPv4Interface(address.strip()).ip)
 
-    def refresh_firewall(self, denylist: typing.List[FirewallEntry]):
+    def _exclude_network(
+        self,
+        networks: typing.List[NetworkT],
+        excludes: typing.List[NetworkT],
```
jdkandersson, line 74: Possibly rename to `exclude` since it already implies 1 or more

src/firewall.py
```diff
@@ -66,7 +68,38 @@ def get_host_ip(self) -> str:
         )
         return str(ipaddress.IPv4Interface(address.strip()).ip)
 
-    def refresh_firewall(self, denylist: typing.List[FirewallEntry]):
+    def _exclude_network(
+        self,
+        networks: typing.List[NetworkT],
```
jdkandersson, line 73: can this use the `list` type?

src/firewall.py
```diff
@@ -66,7 +68,38 @@ def get_host_ip(self) -> str:
...
+        excludes: typing.List[NetworkT],
+    ) -> typing.Set[NetworkT]:
```
jdkandersson, line 75: Can this be the `set` type?

src/firewall.py
jdkandersson: Can these be an `iterable`?

tests/unit/test_firewall.py
```diff
@@ -0,0 +1,97 @@
...
+import pytest
+
+from firewall import Firewall
+
+
+@pytest.mark.parametrize(
+    "domain_ranges, excludes_ranges, expected",
+    [
+        pytest.param(
+            [IPv4Network("10.0.0.0/8")],
```
jdkandersson, line 17: could these be tuples?

tests/unit/test_firewall.py
```diff
@@ -0,0 +1,97 @@
...
+
+from ipaddress import IPv4Network
+
+import pytest
+
+from firewall import Firewall
+
+
+@pytest.mark.parametrize(
+    "domain_ranges, excludes_ranges, expected",
```
jdkandersson, line 14: perhaps `exclude_ranges`

tests/unit/test_firewall.py
```diff
@@ -0,0 +1,97 @@
...
+"""Test cases for firewall module."""
+
+from ipaddress import IPv4Network
+
+import pytest
+
+from firewall import Firewall
+
+
+@pytest.mark.parametrize(
```
jdkandersson, line 13: Test cases:
* empty IP address range with no, one and multiple exclude
* single IP address range (so a range with just 1 IP address) with no, one and multiple exclude where for single and multiple one or more match including single IP and range for these
* single Ip address range with multiple IP addresses with no, single and multiple exclude where the exclude are single IP address or multiple IP address ranges
* similar for multiple IP address ranges

tests/unit/test_firewall.py
yanksyoon: I made them into lists, since we need to perform `append` operations, converting them to `list` will change the `IPvXNetwork` into `IPvXAddress` and fail the algo unfortunately.

src/firewall.py
yanksyoon: Changed into list, due to

> I made them into lists, since we need to perform append operations, converting them to list will change the IPvXNetwork into IPvXAddress and fail the algo unfortunately.

src/firewall.py
yanksyoon: Changed into list, due to

> I made them into lists, since we need to perform append operations, converting them to list will change the IPvXNetwork into IPvXAddress and fail the algo unfortunately.

src/firewall.py
yanksyoon: Renamed, thank you!

src/charm.py
yanksyoon: Renamed, thank you!

tests/unit/test_firewall.py
yanksyoon: Added, thank you!

jdkandersson (approved):

cbartz (approved):

PR #217, opened 2024-02-01
Ensure event timer is active

cbartz:
### Overview

Ensure that the event timer is active on update_status. Add a log if the timer is not active.

### Rationale

We found on one deployment that the event timer was not active / could not be reactivated, resulting in only update status being run and no reconciliation.

### Juju Events Changes
The `update_status` hook now checks for the timer. There is a major refactoring going on in https://github.com/canonical/github-runner-operator/pull/199, the location of the check may be adjusted.

### Checklist

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -680,7 +680,6 @@ class GithubRunnerCharm(CharmBase):
             event: Event of stopping the charm.
         """
         try:
-            self._event_timer.disable_event_timer("update-dependencies")
             self._event_timer.disable_event_timer("reconcile-runners")
         except TimerDisableError as ex:
             logger.exception("Failed to stop the timer")
--- a/src/charm.py
+++ b/src/charm.py
@@ -42,7 +42,7 @@ from errors import (
     RunnerError,
     SubprocessError,
 )
-from event_timer import EventTimer, TimerDisableError, TimerEnableError
+from event_timer import EventTimer, TimerDisableError, TimerEnableError, TimerStatusError
 from firewall import Firewall, FirewallEntry
 from github_type import GitHubRunnerStatus
 from runner import LXD_PROFILE_YAML
@@ -157,20 +157,25 @@ class GithubRunnerCharm(CharmBase):
             return
 
         self._event_timer = EventTimer(self.unit.name)
-        if not self._event_timer.is_active(RECONCILE_RUNNERS_EVENT):
-            logger.warning(
-                "Reconciliation event timer is not activated - "
-                "this should only happen on the first charm execution after a deployment "
-                "on a new machine"
-            )
-            try:
-                self._set_reconcile_timer()
-            except TimerEnableError as ex:
-                logger.exception("Failed to start the event timer")
-                self.unit.status = BlockedStatus(
-                    (f"Failed to start timer for regular reconciliation" f"checks: {ex}")
+        try:
+            reconcile_evt_timer_is_active = self._event_timer.is_active(RECONCILE_RUNNERS_EVENT)
+        except TimerStatusError:
+            logger.exception("Failed to check the timer status")
+        else:
+            if not reconcile_evt_timer_is_active:
+                logger.warning(
+                    "Reconciliation event timer is not activated - "
+                    "this should only happen on the first charm execution after a deployment "
+                    "on a new machine."
                 )
-                return
+                try:
+                    self._set_reconcile_timer()
+                except TimerEnableError as ex:
+                    logger.exception("Failed to start the event timer")
+                    self.unit.status = BlockedStatus(
+                        (f"Failed to start timer for regular reconciliation checks: {ex}")
+                    )
+                    return
 
         if LXD_PROFILE_YAML.exists():
             if self.config.get("test-mode") != "insecure":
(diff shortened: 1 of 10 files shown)


src/charm.py
```diff
@@ -154,11 +156,26 @@ def __init__(self, *args, **kargs) -> None:
             self.unit.status = ops.BlockedStatus(exc.msg)
             return
 
+        self._event_timer = EventTimer(self.unit.name)
+        if not self._event_timer.is_active(RECONCILE_RUNNERS_EVENT):
```
amandahla, line 160: What happens if this command for checking if is active fails?

src/charm.py
```diff
@@ -154,11 +156,26 @@ def __init__(self, *args, **kargs) -> None:
...
+            logger.warning(
+                "Reconciliation event timer is not activated - "
+                "this should only happen on the first charm execution after a deployment "
+                "on a new machine"
+            )
+            try:
+                self._set_reconcile_timer()
+            except TimerEnableError as ex:
+                logger.exception("Failed to start the event timer")
+                self.unit.status = BlockedStatus(
```
amandahla, line 170: Out of curiosity: is there an alarm/notification for this case?

src/charm.py
```diff
@@ -154,11 +156,31 @@ def __init__(self, *args, **kargs) -> None:
             self.unit.status = ops.BlockedStatus(exc.msg)
             return
 
+        self._event_timer = EventTimer(self.unit.name)
+        try:
+            reconcile_evt_timer_is_active = self._event_timer.is_active(RECONCILE_RUNNERS_EVENT)
```
yhaliaw, line 161: Can `reconcile_evt_timer_is_active` be rename to `reconcile_event_time_is_active` or `reconcile_time_is_active`? Or something like that?
I feel `evt` a common shorthand for `event`.

src/charm.py
```diff
@@ -154,11 +156,31 @@ def __init__(self, *args, **kargs) -> None:
...
+                    "Reconciliation event timer is not activated - "
+                    "this should only happen on the first charm execution after a deployment "
+                    "on a new machine."
+                )
+                try:
+                    self._set_reconcile_timer()
+                except TimerEnableError as ex:
+                    logger.exception("Failed to start the event timer")
+                    self.unit.status = BlockedStatus(
+                        (f"Failed to start timer for regular reconciliation checks: {ex}")
```
yhaliaw, line 176: Is the `(`, `)` needed?

src/charm.py
```diff
@@ -484,15 +514,11 @@ def _on_config_changed(self, _event: ConfigChangedEvent) -> None:
...
-                event_name="reconcile-runners",
-                interval=int(self.config["reconcile-interval"]),
-                timeout=int(self.config["reconcile-interval"]) - 1,
-            )
+            self._set_reconcile_timer()
         except TimerEnableError as ex:
             logger.exception("Failed to start the event timer")
             self.unit.status = BlockedStatus(
-                (f"Failed to start timer for regular reconciliation" f"checks: {ex}")
+                (f"Failed to start timer for regular reconciliation checks: {ex}")
```
yhaliaw, line 521: Same as above

tests/unit/test_event_timer.py
```diff
@@ -0,0 +1,44 @@
...
+from event_timer import EventTimer
+
+
+def test_is_active_true():
+    """
+    arrange: create an EventTimer object and mock exec command to return zero exit code
+    act: call is_active
+    assert: return True
+    """
+    evt = EventTimer(secrets.token_hex(16))
```
yhaliaw, line 15: I feel `event` is clearer as a variable name

tests/unit/test_event_timer.py
```diff
@@ -0,0 +1,44 @@
...
+    assert evt.is_active(secrets.token_hex(16))
```
yhaliaw, line 16: Maybe `evt.is_active(event_name=secrets.token_hex(16))`  makes it clearer that you are using a generated string as stub event name.
Same applies to other usage of `secrets.token_hex` in this file.

src/charm.py
cbartz: Thx, nice catch.

src/charm.py
cbartz: No, not at the moment. One could try to create an alert based on the log. But unfortunately there seems to be a bug in the grafana agent / loki charm that does not maintain the correct juju topology (so only log alerts can be defined for the grafana agent).

cbartz:
Moved the check to the `update_status` hook. Reason: If placed in init, the code will be called before the install hook (first hook is `runner_storage_attached`), and the code would enable the timer that starts the service directly, and since the charm is not yet prepared (e.g. repo policy is not installed) it will fail and timeout after the reconciliation interval (this will resolve itself but still delay the installation for 10 minutes).

merkata (dismissed):
lgtm

amandahla (dismissed):
LGTM

amandahla (dismissed):
LGTM

merkata (dismissed):

amandahla (approved):
Nice!

merkata (approved):

PR #223, opened 2024-02-13
Remove proxy vars when using aproxy

cbartz:
### Overview

Remove the proxy vars `http_proxy`, `https_proxy` and `no_proxy` in runner's `/etc/environment` and `.env` files if runner is configured to use aproxy.

Add integration tests that test the different behaviour when using http proxy with aproxy vs http proxy without aproxy.

### Rationale

These variables are no longer needed for the default ports 80 and 443 as aproxy handles the traffic accordingly. Furthermore, I have observed a case where the `no_proxy` setting was not respected as attended (traffic to 10.0.0.0/8 was redirected to squid proxy), resulting in a failed workflow, whereas the workflow would succeed without the env variables set.
 See also [this](https://about.gitlab.com/blog/2021/01/27/we-need-to-talk-no-proxy/) for problems with env variables.  Apparently Python does not respect CIDR blocks in the `no_proxy` setting, according to this blog post.

### Module Changes

- `charm_state.ProxyConfig` is used everywhere now (except in `runners.py`)
- `charm_state.ProxyConfig` also takes care of setting no_proxy accordingly (previously in `charm.py`)
- The `ProxySetting` for the runners is now set in the `RunnerManager`

### Checklist

### DIFF
--- a/src/runner.py
+++ b/src/runner.py
@@ -38,7 +38,13 @@ from errors import (
 from lxd import LxdInstance
 from lxd_type import LxdInstanceConfig
 from runner_manager_type import RunnerManagerClients
-from runner_type import GithubOrg, RunnerConfig, RunnerStatus, VirtualMachineResources
+from runner_type import (
+    GithubOrg,
+    ProxySetting,
+    RunnerConfig,
+    RunnerStatus,
+    VirtualMachineResources,
+)
 from utilities import execute_command, retry
 
 logger = logging.getLogger(__name__)
@@ -121,6 +127,10 @@ class Runner:
         self.instance = instance
 
         self._shared_fs: Optional[shared_fs.SharedFilesystem] = None
+
+        if aproxy_address := self.config.proxies.get("aproxy_address"):
+            # If aproxy is used, the proxy environment variables are not necessary.
+            self.config.proxies = ProxySetting(aproxy_address=aproxy_address)
 
         # If the proxy setting are set, then add NO_PROXY local variables.
         if self.config.proxies.get("http") or self.config.proxies.get("https"):
--- a/src/runner.py
+++ b/src/runner.py
@@ -622,9 +622,13 @@ class Runner:
         docker_client_proxy = {
             "proxies": {
                 "default": {
-                    "httpProxy": self.config.proxies.http or "",
-                    "httpsProxy": self.config.proxies.https or "",
-                    "noProxy": self.config.proxies.no_proxy or "",
+                    key: value
+                    for key, value in (
+                        ("httpProxy", self.config.proxies.http),
+                        ("httpsProxy", self.config.proxies.https),
+                        ("noProxy", self.config.proxies.no_proxy),
+                    )
+                    if value
                 }
             }
         }
--- a/src/runner.py
+++ b/src/runner.py
@@ -38,13 +38,7 @@ from errors import (
 from lxd import LxdInstance
 from lxd_type import LxdInstanceConfig
 from runner_manager_type import RunnerManagerClients
-from runner_type import (
-    GithubOrg,
-    ProxySetting,
-    RunnerConfig,
-    RunnerStatus,
-    VirtualMachineResources,
-)
+from runner_type import GithubOrg, RunnerConfig, RunnerStatus, VirtualMachineResources
 from utilities import execute_command, retry
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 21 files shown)


weiiwang01:
> @jdkandersson If we remove these variables, http(s) requests to non-standard ports will no longer go through the http proxy, as aproxy is only configured for traffic through standard ports (we use nftables, which is a network-layer firewall). We don't currently set these variables for the docker proxy when aproxy is active, so traffic from docker containers is already only handled by aproxy. We should decide how to proceed. I think it is cleaner to remove these variables and add documentation when a user's workflow requires access to a non-standard port and how to deal with it (by setting the http_proxy variable). As mentioned in my PR description, the current state can also break workflows (if the no_proxy variable is not interpreted as desired by an application). cc @weiiwang01 @gregory-schiano

I have tested with ngrm, which only uses aproxy without any proxy environment variables, and [our unit and integration tests](https://github.com/canonical/wordpress-k8s-operator/pull/191) are running smoothly. Since proxies are only necessary for services on the public internet, non-standard ports on the public internet appear to be a pretty edge case. We can modify the nftables firewall rules to log and block such connections, providing users with a way to become aware of and update their GitHub workflow to explicitly use the proxy. And if some applications need TLS on non `443` ports, I can add those ports to the `aproxy` as well.

One important optimization to consider, either as a subsequent step or in conjunction with this one, is to exclude Canonical services (such as archive.ubuntu.com and launchpad.net) from the `nftables` redirection for `aproxy`. We should already have a list of their IP addresses used in our infrastructure.

jdkandersson:
I think removing the environment variables is fine, the way we should go about it is to let people know we'll do this, e.g., in 2 weeks time (which should just coincide with the stable release so if we merge this now we would give 2 weeks notice essentially) and suggest that they update their workflows to remove any use of proxy environment variables. If we get any push back we can reconsider

src/runner.py
```diff
@@ -122,6 +128,10 @@ def __init__(
 
         self._shared_fs: Optional[shared_fs.SharedFilesystem] = None
 
+        if aproxy_address := self.config.proxies.get("aproxy_address"):
+            # If aproxy is used, the proxy environment variables are not necessary.
+            self.config.proxies = ProxySetting(aproxy_address=aproxy_address)
```
jdkandersson, line 133: I'm guessing that this has the side-effect of not setting the proxy environment variables? It is a bit confusing that in order to remove proxy settings we set something else. Perhaps take a look at the code to see whether we can make the decision of what environment variables we set for the proxy a bit more explicit?

src/runner.py
jdkandersson, line 1: Perhaps this will come shortly - should there be any tests? I would be surprised if nothing fails in our test suit?

src/runner.py
cbartz: I do a reset here to avoid mutating the original dict, as it could be reused anywhere in the code. But I'll update the code to avoid confusion.

src/runner.py
cbartz: Updated

cbartz:
@weiiwang01 Could you please review this PR? Thank you.

src/runner_manager.py
```diff
@@ -90,10 +89,15 @@ def __init__(
...
         self.session.mount("https://", adapter)
         if self.proxies:
+            proxy_dict = {
+                "http_proxy": self.proxies.http,
+                "https_proxy": self.proxies.https,
+                "no_proxy": self.proxies.no_proxy,
+            }
             # setup proxy for requests
-            self.session.proxies.update(self.proxies)
+            self.session.proxies.update(proxy_dict)
```
weiiwang01, line 98: `requests` proxy configuration is not keyed with `http_proxy` or `https_proxy`, it's keyed with `http` and `https`. And I don't remember there's a `no_proxy` for the proxy configuration.

See: https://requests.readthedocs.io/en/latest/user/advanced/#proxies
```python
import requests

proxies = {
  'http': 'http://10.10.1.10:3128',
  'https': 'http://10.10.1.10:1080',
}

requests.get('http://example.org', proxies=proxies)
```

cbartz: Thanks, good catch, seems to be wrongly refactored by me.

cbartz: requests seem to support `no_proxy` since 2.14.0 https://pypi.org/project/requests/2.17.2/

cbartz: After looking at the code again, I think setting the proxy vars here and below is no longer necessary, because we have lines 76-81 which set the env vars for the process and its childs, and requests and urllib seem to respect this. (Otherwise the integration tests should have failed too, because I used the wrong vars here).

@yhaliaw wdyt?

weiiwang01 (dismissed):

src/runner.py
```diff
@@ -640,9 +622,9 @@ def _configure_docker_proxy(self):
         docker_client_proxy = {
             "proxies": {
                 "default": {
-                    "httpProxy": self.config.proxies["http"],
-                    "httpsProxy": self.config.proxies["https"],
-                    "noProxy": self.config.proxies["no_proxy"],
+                    "httpProxy": self.config.proxies.http or "",
+                    "httpsProxy": self.config.proxies.https or "",
+                    "noProxy": self.config.proxies.no_proxy or "",
```
jdkandersson, line 627: It would probably better not to set the environment variables rather than setting them to empty strings if the proxy information is not defined. You can do this concisely using something like `{key: value for key, value in (("httpProxy", self.config.proxies.http), ...) if value}`

src/runner_manager.py
```diff
@@ -388,6 +378,50 @@ def _issue_reconciliation_metric(
...
+        Args:
+            name: Name of the runner.
+
+        Returns:
+            Configuration for the runner.
+        """
+        if self.proxies and not self.proxies.use_aproxy:
+            # If the proxy setting are set, then add NO_PROXY local variables.
+            if self.proxies.no_proxy:
+                no_proxy = self.proxies.no_proxy + ","
```
jdkandersson, line 396: What about using an f-string?

jdkandersson (dismissed):

weiiwang01 (dismissed):

weiiwang01 (approved):

jdkandersson (approved):

PR #225, opened 2024-02-15
Initial OpenStack integration

cbartz:
Applicable spec: ISD-100 (private)

### Overview

Add support for using OpenStack virtual machines instead of LXD virtual machines. This PR only introduces the initial integration by providing an experimental option to set a `clouds.yaml` file whose credentials are used to verify access to the cloud. There will be future PRs for the rest of the functionality (spawning runners, building images).

An integration test using microstack is also added. 

Initial code has been provided by @yanksyoon .

### Rationale

To support OpenStack integration, we need to start by providing a configuration option and a first basic integration. After this work, the remaining features can be developed in parallel.

### Juju Events Changes

`on_config_changed` now tests the OpenStack integration by calling a method (which is likely to be removed in the future).

### Module Changes

- `charm_state`: validates the yaml
- `openstack_manager`: further validates the yaml semantically, writes config to disk, provides a function to interact with the cloud

### Checklist

Documentation on charmhub is not updated as the openstack feature is not yet ready.


scripts/setup-microstack.sh
```diff
@@ -0,0 +1,54 @@
...
+unset HTTP_PROXY HTTPS_PROXY NO_PROXY http_proxy https_proxy no_proxy
+# microk8s charm installed by microstack tries to create an alias for kubectl and fails otherwise
+sudo snap remove kubectl
+# Install microstack
+sudo snap install openstack --channel 2023.1 --devmode
+sunbeam prepare-node-script | bash -x
+sleep 10
+# The following can take a while....
+retry 'sudo -g snap_daemon sunbeam cluster bootstrap --accept-defaults' 'Waiting for cluster bootstrap to complete' 3
+sg snap_daemon -c "sunbeam openrc" > admin-openrc
```
yanksyoon, line 45: Out of curiosity, is this because we're passed on `openrc` from the IS?
I wanted to suggest
```
$ sunbeam cloud-config -h
Usage: sunbeam cloud-config [OPTIONS]

  Generate or Update clouds.yaml.

Options:
  -c, --cloud TEXT       Name of the cloud
  -a, --admin            Generate cloud-config for cloud admin user. If not
                         specified cloud-config is generated for the
                         demonstration user setup during configure.
  -u, --update           Create/update config in the file specified in cloud-
                         file option
  -f, --cloud-file FILE  Output file for cloud yaml, defaults to
                         $HOME/.config/openstack/clouds.yaml
  -h, --help             Show this message and exit.

$ sunbeam cloud-config -a
clouds:
  sunbeam:
    auth:
      auth_url: http://10.20.21.12/openstack-keystone
      password: M0Pzc2D6KYlb
      project_domain_name: admin_domain
      project_name: admin
      user_domain_name: admin_domain
      username: admin
```
which might be easier to use for testing input, since we're using openstack `clouds.yaml` - we don't need to format anything in this case :)

yanksyoon (dismissed):
LGTM, minor question!

src/charm_state.py
```diff
@@ -85,7 +91,26 @@ def from_charm(cls, charm: CharmBase) -> "CharmConfig":
...
+        if openstack_clouds_yaml_str:
+            try:
+                openstack_clouds_yaml = yaml.safe_load(openstack_clouds_yaml_str)
+            except yaml.YAMLError as exc:
+                logger.error("Invalid openstack-clouds-yaml config: %s.", exc)
+                raise CharmConfigInvalidError(
+                    "Invalid openstack-clouds-yaml config. Invalid yaml."
+                ) from exc
+            try:
+                openstack_manager.initialize(openstack_clouds_yaml)
```
jdkandersson, line 104: The type hint for `initialise` is that `openstack_clouds_yaml` is a `dict` which is not guaranteed at this point - it could be a `list`, `str` or `int` etc. Perhaps checking it is a `dict` is worthwhile here as well? Also potentially means that additional tests could be written here

src/openstack_manager.py
```diff
@@ -0,0 +1,110 @@
...
+
+    Args:
+        cloud_config: The configuration in clouds.yaml format to validate.
+
+    Raises:
+        InvalidConfigError: if the format of the config is invalid.
+    """
+    # dict of format: {clouds: <cloud-name>: <cloud-config>}
+    try:
+        clouds = list(cloud_config["clouds"].keys())
```
jdkandersson, line 32: All sorts of things could be in `clouds`, it doesn't necessarily have to be iterable which is what i think is required by the `list` constructor. Given that we know that it is going to come from a `yaml`, perhaps check it is a `list`? Also indicates that we should do additional testing here, e.g., passing invalid cloud config yaml in all ways that we can think of

src/openstack_manager.py
```diff
@@ -0,0 +1,110 @@
...
+
+
+def _write_config_to_disk(cloud_config: dict) -> None:
+    """Write the cloud configuration to disk.
+
+    Args:
+        cloud_config: The configuration in clouds.yaml format to write to disk.
+    """
+    CLOUDS_YAML_PATH.parent.mkdir(parents=True, exist_ok=True)
+    with CLOUDS_YAML_PATH.open("w", encoding="utf-8") as clouds_yaml:
```
jdkandersson, line 47: What about `Path.write_text` https://docs.python.org/3/library/pathlib.html#pathlib.Path.write_text?

src/openstack_manager.py
```diff
@@ -0,0 +1,110 @@
...
+
+    Args:
+        cloud_config: The configuration in clouds.yaml format to apply.
+    Raises:
+        InvalidConfigError: if the config has not all required information.
+
+    Returns:
+        An openstack.connection.Connection object.
+    """
+    clouds = list(cloud_config["clouds"].keys())
```
jdkandersson, line 62: I'm guessing at this point we assume that the configuration is valid? Perhaps worth mentioning in the docstring

src/openstack_manager.py
```diff
@@ -0,0 +1,110 @@
...
+    if len(clouds) > 1:
+        logger.warning("Multiple clouds defined in clouds.yaml. Using the first one to connect.")
+    cloud_name = clouds[0]
+
+    # api documents that keystoneauth1.exceptions.MissingRequiredOptions can be raised but
+    # I could not reproduce it. Therefore, no catch here.
```
jdkandersson, line 68: Perhaps worth checking with the team? If the documentation says it can be raised we should probably handle it

src/openstack_manager.py
```diff
@@ -0,0 +1,110 @@
...
+
+
+def list_projects(cloud_config: dict) -> list[Project]:
+    """List all projects in the OpenStack cloud.
+
+    The purpose of the method is just to try out openstack integration and
+    it may be removed in the future.
+
+    It currently returns objects directly from the sdk,
+    which may not be ideal (mapping to domain objects may be preferable).
```
jdkandersson, line 94: Yeh this shouldn't return objects from the SDK since that breaks encapsulation. Since we don't really need the project, perhaps just return the number of projects or a list of project names?

src/openstack_manager.py
```diff
@@ -0,0 +1,110 @@
...
+
+    Returns:
+        A list of projects.
+    """
+    conn = _create_connection(cloud_config)
+    try:
+        projects = conn.list_projects()
+        logger.debug("OpenStack connection successful.")
+        logger.debug("Projects: %s", projects)
+        # pylint thinks this isn't an exception
```
jdkandersson, line 104: Perhaps worth sharing with the OpenStack team for them to investigate

tests/unit/test_charm_state.py
```diff
@@ -229,15 +252,55 @@ def test_from_charm_ssh_debug_info():
...
+    act: Retrieve state from charm.
+    assert: CharmConfigInvalidError is raised.
+    """
+    charm.config[charm_state.OPENSTACK_CLOUDS_YAML_CONFIG_NAME] = json.dumps(clouds_yaml)
+    charm_state.openstack_manager.initialize.side_effect = OpenStackInvalidConfigError("invalid")
+
+    with pytest.raises(CharmConfigInvalidError) as exc:
+        State.from_charm(charm)
+    assert "Invalid openstack config. Not able to initialize openstack integration." in str(
+        exc.value
```
jdkandersson, line 305: We probably will need more here as noted earlier

tests/unit/test_openstack_manager.py
```diff
@@ -0,0 +1,149 @@
...
+
+    openstack_manager.initialize(clouds_yaml)
+
+    assert yaml.safe_load(clouds_yaml_path.read_text(encoding="utf-8")) == clouds_yaml
+
+
+@pytest.mark.parametrize(
+    "invalid_yaml, expected_err_msg",
+    [
+        pytest.param({"wrong-key": _create_clouds_yaml()["clouds"]}, INVALID_CLOUDS_YAML_ERR_MSG),
```
jdkandersson, line 83: We'll probably need more here as noted earlier

jdkandersson (commented):
Thanks, I think we just need to do a few more checks on the input

src/openstack_manager.py
yanksyoon: I think this can be changed from a the follow-up PRs since we'll have a more solid use-cases on what we need.

src/openstack_manager.py
yanksyoon: Just tried reproducing with invalid `clouds.yaml` (commented out fields, deleted fields) - they don't raise this exception. 
```
clouds:
  sunbeam:
    auth:
      auth_url: http://10.20.21.12/openstack-keystone
#      password: 8gfWnW157KJ7
#      project_domain_name: admin_domain
#      project_name: admin
#      user_domain_name: admin_domain
#      username: admin
```
Raised an [issue on the bugtracker](https://bugs.launchpad.net/openstacksdk/+bug/2054361).

src/openstack_manager.py
yanksyoon: We're checking for the `keys()` of the dict, so I think as long as the validation of `dict` for cloud_config is done w/ reference to the comment above, it should be okay :D thank you for the review!

src/openstack_manager.py
yanksyoon: Filed a [bug on launchpad](https://bugs.launchpad.net/openstacksdk/+bug/2054364)

tests/unit/test_charm_state.py
yanksyoon: Added validation on the `openstack_manager.py` module side of things, I think mocking the behavior here is okay for now. Would type checking be preferred on the `charm_state` side?

src/openstack_manager.py
yanksyoon: Fixed, thank you!

tests/unit/test_charm_state.py
yanksyoon: On a second thought, it makes more sense for the module to take in dict as a parameter than Any. moving this preliminary check to `charm_state.py`

src/charm_state.py
yanksyoon: Added tests to check parsed type, thank you!

src/openstack_manager.py
yanksyoon: Added description to docstring

jdkandersson (approved):

yanksyoon (approved):

PR #239, opened 2024-03-05
feat: build openstack image

yanksyoon:
Applicable spec: None

### Overview

This PR adds building Openstack image feature, using ubuntu server image.

Out of scope:
- openstack network/ssh
- openstack server create added to test the image.

The current self-hosted runners have 40GiB disk, which isn't enough to run the integration test with [microstack](https://microstack.run/docs/single-node).
Due to such reasons, I've run the integration test on a local machine with enough disk space. Here is the link to [E2E test run](https://github.com/canonical/github-runner-operator/actions/runs/8233166340/job/22512076101). All the runs fail except:
- firewall test
- docker daemon/docker client proxy (Proxy configuration on juju environment TBD)

### Rationale

- This change is a step to support launching Github runners on Openstack VMs. This helps: 
1. support additional cloud type to launch Github runners
2. Reduces a layer of nested virtualization - which is a step to support ARM64 runners

### Juju Events Changes

- All juju event handlers lead to BlockedStatus on Openstack charm config

### Module Changes

- `scripts/setup-microstack.sh`: This now sets up microstack with demo-user, which sets up other components such as network, user, flavor, ...etc.
- `scripts/build-openstack-image.sh`: Builds openstack image on charm.
- `src/charm.py`: All event handlers now go to blockestatus when openstack config is set.
- `src/charm_state.py`: Moved openstack initialization functions from `openstack_manager.py` to remove circular dependency.
- `src/github_client.py`: Refactored getting github runner application.
- `templates/openstack-userdata.sh.j2`: Cloud-init's userdata, which runs when the instance is launched. Responsible for running/registering Github runner on startup.

### Library Changes

None.

### Checklist

- [ ] The [contributing guide](https://github.com/canonical/is-charms-contributing-guide) was applied

The contributing guide is not respected in the case of integration tests because the testing infrastructure cannot cater for the disk requirements.


scripts/setup-microstack.sh
```diff
@@ -39,10 +39,12 @@ sunbeam prepare-node-script | bash -x
 sleep 10
 # The following can take a while....
 retry 'sudo -g snap_daemon sunbeam cluster bootstrap --accept-defaults' 'Waiting for cluster bootstrap to complete' 3
-clouds_yaml="${PWD}/admin-clouds.yaml"
-sg snap_daemon -c "sunbeam cloud-config -a" | tee "$clouds_yaml"
+# 2024/03/11 Demo user setup should be removed after openstack server creation PR.
+retry 'sudo -g snap_daemon sunbeam configure --accept-defaults --openrc demo-openrc' 'Configuring sunbeam cluster' 3
+clouds_yaml="${PWD}/clouds.yaml"
+sg snap_daemon -c "sunbeam cloud-config" | tee "$clouds_yaml"
```
yanksyoon, line 45: Use a pre-configured microstack to reduce the scope of PR. Otherwise,
network creation, flavour creation ...etc would be required.

src/charm.py
```diff
@@ -344,6 +345,41 @@ def _on_install(self, _event: InstallEvent) -> None:
         """
         state = self._setup_state()
 
+        if state.charm_config.openstack_clouds_yaml:
```
yanksyoon, line 348: The changes on charm.py blocks on all hooks on openstack configuration - other hooks have dependencies, e.g. on_config_changed depends on on_install hook to run properly.

src/charm_state.py
```diff
@@ -165,6 +162,54 @@ def _valid_storage_size_str(size: str) -> bool:
     return size[-3:] in valid_suffixes and size[:-3].isdigit()
 
 
+OPENSTACK_CLOUDS_YAML_CONFIG_NAME = "experimental-openstack-clouds-yaml"
+CLOUDS_YAML_PATH = Path(Path.home() / ".config/openstack/clouds.yaml")
+
+
+def _validate_openstack_cloud_config(cloud_config: dict) -> None:
```
yanksyoon, line 169: These functions have been moved from `openstack_manager.py` to remove circular dependency (not just in typing)

src/openstack_manager.py
```diff
@@ -113,3 +85,196 @@ def list_projects(cloud_config: dict) -> list[Project]:
...
+        registration_token: Token for registering the runner on GitHub.
+        github_path: The GitHub repo/org path
+        openstack_image: The Openstack image to use to boot the instance with.
+    """
+
+    name: str
+    labels: Iterable[str]
+    registration_token: str
+    github_path: GithubPath
+    openstack_image: openstack.image.v2.image.Image
```
yanksyoon, line 155: The interfaces are subject to change during refactor - which is out of the scope of this PR.

jdkandersson:
As discussed, let's change the disk size of `2xlarge` so that we can run the test 🙂

scripts/build-openstack-image.sh
```diff
@@ -0,0 +1,177 @@
...
+cleanup() {
+    sudo umount /mnt/ubuntu-image/dev/ || true
+    sudo umount /mnt/ubuntu-image/proc/ || true
+    sudo umount /mnt/ubuntu-image/sys/ || true
+    sudo umount /mnt/ubuntu-image || true
+    sudo qemu-nbd --disconnect /dev/nbd0
+}
+
+if [[ -n "$HTTP_PROXY" ]]; then
+    /usr/bin/echo "HTTP_PROXY=$HTTP_PROXY" >> /etc/environment
```
jdkandersson, line 48: Wouldn't we endlessly keep adding this on every script run?

src/charm.py
```diff
@@ -344,6 +348,41 @@ def _on_install(self, _event: InstallEvent) -> None:
...
+                    cloud_config=state.charm_config.openstack_clouds_yaml,
+                    instance_config=instance_config,
+                    proxies=state.proxy_config,
+                    dockerhub_mirror=state.charm_config.dockerhub_mirror,
+                    ssh_debug_connections=state.ssh_debug_connections,
+                )
+                logger.info("OpenStack instance: %s", instance)
+            # Test out openstack integration and then go
+            # into BlockedStatus as it is not supported yet
+            self.unit.status = BlockedStatus(
```
jdkandersson, line 380: Could this call the shared `_block_on_openstack_config` function?

src/charm.py
```diff
@@ -384,6 +423,14 @@ def _on_install(self, _event: InstallEvent) -> None:
 
         self.unit.status = ActiveStatus()
 
+    def _block_on_openstack_config(self) -> None:
```
jdkandersson, line 426: Perhaps this can also do the openstack cloud config check and return a boolean for the caller to determine whether they should exit the hook?

src/charm_state.py
jdkandersson: We should look at this in a bit more detail, it doesn't belong here. It could go into a third module around openstack initialisation or something like that

src/charm_state.py
```diff
@@ -165,6 +162,54 @@ def _valid_storage_size_str(size: str) -> bool:
...
+        cloud_config: The configuration in clouds.yaml format to validate.
+
+    Raises:
+        InvalidConfigError: if the format of the config is invalid.
+    """
+    # dict of format: {clouds: <cloud-name>: <cloud-config>}
+    try:
+        clouds = list(cloud_config["clouds"].keys())
+    except KeyError as exc:
+        raise OpenStackInvalidConfigError("Invalid clouds.yaml.") from exc
```
jdkandersson, line 182: Perhaps it could specifically say that the `clouds` key is missing?

src/github_client.py
```diff
@@ -80,6 +81,38 @@ def get_runner_applications(self, path: GithubPath) -> RunnerApplicationList:
...
+            os: The operating system that the runner binary should run on.
+
+        Raises:
+            RunnerBinaryError: If the runner application for given architecture and OS is not
+                found.
+
+        Returns:
+            The runner application.
+        """
+        runner_applications = self.get_runner_applications(path=path)
```
jdkandersson, line 103: Could this function be private/ removed and moved into this function now?

src/openstack_manager.py
```diff
@@ -2,58 +2,45 @@
...
 
-    Args:
-        cloud_config: The configuration in clouds.yaml format to write to disk.
-    """
-    CLOUDS_YAML_PATH.parent.mkdir(parents=True, exist_ok=True)
-    CLOUDS_YAML_PATH.write_text(encoding="utf-8", data=yaml.dump(cloud_config))
 
-
-def _create_connection(cloud_config: dict) -> openstack.connection.Connection:
+def _create_connection(cloud_config: dict[str, dict]) -> openstack.connection.Connection:
```
jdkandersson, line 38: Perhaps this can be a typed dict with the `clouds` key and the value of that just an arbitrary dictionary?

src/openstack_manager.py
```diff
@@ -113,3 +85,196 @@ def list_projects(cloud_config: dict) -> list[Project]:
...
+        Command to execute to build runner image.
+    """
+    http_proxy = proxies.http if (proxies and proxies.http) else ""
+    https_proxy = proxies.https if (proxies and proxies.https) else ""
+    no_proxy = proxies.no_proxy if (proxies and proxies.no_proxy) else ""
+
+    environment = jinja2.Environment(loader=jinja2.FileSystemLoader("templates"), autoescape=True)
+    docker_proxy_service_conf_content = environment.get_template("systemd-docker-proxy.j2").render(
+        proxies=proxies
+    )
```
jdkandersson, line 109: Consider making these separate functions for unit tests rather than having to access index x of the command which might change

src/openstack_manager.py
```diff
@@ -113,3 +85,196 @@ def list_projects(cloud_config: dict) -> list[Project]:
...
+        raise ImageBuildError("Failed to fetch runner application.") from exc
+
+    try:
+        execute_command(_build_image_command(runner_application, proxies), check_exit=True)
+    except SubprocessError as exc:
+        raise ImageBuildError("Failed to build image.") from exc
+
+    try:
+        conn = _create_connection(cloud_config)
+        arch = "amd64" if runner_application["architecture"] == "x64" else "arm64"
```
jdkandersson, line 191: Perhaps have a check here if the architecture is unexpected?

src/openstack_manager.py
```diff
@@ -113,3 +85,196 @@ def list_projects(cloud_config: dict) -> list[Project]:
...
+        existing_image: openstack.image.v2.image.Image
+        for existing_image in conn.search_images(name_or_id=IMAGE_NAME):
+            # images with same name (different ID) can be created and will error during server
+            # instantiation.
+            if not conn.delete_image(name_or_id=existing_image.id, wait=True):
+                raise ImageBuildError("Failed to delete duplicate image on Openstack.")
+        return conn.create_image(
```
jdkandersson, line 198: We probably shouldn't return a type from the open stack package in a public function

src/openstack_manager.py
```diff
@@ -113,3 +85,196 @@ def list_projects(cloud_config: dict) -> list[Project]:
...
+
+
+@retry(tries=5, delay=5, max_delay=60, backoff=2, local_logger=logger)
+def create_instance(
+    cloud_config: dict[str, dict],
+    instance_config: InstanceConfig,
+    proxies: Optional[ProxyConfig] = None,
+    dockerhub_mirror: Optional[str] = None,
+    ssh_debug_connections: list[SSHDebugConnection] | None = None,
+) -> openstack.compute.v2.server.Server:
```
jdkandersson, line 242: Perhaps change this return type too

src/openstack_manager.py
```diff
@@ -113,3 +85,196 @@ def list_projects(cloud_config: dict) -> list[Project]:
...
+        instance_config: The configuration values for Openstack instance to launch.
+
+    Raises:
+        InstanceLaunchError: if any errors occurred while launching Openstack instance.
+
+    Returns:
+        The created server.
+    """
+    environment = jinja2.Environment(loader=jinja2.FileSystemLoader("templates"), autoescape=True)
+    env_contents = environment.get_template("env.j2").render(
```
jdkandersson, line 256: Potentially separate into another function for unit testing

jdkandersson (commented):
Thanks, some minor comments

src/openstack_manager.py
```diff
@@ -2,58 +2,45 @@
...
 
 """Module for handling interactions with OpenStack."""
+import json
 import logging
-from pathlib import Path
+import secrets
+from dataclasses import dataclass
 
+# subprocess module is used to call image build script.
+from subprocess import SubprocessError  # nosec
```
cbartz, line 11: I know it may be confusing, but its another SubprocessError that you need to catch, its from `errors.SubprocessError`.

src/openstack_manager.py
```diff
@@ -113,3 +85,196 @@ def list_projects(cloud_config: dict) -> list[Project]:
         ) from exc
 
     return projects
+
+
+class ImageBuildError(Exception):
+    """Exception representing an error during image build process."""
```
cbartz, line 91: Sometimes errors are in `errors`, and sometimes errors are in their respective modules. I think we should agree on an approach here.

tests/unit/test_openstack_manager.py
```diff
@@ -46,104 +24,243 @@ def mock_openstack(monkeypatch: pytest.MonkeyPatch, projects) -> MagicMock:
...
+            arch=MagicMock(),
+            cloud_config=MagicMock(),
+            github_client=mock_github_client,
+            path=MagicMock(),
+        )
+
+    assert "Failed to fetch runner application." in str(exc)
+
+
+def test_build_build_image_script_error(monkeypatch: pytest.MonkeyPatch):
```
cbartz, line 155:
```suggestion
def test_build_image_script_error(monkeypatch: pytest.MonkeyPatch):
```

tests/unit/test_openstack_manager.py
```diff
@@ -46,104 +24,243 @@ def mock_openstack(monkeypatch: pytest.MonkeyPatch, projects) -> MagicMock:
...
+        )
+
+    assert "Failed to build image." in str(exc)
+
+
+def test_build_image_delete_image_error(monkeypatch: pytest.MonkeyPatch):
+    """
+    arrange: given a mocked openstack connection that returns existing images and delete_image
+        that returns False (failed to delete image).
+    act: when bulid_image is called.
```
cbartz, line 182:
```suggestion
    act: when build_image is called.
```

tests/unit/test_openstack_manager.py
```diff
@@ -46,104 +24,243 @@ def mock_openstack(monkeypatch: pytest.MonkeyPatch, projects) -> MagicMock:
...
+            path=MagicMock(),
+        )
+
+    assert "Failed to delete duplicate image on Openstack." in str(exc)
+
+
+def test_build_image_create_image_error(monkeypatch: pytest.MonkeyPatch):
+    """
+    arrange: given a mocked connection that raises OpenStackCloudException on create_image.
+    act: when bulid_image is called.
```
cbartz, line 215:
```suggestion
    act: when build_image is called.
```

tests/integration/helpers.py
```diff
@@ -320,6 +324,8 @@ async def deploy_github_runner_charm(
     https_proxy: str,
     no_proxy: str,
     reconcile_interval: int,
+    constraints: dict | None = None,
+    config: dict | None = None,
```
cbartz, line 328: These are missing in the docstring

tests/integration/test_openstack.py
```diff
@@ -4,37 +4,52 @@
 """Integration tests for OpenStack integration."""
 
 
-import yaml
+import openstack
```
cbartz, line 7:
```suggestion
```
unused

cbartz (commented):
Looks nice, have a few comments. I also think that openstack initialisation code should not go into charm state.

tests/integration/test_openstack.py
yanksyoon: Thanks for catching this!

tests/integration/helpers.py
yanksyoon: Added, thanks!

tests/unit/test_openstack_manager.py
yanksyoon: Oof, fixed all the typos, thanks!

src/openstack_manager.py
yanksyoon: You're right, thanks for catching this!

src/openstack_manager.py
yanksyoon: You're right, thanks!

charmcraft.yaml
```diff
@@ -14,6 +14,7 @@ parts:
       - pkg-config # for cryptography
     prime:
       - scripts/build-image.sh
```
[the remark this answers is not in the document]
yanksyoon: Updated, thanks!

scripts/build-openstack-image.sh
```diff
@@ -0,0 +1,177 @@
+#!/usr/bin/env bash
+
+# Copyright 2024 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+set -e
```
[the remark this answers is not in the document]
yanksyoon: Updated, thanks!

src/openstack_manager.py
yanksyoon: Refactored, thank you!

src/openstack_manager.py
yanksyoon: Fixed the return types to a more generic one, thank you!

src/openstack_manager.py
yanksyoon: Added a check w/ unit tests, thank you!

src/openstack_manager.py
yanksyoon: Added CloudConfig type, thank you!

src/github_client.py
yanksyoon: Refactored, thank you!

src/charm_state.py
yanksyoon: That is indeed much more helpful. Thank you!

src/charm.py
yanksyoon: Updated! thank you!

src/charm.py
yanksyoon: Updated, thank you!

scripts/build-openstack-image.sh
yanksyoon: Updated to conditionally update when
1. the value doesn't exist
2. the value is different from what is supposed to be set

This should be refactored into a Pythonic script later on.

Thank you for the review!

scripts/build-openstack-image.sh
```diff
@@ -0,0 +1,177 @@
...
+if [[ -n "$DOCKER_PROXY_CONF" ]]; then
+    mkdir -p /root/.docker
+    echo "$DOCKER_PROXY_CONF" > /root/.docker/config.json
+    mkdir -p /home/ubuntu/.docker
+    echo "$DOCKER_PROXY_CONF" > /home/ubuntu/.docker/config.json
+fi
+
+# Reduce image size
+/usr/bin/npm cache clean --force
+DEBIAN_FRONTEND=noninteractive /usr/bin/apt-get clean
```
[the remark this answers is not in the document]
jdkandersson: yeh I think Python would help here especially to be able to re-use and to handle retries etc

jdkandersson (commented):
very close now

src/charm_state.py
```diff
@@ -162,54 +164,6 @@ def _valid_storage_size_str(size: str) -> bool:
     return size[-3:] in valid_suffixes and size[:-3].isdigit()
 
 
-OPENSTACK_CLOUDS_YAML_CONFIG_NAME = "experimental-openstack-clouds-yaml"
```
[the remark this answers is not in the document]
yanksyoon: I've replaced it with a string from a constant for consistency, perhaps all other config values can also be defined as constant as well then. Thank you for the review, updating!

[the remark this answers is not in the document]
yanksyoon: To reduce the complexity for this PR, would it be fine if I push this change with consistency fixes, flake8 and documentation fixes on another follow up PR please? (`chore/flake8_consistency` branch)

jdkandersson (dismissed):

src/charm.py
```diff
@@ -375,12 +375,7 @@ def _on_install(self, _event: InstallEvent) -> None:
                     ssh_debug_connections=state.ssh_debug_connections,
                 )
                 logger.info("OpenStack instance: %s", instance)
-            # Test out openstack integration and then go
-            # into BlockedStatus as it is not supported yet
-            self.unit.status = BlockedStatus(
-                "OpenStack integration is not supported yet. "
-                "Please remove the openstack-clouds-yaml config."
-            )
+            self._block_on_openstack_config(state)
```
cbartz, line 378: The order of the fct is inconsistent with our contribution guide https://github.com/canonical/is-charms-contributing-guide/blob/main/CONTRIBUTING.md#function-and-method-Ordering (_block_on_openstack_config should be listed before this fct), but it doesn't seem to be followed in a lot of other PRs in our repos, so perhaps this rule should be revised in the contribution guide, or stated as a recommendation, otherwise I'll have to mention this nitpick in a lot of PRs. cc @jdkandersson Feel free to resolve this conversation.

tests/unit/test_openstack_manager.py
```diff
@@ -87,6 +113,203 @@ def test_list_projects_missing_credentials_error(
...
+            templates_env=environment,
+            proxies=proxy_config,
+            dockerhub_mirror=dockerhub_mirror,
+            ssh_debug_connections=ssh_debug_connections,
+        )
+        == expected_env_contents
+    )
+
+
 def test__build_image_command():
```
cbartz, line 313:
```suggestion
def test_build_image_command():
```

cbartz (dismissed):

src/charm_state.py
yanksyoon: This has now been refactored into a separate module. Thank you!

tests/unit/test_openstack_manager.py
yanksyoon: This is a test for an internal function `_build_image_command`, and there is a fct `build_image` so i think this is alright :)

src/charm.py
yanksyoon: Updated, thank you!

cbartz (dismissed):

jdkandersson (dismissed):

jdkandersson (dismissed):

jdkandersson:
As discussed, the tests that are expected to fail, let's mark them as xfail in pytest

jdkandersson (dismissed):

jdkandersson (dismissed):

src/openstack_cloud/openstack_manager.py
```diff
@@ -0,0 +1,378 @@
...
+
+IMAGE_PATH_TMPL = "jammy-server-cloudimg-{architecture}-compressed.img"
+IMAGE_NAME = "jammy"
+BUILD_OPENSTACK_IMAGE_SCRIPT_FILENAME = "scripts/build-openstack-image.sh"
+
+
+def _create_connection(cloud_config: dict[str, dict]) -> openstack.connection.Connection:
+    """Create a connection object.
+
+    This method should be called with a valid cloud_config. See def _validate_cloud_config.
```
yhaliaw, line 43:
```suggestion
    This method should be called with a valid cloud_config. See _validate_cloud_config.
```
Minor suggestion

src/openstack_cloud/openstack_manager.py
```diff
@@ -0,0 +1,378 @@
...
+
+    Returns:
+        A list of projects.
+    """
+    conn = _create_connection(cloud_config)
+    try:
+        projects = conn.list_projects()
+        logger.debug("OpenStack connection successful.")
+        logger.debug("Projects: %s", projects)
+        # pylint thinks this isn't an exception
```
yhaliaw, line 83: It might be good to point out which exception is causing the problem for pylint. For a reader without context, it might be `OpenStackUnauthorizedError` or `keystoneauth1.exceptions.http.Unauthorized`. I feel explicitly mention the exception would be a good idea.

I encountered this issue as well. It seems pylint does not think `keystoneauth1.exceptions.http.Unauthorized` is an exception. However, from the source code it seems to be a child class of `Exception`. I am unable to figure why pylint is throwing this error.

I can also include this change in the follow-up PR as well. Just want to confirm if your find points toward `keystoneauth1.exceptions.http.Unauthorized` being the problem as well.

src/openstack_cloud/openstack_manager.py
```diff
@@ -0,0 +1,378 @@
...
+        proxies: HTTP proxy settings.
+
+    Returns:
+        Command to execute to build runner image.
+    """
+    docker_proxy_service_conf_content = _generate_docker_proxy_unit_file(proxies=proxies)
+
+    http_proxy = str(proxies.http) if (proxies and proxies.http) else ""
+    https_proxy = str(proxies.https) if (proxies and proxies.https) else ""
+    no_proxy = str(proxies.no_proxy) if (proxies and proxies.no_proxy) else ""
```
yhaliaw, line 149: Would it be good to extract the None check for `proxies` out? It does increase the nesting by one, but it would avoid repeated None check for line 147 - 149.

src/openstack_cloud/openstack_manager.py
```diff
@@ -0,0 +1,378 @@
...
+    """
+
+    name: str
+    labels: Iterable[str]
+    registration_token: str
+    github_path: GithubPath
+    openstack_image: openstack.image.v2.image.Image
+
+
+def _get_supported_runner_arch(arch: str) -> Literal["amd64", "arm64"]:
```
yhaliaw, line 188: Could the type `Arch` defined in `charm_state.py` be used instead of `Literal["amd64", "arm64"]`?

src/openstack_cloud/openstack_manager.py
```diff
@@ -0,0 +1,378 @@
...
+from utilities import execute_command, retry
+
+logger = logging.getLogger(__name__)
+
+IMAGE_PATH_TMPL = "jammy-server-cloudimg-{architecture}-compressed.img"
+IMAGE_NAME = "jammy"
+BUILD_OPENSTACK_IMAGE_SCRIPT_FILENAME = "scripts/build-openstack-image.sh"
+
+
+def _create_connection(cloud_config: dict[str, dict]) -> openstack.connection.Connection:
```
yhaliaw, line 40: We might need to manage the openstack connection differently.

The `openstack.connection.Connection` has defined the `__enter__`, and `__exit__` method, therefore it is likely designed to be used a context manager. 
The `__exit__` includes a call to `close` which does some cleanup: https://github.com/openstack/openstacksdk/blob/601b21acadf30f54e158523b48af47d3cac23fa5/openstack/connection.py#L654
We should be ensure `close`\`__exit__` is called at the end of each connection.

Also we would need to convert all `keystoneauth1.exceptions.http.Unauthorized` to `OpenStackUnauthorizedError` so the `charm.py` can just deal with `OpenStackUnauthorizedError` and not leak implementation detail to `charm.py`.

I was planning to do something like this:
```
from contextlib import contextmanager

@contextmanager
def _create_connection(self) -> Iterator[OpenstackConnection]:
    try:
        with OpenstackConnection(self.cloud_name) as conn:
            yield conn
    except keystoneauth1.exceptions.http.Unauthorized as exc:
        logger.exception("Unable to connect to openstack")
        raise OpenStackUnauthorizedError() from exc
```

Which would be used like:
```
with self._create_connection as conn:
    # conn is a openstack.connection.Connection instance.
    # Use conn to call openstack API.
```

src/openstack_cloud/openstack_manager.py
```diff
@@ -0,0 +1,378 @@
...
+    return InstanceConfig(
+        name=f"{app_name}-{unit_num}-{suffix}",
+        labels=(app_name, "jammy"),
+        registration_token=registration_token,
+        github_path=path,
+        openstack_image=openstack_image,
+    )
+
+
+class InstanceLaunchError(Exception):
```
yhaliaw, line 288: I think "public" exceptions are generally defined in the `errors.py` file.
By "public", I am referring to exception that would be import and caught in another module, e.g., `charm.py`.

It seems this exception would be caught by another module?

If this is moved to `errors.py`, it might be good to rename with Openstack in the name.

templates/openstack-userdata.sh.j2
```diff
@@ -0,0 +1,19 @@
...
+su - ubuntu -c "cd ~/actions-runner && ./config.sh \
+    --url {{ github_url }} \
+    --token {{ token }} --ephemeral --unattended \
+    --labels {{ instance_labels }} --name {{ instance_name }}"
+
+# Run runner
+su - ubuntu -c "cd ~/actions-runner && /home/ubuntu/actions-runner/run.sh"
+
+# cleanup
+sleep 2
```
yhaliaw, line 18: Line 17 - 19 does not seems to be cleanup.
Is there a reason for 2 seconds? Are we waiting for something?

yhaliaw (commented):
Reviewed up to tests.

src/openstack_cloud/openstack_manager.py
yanksyoon: This is a conversion to ubuntu cloud_images supported arch which is different - rather than "x64" it uses the string "amd64". Adding a comment to clarify that!

src/openstack_cloud/openstack_manager.py
yanksyoon: Added a comment, thank you! I've also added a connection.authorize() as well - could you have a look?

src/openstack_cloud/openstack_manager.py
yanksyoon: Refactored, thank you!

src/openstack_cloud/openstack_manager.py
yanksyoon: Refactored, thank you!
I don't think instantiating `OpenstackConnection` would raise - hence I've added `conn.authorize()` before yielding the connection so that we can catch the `openstack.exceptions.HttpException` (notice that it isn't keystoneauth exception) before we let the users use the connection.

src/openstack_cloud/openstack_manager.py
yanksyoon: Moved, thank you!

templates/openstack-userdata.sh.j2
yanksyoon: Removed, thank you!
(This was a placeholder for post-runner actions)

yhaliaw (approved):

jdkandersson (approved):

PR #248, opened 2024-03-21
chore: flake8 fixes

yanksyoon:
Applicable spec: N/A

### Overview

- Enable flake8 settings to match other repository standards
- Fix issues detected by enabling it

### Rationale

- To raise the code standards for this repository.

### Juju Events Changes

None.

### Module Changes

Mostly all modules, fixed stylistic issues.

### Library Changes

None.

### Checklist


pyproject.toml
```diff
@@ -4,30 +4,34 @@
...
+fail_under = 83
 show_missing = true
 
-
 [tool.pytest.ini_options]
 minversion = "6.0"
 log_cli_level = "INFO"
 
+[tool.pylint.'MESSAGES CONTROL']
+extension-pkg-whitelist = "pydantic"
```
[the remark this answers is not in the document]
yanksyoon: Thanks reviewdawg!

src/charm.py
```diff
@@ -90,7 +96,13 @@ def catch_charm_errors(func: Callable[[CharmT, EventT], None]) -> Callable[[Char
     """
 
     @functools.wraps(func)
-    def func_with_catch_errors(self, event: EventT) -> None:
+    # flake8 thinks the event argument description is missing in the docstring.
+    def func_with_catch_errors(self: CharmT, event: EventT) -> None:
```
jdkandersson, line 100: Is a type definition on self required?

src/charm.py
```diff
@@ -90,7 +96,13 @@ def catch_charm_errors(func: Callable[[CharmT, EventT], None]) -> Callable[[Char
...
+        """Handle errors raised while handling charm events.
+
+        Args:
+            event: The charm event to handle.
+        """  # noqa: D417
```
jdkandersson, line 105: Perhaps double check this?

src/github_type.py
```diff
@@ -33,15 +38,16 @@ class RunnerApplication(TypedDict, total=False):
...
+        sha256_checksum: SHA256 Checksum of the runner application.
     """
 
     os: Literal["linux", "win", "osx"]
     architecture: Literal["arm", "arm64", "x64"]
     download_url: str
     filename: str
-    temp_download_token: NotRequired[str]
-    sha256_checksum: NotRequired[str]
+    # flake8-docstrings-complete thinks these attributes should not be described in the docstring.
```
jdkandersson: Perhaps double check this?

templates/openstack-userdata.sh.j2
```diff
@@ -13,3 +13,10 @@ su - ubuntu -c "cd ~/actions-runner && ./config.sh \
 
 # Run runner
 su - ubuntu -c "cd ~/actions-runner && /home/ubuntu/actions-runner/run.sh"
+<<<<<<< HEAD
+
+# cleanup
+sleep 2
+ps -aux
+=======
+>>>>>>> c57beb0daae5a7c242a7eb89409db8b6d815029b
```
jdkandersson, line 22: Should this be here?

jdkandersson (commented):
Thanks for putting this together, looks like there was a lot to fix!

src/charm.py
yanksyoon: Yes, mypy will error with the following error if the type is removed.
```
src/charm.py:100: error: Function is missing a type annotation for one or more arguments  [no-untyped-def]
```

src/charm.py
yanksyoon: I think this is a bug with flake8 with custom types - this does error with the following error message:
```
github-runner-operator/src/charm.py:101:1: D417 Missing argument descriptions in the docstring
```

templates/openstack-userdata.sh.j2
yanksyoon: Fixed, seems like VS-Code wasn't quite able to check this conflict.

.woke.yaml
```diff
@@ -1,3 +1,4 @@
-ignore_files:
-  # Ignore pylintrc as it uses non compliant terminology: whitelist
-  - .pylintrc
+rules:
+  # Ignore whitelist - we are using it to ignore pydantic in pyptoject.toml and pylintrc
+  - name: whitelist
+
```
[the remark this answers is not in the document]
yanksyoon: Fixed, thank you!

src/charm.py
jdkandersson: Hmm, perhaps because of the decorator

jdkandersson (dismissed):

src/shared_fs.py
```diff
@@ -294,4 +320,8 @@ def move_to_quarantine(runner_name: str) -> None:
             f"Failed to archive shared filesystem for runner {runner_name}"
         ) from exc
 
-    delete(runner_name)
+    try:
+        delete(runner_name)
+    except DeleteSharedFilesystemError as exc:
+        logger.error("Error deleting shared filesystem, %s", exc)
+        raise
```
cbartz, line 327: I think this will lead to duplicate logs (e.g. the caller in runner_metrics will also log the exception (with trace)). I have seen this pattern (catch exc, log and reraise) in several places in this PR and I wonder if we should not generally let the caller log such exceptions. The callee should throw the exception and let the caller do what they want with it. Thoughts @jdkandersson @gregory-schiano ?

src/utilities.py
```diff
@@ -210,6 +221,10 @@ def bytes_with_unit_to_kib(num_bytes: str) -> int:
     Args:
         num_bytes: A positive integer followed by one of the following unit: KiB, MiB, GiB, TiB,
             PiB, EiB.
+
+    Raises:
+        ValueError: If invalid unit waas detected.
```
cbartz, line 226:
```suggestion
        ValueError: If invalid unit was detected.
```

tests/unit/factories.py
```diff
@@ -50,58 +69,69 @@ class Meta:
...
 
     relations: dict[str, list] = {COS_AGENT_INTEGRATION_NAME: [], DEBUG_SSH_INTEGRATION_NAME: []}
 
 
 class MockGithubRunnerCharmFactory(factory.Factory):
-    """Mock GithubRunnerCharm."""
+    """Mock GithubRunnerCharm."""  # noqa: DCO060
+
+    # Docstrings have been abbreviated for factories, checking for docstrings on model attributes
+    # can be skipped.
```
cbartz, line 109: Could you please move the explanation of DC060 to the top so that the reader who is not looking at the whole file does not have to look it up? Thank you.

tests/unit/mock.py
```diff
@@ -44,39 +45,87 @@ class MockLxdInstanceManager:
...
+        Returns:
+            Instance with given name.
+        """
         return self.instances[name]
 
     def all(self):
+        """Return all instances that has not been deleted.
+
+        Returns:
+            All Lxd fake instances that has not been deleted.
```
cbartz, line 79:
```suggestion
            All Lxd fake instances that have not been deleted.
```

tests/unit/mock.py
```diff
@@ -86,67 +135,157 @@ class MockLxdInstance:
...
+            If storage pool of given name exists.
+        """
         if name in self.pools:
-            return not self.pools[name].delete
+            return not self.pools[name].deleted
         else:
             return False
 
     def create(self, config):
+        """Fake LxdStoragePoolManager create runction.
```
cbartz, line 281:
```suggestion
        """Fake LxdStoragePoolManager create function.
```

tests/unit/test_runner_manager.py
```diff
@@ -174,10 +174,27 @@ def test_update_runner_bin(runner_manager: RunnerManager):
...
+
+            Args:
+                args: Placeholder for positional arguments.
+                kwargs: Placeholder for keyword arguments.
+            """
             self.status_code = 200
 
-        def iter_content(self, *arg, **kargs):
+        def iter_content(self, *args, **kwargs):
+            """Mock content iterator reteruning an iterator over a single test runner binary.
```
cbartz, line 189:
```suggestion
            """Mock content iterator returning an iterator over a single test runner binary.
```

cbartz (commented):
Looks good, thanks! I have mostly nitpicks, and a problem with the catch exc/log/raise pattern, as it can lead to duplicate logs.

tests/unit/test_runner_manager.py
yanksyoon: Thanks for catching all the errors/typos @cbartz !

src/metrics.py
```diff
@@ -124,6 +125,7 @@ class RunnerStop(Event):
         status: A string describing the reason for stopping the runner.
         status_info: More information about the status.
         job_duration: The duration of the job in seconds.
+        job_conclusion: The job conclusion, e.g. "success", "neutral", "failure", ...
```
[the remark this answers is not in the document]
yanksyoon: You're right, I saw this on the [discussion](https://github.com/orgs/community/discussions/9875) and mistook it for documentation off of memory :D

src/shared_fs.py
yanksyoon: Right, this was done to please the linter (re-raise without doing anything to exc). I think we may want to have a linter that lets `Raises` section without explicitly needing to raise the exception. (cc @yhaliaw , I think you mentioned there weren't quite a linter to make sure all exceptions are covered, something along the context here that you might be interested in)

jdkandersson (dismissed):

src/shared_fs.py
cbartz: I am not sure if we should add code complexity (try catch raise) just to please the linter. I think the linting rules come from https://github.com/jdkandersson/flake8-docstrings-complete#fix-dco051, which is related to https://jdkandersson.com/2023/01/07/writing-great-docstrings-in-python/, which says not to include exceptions in the docstring if they come from a function within a function:

>  I tend not to include exceptions raised by functions called within the function since that can explode in complexity. The function should gracefully handle those exceptions anyway. 

So maybe remove the exceptions from the docstring and these constructs at least for functions marked private (so not part of the public module interface)?

src/runner_metrics.py
```diff
@@ -139,32 +175,20 @@ def _extract_metrics_from_fs(fs: shared_fs.SharedFilesystem) -> Optional[RunnerM
...
-    if not isinstance(post_job_metrics, dict) and post_job_metrics is not None:
-        raise CorruptMetricDataError(
-            f"Post-job metrics for runner {runner_name} is not a JSON object."
+        post_job_metrics = _extract_metrics_from_fs_file(
+            fs=fs, runner_name=runner_name, filename=POST_JOB_METRICS_FILE_NAME
         )
+        logger.debug("Post-job metrics for runner %s: %s", runner_name, post_job_metrics)
+    except CorruptMetricDataError as exc:
+        logger.error("Error extracting metrics from shared filesystem, %s", exc)
+        raise
```
cbartz, line 191: I think the try except and raise is not necessary here (the fct later raises CorruptMetricDataError).

pyproject.toml
```diff
@@ -4,30 +4,34 @@
...
     "src/lxd.py",
-    # Contains interface for calling repo policy compliance service. Tested in integration test and end to end tests.
+    # Contains interface for calling repo policy compliance service. Tested in integration test 
+    # and end to end tests.
     "src/repo_policy_compliance_client.py",
 ]
 
 [tool.coverage.report]
-fail_under = 38
+fail_under = 83
```
cbartz, line 21: Not directly related to this PR, but we may need a follow-up PR to meet the team standard (requires at least 85): https://github.com/canonical/is-charms-contributing-guide/blob/main/CONTRIBUTING.md#test-coverage

And thanks for updating this!

src/shared_fs.py
gregory-schiano: Well I tend to agree with Christopher, we should have constructs that try/except/raise just to please the linter, worst case we ignore the linter rule.
In my opinion the only condition a try/except/raise block should occur is when we want to wrap the exception in a custom error (which should be systematically done when the exception is a system or a library exception).
I agree that documenting all the underlying exception is complex and could lead to LOTs of exception, hence wrapping the relevant one in custom errors as parent custom error is an error to be excepted potentially anywhere in our application.

jdkandersson: Yeh just re-raising is generally not great. What can be reasonable is to catch and raise a different exception because the module has additional context to add to the exception. For example, at the low level, you might get an exception raised around that a file doesn't exists. Then the module might handle that exception and put it into context, such as that there was an error creating the configuration file. Then it is valuable to re-map to provide that extra context.

Log messages again can be useful if there is unique context that can only be added by a given function

I generally don't like letting exceptions bubble up because how is the caller supposed to know that they should handle it?

yanksyoon: How about re-raising it and removing the logs but silencing the linter to avoid duplicate logs?
I do agree that another exception should be defined for each module and let the caller handle those exceptions, perhaps in the refactor PR we should make sure that is the case. I don't mind defining the exceptions here either but I'm just afraid that it might introduce a lot of merge conflicts with Openstack PR.

jdkandersson: Perhaps initially just do it in the code being changed and then a follow up PR can do the rest?

src/runner_metrics.py
yanksyoon: I think it might be better to separate the two `CorruptMetricDataError`, one that can be raised from FS interaction
the other validation error from pydantic.
The reason I think it should be separated is because it is better to be explicit about what functions could be responsible for the error, rather than letting the error bleed/bubble up. For example, when we decide to change the error thrown by the `_extract_metrics_from_fs_file` as `ExtractMetricsError` from `CorruptMetricDataError`, it is easier to notice the change since it is wrapped in try/except. If we let it bubble up without the try/except, we might not have caught it and cause an error which is harder to catch later on.

Would this be a reasonable opinion to keep this?

src/shared_fs.py
yanksyoon: Updated - all direct re-raises are now pylint-ignored with comment.
```
# 2024/04/02 - We should define a new error, wrap it and re-raise it.
```
to be fixed in a followup PR.
Ticket for followup has been created with the title "GitHub self-hosted runner: code style issues fix for direct re-raises".

yanksyoon (commented):
Updated PR:
1. pylint ignore direct re-raises, to be fixed with follow up story to catch all direct-reraise to be wrapped in a better-defined exception.
2. All code should now use `from errors import <ERROR>` rather than using `import errors, errors.<ERROR>`

jdkandersson (dismissed):

src/runner_metrics.py
cbartz: Yes, I understand the reasoning. But could you delete the error log in line 190? The fct `_extract_fs` will log the whole exception trace in line 244.

cbartz (dismissed):

cbartz (dismissed):

src/charm.py
```diff
@@ -75,7 +81,7 @@ class ReconcileRunnersEvent(EventBase):
     """Event representing a periodic check to ensure runners are ok."""
 
 
-CharmT = TypeVar("CharmT")
+CharmT = Any
```
merkata, line 84: "Any" is a bit broad, is there another option here?

src/github_client.py
```diff
@@ -28,12 +29,37 @@
...
+    Args:
+        func: The target function to catch common errors for.
+
+    Returns:
+        The decorated function.
+    """
 
     @functools.wraps(func)
-    def wrapper(*args, **kwargs):
+    def wrapper(*args: ParamT.args, **kwargs: ParamT.kwargs) -> ReturnT:
```
merkata, line 49: Above I saw "Any" being used for `args` and `kwargs`, I propose to pick one or the other to be consistent.

src/runner_manager.py
```diff
@@ -481,18 +533,14 @@ def reconcile(self, quantity: int, resources: VirtualMachineResources) -> int:
         Returns:
             Difference between intended runners and actual runners.
         """
-        if self.config.are_metrics_enabled:
-            start_ts = time.time()
+        start_ts = time.time()
```
merkata, line 536: Is this relevant to flake8? It seems to remove a condition.

tests/unit/factories.py
```diff
@@ -50,58 +70,66 @@ class Meta:
...
+
         model = MagicMock
 
     planned_units = 1
     name = "github-runner"
 
 
 class MockGithubRunnerCharmModelFactory(factory.Factory):
-    """Mock github-runner charm model."""
+    """Mock github-runner charm model."""  # noqa: DCO060
```
merkata, line 96: What's a DCO060?

merkata (dismissed):
Nicely done!

weiiwang01 (dismissed):

src/runner_manager.py
yanksyoon: The conditional here makes the function have too many branching statements which flake8 wasn't happy about :( This doesn't impact the code in any way hence i've removed the conditional here.

src/charm.py
yanksyoon: Ah yes, thank you! Replaced with the actual charm type - by sarcrificing the generic function to be a little less generic :D

src/github_client.py
yanksyoon: Good point! I've diffeentiated Any with ParamT according to:

ParamT: If the function needs to be generic - the return types depend on the input types.
Any: if there is a concrete type being passed to it but we don't quite know the input, for example: Charm init functions - the args don't need to be generic but to cater for the lib I've typed it as any.

Would this be an acceptable reasoning?

tests/unit/factories.py
yanksyoon: This is to satisfy `flake8-docstrings-complete`.
The explanation for this error is at the top of the file - there was feedback to mention it once rather than repeating the reason for ignoring it on every line where it's used within the file :)
Would you prefer that it be on every line? 
Here's the explanation for DC060!
```
# DC060: Docstrings have been abbreviated for factories, checking for docstrings on model
# attributes can be skipped.
```

pyproject.toml
```diff
@@ -44,14 +48,20 @@ select = ["E", "W", "F", "C", "N", "R", "D", "H"]
 # Ignore D107 Missing docstring in __init__
 ignore = ["W503", "E501", "D107"]
 # D100, D101, D102, D103: Ignore missing docstrings in tests
```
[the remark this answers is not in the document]
yanksyoon: Thanks for catching this! Updated the description!

weiiwang01 (approved):

merkata (approved):

cbartz (approved):

PR #252, opened 2024-04-05
Align upgrade_charm hook with install hook

cbartz:
### Overview

Match the code in the upgrade_charm hook to that in the install hook. 

Call 

- `runner_manager.build_runner_image` (conditional if no image is available)
- `runner_manager.schedule_build_runner_image`
- `_set_reconcile_timer`

in both hooks.

### Rationale

Avoid the need for redeployments and the operator having to read the release list to decide if they need to redeploy or if they can just do a charm upgrade.

upgrade-charm event should have similar long-term side effects on the system as the `install` event.
Also, the code currently relies on `config-changed` to be triggered (which calls `_set_reconcile_timer`). The code should be written agnostic to a guaranteed event chain ([recommended by the Juju team](https://matrix.to/#/!xzmWHtGpPfVCXKivIh:ubuntu.com/$QgmiPrkorxFIcW4N8wp-TM_QUsxmJj5NaYbdza66EIA?via=ubuntu.com&via=matrix.org&via=mx.aouss.it)).

We also need to call `build_runner_image` (if there is no image) in upgrade-charm to ensure that upgrading from old charm revisions works (otherwise there would be no image after the upgrade).

### Module Changes

- `charm._on_install` and `charm._on_upgrade_charm` are adapted and common code is extracted in another method.
- `lxd.LxdImageManager.exists`  and `runner_manager.RunnerManager.has_runner_image` has been added to check if an image is already there

### Checklist


src/charm.py
```diff
@@ -356,6 +356,31 @@ def _block_on_openstack_config(self, state: CharmState) -> bool:
...
+            self._install_deps()
+            self._start_services(state.charm_config.token, state.proxy_config)
+            metrics.setup_logrotate()
+        except (LogrotateSetupError, SubprocessError) as err:
+            logger.exception(err)
+            if isinstance(err, LogrotateSetupError):
+                msg = "Failed to setup logrotate"
+            else:
+                msg = "Failed to install dependencies"
+            self.unit.status = BlockedStatus(msg)
```
amandahla, line 377: What would be the action to unblock the charm?

src/charm.py
```diff
@@ -356,6 +356,31 @@ def _block_on_openstack_config(self, state: CharmState) -> bool:
             return True
         return False
 
+    def _common_install_code(self, state: CharmState) -> bool:
```
yanksyoon, line 359: May I suggest moving this to perhaps `lxd.py`? I think that will help us start separating openstack implementations to lxd implementations for the refactor :D

src/charm.py
```diff
@@ -356,6 +356,31 @@ def _block_on_openstack_config(self, state: CharmState) -> bool:
...
+        """Installation code shared between install and upgrade hook.
+
+        Returns:
+            Whether the installation code was successful.
+        """
+        self.unit.status = MaintenanceStatus("Installing packages")
+        try:
+            # The `_start_services`, `_install_deps` includes retry.
+            self._install_deps()
```
yanksyoon, line 368: Out of the scope of this implementation but I'm wondering if the way we do apt-installs can use the [apt lib for machine charms](https://charmhub.io/operator-libs-linux/libraries/apt) rather than doing subprocess commands. This might be beneficial so that we can keep the SubprocessError separate to custom execution commands rather than apt-install and command executions.

There is a separate ticket for it (ISD-1394) so no need to do it here to keep the scope!

src/charm.py
```diff
@@ -356,6 +356,31 @@ def _block_on_openstack_config(self, state: CharmState) -> bool:
...
+            self._start_services(state.charm_config.token, state.proxy_config)
+            metrics.setup_logrotate()
+        except (LogrotateSetupError, SubprocessError) as err:
```
yanksyoon, line 371: I'm wondering if it would help to keep the exception scopes limited to the function that can raise it - i.e. try except block only for metrics.setup_logrotate for LogrotateSetupError and SubprocessError for apt-installs.

tests/integration/test_charm_no_runner.py
```diff
@@ -191,3 +192,29 @@ async def test_reconcile_runners(model: Model, app_no_runner: Application) -> No
...
+async def test_charm_upgrade(model: Model, app_no_runner: Application, charm_file: str) -> None:
+    """
+    arrange: A working application with no runners.
+    act: Upgrade the charm.
+    assert: The upgrade_charm hook ran successfully.
+    """
+    await app_no_runner.refresh(path=charm_file)
+
+    # It may take some time between the charm upgrade and the run of the upgrade charm hook,
+    # so set the idle_period high.
```
yanksyoon, line 208: I can feel the pain of you having to have waited to have found this :joy:

tests/integration/test_charm_no_runner.py
```diff
@@ -191,3 +192,29 @@ async def test_reconcile_runners(model: Model, app_no_runner: Application) -> No
...
+    unit = app_no_runner.units[0]
+    unit_name_without_slash = unit.name.replace("/", "-")
+    ret_code, stdout = await run_in_unit(
+        unit=unit, command=f"cat /var/log/juju/unit-{unit_name_without_slash}.log"
+    )
+    assert ret_code == 0, f"Failed to read the log file: {stdout}"
+    assert stdout, "Log file is empty."
+    assert "Emitting Juju event upgrade_charm." in stdout, "upgrade_charm event not yet fired."
+    # There may be a race condition: if the hook is executed shortly after idle,
+    # it may not have finished yet, so check for idle again.
```
yanksyoon, line 219: I'm wondering if the `helpers.wait_for` function might help here for the log message to appear before doing the assertions.

yanksyoon (dismissed):
Thanks for fixing this bug! It's great to be able to test upgrades from now on!
(minor comments)

src/charm.py
cbartz: I think `lxd.py` is more about abstracting low-level lxd commands. Have you and @yhaliaw thought about where to put the openstack related charm code? I feel it is better to discuss this first before I move it somewhere and it is out of scope for this PR.

src/charm.py
cbartz: Thanks, I have updated the code to go into error state.

tests/integration/test_charm_no_runner.py
```diff
@@ -204,17 +204,15 @@ async def test_charm_upgrade(model: Model, app_no_runner: Application, charm_fil
...
-    ret_code, stdout = await run_in_unit(
-        unit=unit, command=f"cat /var/log/juju/unit-{unit_name_without_slash}.log"
-    )
-    assert ret_code == 0, f"Failed to read the log file: {stdout}"
-    assert stdout, "Log file is empty."
-    assert "Emitting Juju event upgrade_charm." in stdout, "upgrade_charm event not yet fired."
-    # There may be a race condition: if the hook is executed shortly after idle,
-    # it may not have finished yet, so check for idle again.
+
+    async def is_upgrade_charm_event_emitted():
```
yanksyoon, line 210: I think flake8 would complain about this not having a docstring, would you be able to include it please? LGTM!

yanksyoon (dismissed):

tests/integration/test_charm_no_runner.py
cbartz: updated

tests/integration/helpers.py
```diff
@@ -565,10 +565,11 @@ async def dispatch_workflow(
 
 P = ParamSpec("P")
 R = TypeVar("R")
+S = Callable[P, R] | Callable[P, Awaitable[R]]
 
 
 async def wait_for(
-    func: Callable[P, R],
+    func: S,
```
yanksyoon, line 572: 🎉

yanksyoon (dismissed):

yanksyoon (dismissed):

src/charm.py
```diff
@@ -550,36 +562,28 @@ def _on_upgrade_charm(self, _: UpgradeCharmEvent) -> None:
...
+        logger.info("Flushing the runners...")
         runner_manager.flush(FlushMode.FLUSH_BUSY_WAIT_REPO_CHECK)
         self._reconcile_runners(
             runner_manager,
             state.runner_config.virtual_machines,
             state.runner_config.virtual_machine_resources,
         )
 
+        # Ensure that an update to the systemd timer for reconciliation event is written to disk.
+        # The upgrade_charm event is not necessarily followed by a config changed event
```
javierdelapuente, line 583: Reading here it says the opposite (there is always a config-changed):  https://juju.is/docs/sdk/upgrade-charm-event ```In some cases, [config-changed](https://juju.is/docs/sdk/config-changed-event) can be used instead of install and upgrade-charm because it is guaranteed to fire after both.```

cbartz: Thanks for finding this. This contractdicts my experience (reconciliation timer was not updated after a charm upgrade) and the graph in https://juju.is/docs/sdk/charm-lifecycle does not show this either. We'll have to check with the Juju team.

javierdelapuente (dismissed):

src/charm.py
cbartz: fwiw https://juju.is/docs/sdk/charm-lifecycle has been updated

yanksyoon (approved):
Minor nitpick that doesn't need to be implemented here, thank you!

javierdelapuente (approved):

src/lxd.py
```diff
@@ -575,6 +575,24 @@ def create(self, name: str, path: Path) -> None:
         if result.returncode != 0:
             raise LxdError(result.stdout.decode("utf-8"))
 
+    def exists(self, alias: str) -> bool:
```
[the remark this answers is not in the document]
cbartz: Don't you think this might be redundant as it is a method on the `LxdImageManager`?

jdkandersson (approved):

PR #254, opened 2024-04-08
chore: move lxd related files to lxd_cloud module

yanksyoon:
Applicable spec: N/A

### Overview

This moves all lxd related files to new lxd_cloud module.

### Rationale

To prepare for refactor which caters multiple cloud types.

### Juju Events Changes

None.

### Module Changes

All LXD related files have been moved to lxd_cloud module.

### Library Changes

None.

### Checklist


cbartz (commented):
One comment, otherwise it looks fine to me. I assume that some of the functionality now placed in `lxd_cloud` will be extracted into more common modules in the future (e.g. `runner_logs` is almost cloud agnostic).

src/lxd_cloud/__init__.py
```diff
@@ -0,0 +1,4 @@
+# Copyright 2024 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+"""Module for managing local LXD cloud."""
```
yhaliaw, line 4: Would this module be local LXD specific or LXD in general?
We might need to support remote LXD cloud in the future.

yhaliaw (commented):
Maybe name the module something like `local_lxd`, as the module seems to be focused on running runners locally.

src/lxd_cloud/__init__.py
yanksyoon: I was thinking of having an LXD module and separating it via local/remote flag since many of the code within them will be shared. What do you think?

yhaliaw: The local runners needs some logic specific to it, e.g., firewall setup, running a local repo-policy-compliance, etc.

Would it be good for to have the following modules with a eventual RunnerManager interface implementation:
- openstack cloud
- lxd cloud
- local lxd

and modules for the APIs (might be a single file):
- openstack API
- lxd API
- github API

yanksyoon: Sounds good 👍

PR #257, opened 2024-04-08
hotfix: support dash label

yanksyoon:
Applicable spec: N/A

### Overview

Supports dash char to labels

### Rationale

To support labels like two-xlarge

### Juju Events Changes

None.

### Module Changes

None.

### Library Changes

None.

### Checklist

### DIFF
--- a/tests/unit/test_charm_state.py
+++ b/tests/unit/test_charm_state.py
@@ -371,8 +371,14 @@ def test__parse_labels_invalid_labels(label_str: str, falsy_labels: tuple[str]):
         pytest.param(" a, b,   c", ("a", "b", "c"), id="comma separated labels with space"),
         pytest.param("1234", ("1234",), id="numeric label"),
         pytest.param("_", ("_",), id="underscore"),
+        pytest.param("-", ("-",), id="dash only"),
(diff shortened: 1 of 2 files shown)


tests/unit/test_charm_state.py
```diff
@@ -371,8 +371,14 @@ def test__parse_labels_invalid_labels(label_str: str, falsy_labels: tuple[str]):
         pytest.param(" a, b,   c", ("a", "b", "c"), id="comma separated labels with space"),
         pytest.param("1234", ("1234",), id="numeric label"),
         pytest.param("_", ("_",), id="underscore"),
+        pytest.param("-", ("-",), id="dash only"),
```
cbartz, line 374: Not sure if you have already checked, but we may need to check if this is actually an allowed label. But can be done in a follow-up PR.

cbartz (approved):

tests/unit/test_charm_state.py
yanksyoon: image
Github allows many of the labels that we don't allow - while I think flexibility is good, I think we should limit our offering to this set first so that we reduce security risks :)

arturo-seijas (approved):

PR #266, opened 2024-04-23
Add COS integration to Openstack

cbartz:
Applicable spec: ISD-145 (internal)

### Overview

Add cos integration to the openstack branch. The Openstack code is going to be refactored (see ISD 134), so there are parts that contain code duplication and should be addressed in a follow-on.

Integration tests are not included in this PR and will be added in a future PR.

The log retrieval code will be adapted in a future PR to retrieve logs periodically (also for LXD) and is not included here.

### Rationale

This is necessary for observability.

### Juju Events Changes
n/a

### Module Changes
- All common metrics-related modules are moved to the `metrics` package.
- A [python protocol](https://peps.python.org/pep-0544/) is defined to manage metrics storage, and `shared_fs` implements it. This is done to reuse the `runner` metrics module (see also ISD-145).
- `OpenStackRunnerManager` is adapted to output runner related metrics.
- `runner_logs` is split into general and lxd related code
- cloud init file (`openstack-userdata.j2`) is adapted to write out metrics and pre-job content. system shutdown is removed to be able to retrieve metrics (otherwise the instance would have to be restarted).
- pre-job content is adapted to optionally do the repo policy check

### Checklist

The Openstack documentation will be added as soon as we have a MVP landed.


src/openstack_cloud/openstack_manager.py
```diff
@@ -1172,3 +1249,222 @@ def flush(self) -> int:
...
+            except _PullFileError as exc:
+                logger.warning(
+                    "Failed to pull metrics for %s: %s . Will not be able to issue all metrics",
+                    instance_name,
+                    exc,
+                )
+                return
+            except _SSHError as exc:
+                logger.info("Failed to pull metrics for %s: %s", instance_name, exc)
+                continue
```
cbartz, line 1294: The downside here is that if only the `pre-job-metrics.json` is pulled, the runner is considered crashed (see logic in `_issue_reconciliation_metric`), we may want to revisit this. But for now I kept it the same as for LXD.

src/metrics/runner.py
```diff
@@ -118,12 +120,12 @@ def _inspect_file_sizes(fs: shared_fs.SharedFilesystem) -> tuple[Path, ...]:
 
 
 def _extract_metrics_from_fs_file(
```
yhaliaw, line 122: Since the `fs` args are being renamed, would it be good to rename the `fs` in the method names?

src/openstack_cloud/openstack_manager.py
```diff
@@ -1172,3 +1249,222 @@ def flush(self) -> int:
...
+
+    def _pull_file(
```
yhaliaw, line 1296: I think this method should be above `_pull_metrics`, since it relies on this method.

src/openstack_cloud/openstack_manager.py
```diff
@@ -1172,3 +1249,222 @@ def flush(self) -> int:
...
+            local_path: The local path to store the file.
+            max_size: If the file is larger than this, it will not be pulled.
+
+        Raises:
+            _PullFileError: Unable to pull the file from the runner instance.
+            _SSHError: Issue with SSH connection.
+        """
+        try:
+            result = ssh_conn.run(f"stat -c %s {remote_path}", warn=True)
+        except NoValidConnectionsError as exc:
```
yhaliaw, line 1313: I think `TimeoutError` is possible here?

src/openstack_cloud/openstack_manager.py
```diff
@@ -1172,3 +1249,222 @@ def flush(self) -> int:
...
+            if size > max_size:
+                raise _PullFileError(
+                    reason=f"File size of {remote_path} too large {size} > {max_size}"
+                )
+        except ValueError as exc:
+            raise _PullFileError(reason=f"Invalid file size for {remote_path}: {stdout}") from exc
+
+        try:
+            ssh_conn.get(remote=remote_path, local=local_path)
+        except NoValidConnectionsError as exc:
```
yhaliaw, line 1331: Same as above.

yhaliaw (commented):
Reviewed till `_issue_runner_installed_metric`.

src/openstack_cloud/openstack_manager.py
cbartz: Thanks, we just updated the contributing guide to flip the order :D : https://github.com/canonical/is-charms-contributing-guide/pull/77 Do you think its confusing to apply the change partially? cc @yanksyoon

yhaliaw: Oh ok. I think it is fine to have it partially applied.

yanksyoon (approved):
LGTM! Minor nitpicks - no need to implement :D

PR #277, opened 2024-05-06
Integration tests for COS integration for OpenStack

cbartz:
### Overview

Follow-up to https://github.com/canonical/github-runner-operator/pull/266 to provide integration tests running on the openstack cloud.

An integration test related to COS and repo policy compliance will be provided in a follow-up PR.

The OpenStack tests are not yet included in the CI, as there is another PR working on getting Openstack to work on the CI: https://github.com/canonical/github-runner-operator/pull/269

This PR does also refactor the helper modules. Due to parallel development on the openstack integration tests, the other integration tests have not been tested to see if they still work and this will need to be investigated in a follow-up.

The openstack applicable tests are marked now with an `openstack` pytest mark.

This PR also fixes/adds:

- Private ssh keys have now correct permission and are owned by ubuntu user (to avoid ssh permission warning)
- RunnerStop metric is also issued on abnormal termination
- More context on pull metrics file errors (add a warning log that the metrics file size could not be determined)

### Rationale

Add integration tests to ensure proper functioning.

### Juju Events Changes

n/a

### Module Changes

A separate package has been introduced and the helper modules have been placed there. There is a module for LXD and Openstack that contains cloud specific code. An interface `InstanceHelper` has been defined which is implemented for both clouds. It is required to handle cloud specific helper code as `run_in_instance' or `ensure_one_runner' or `get_runner_name'.

The `test_charm_metrics_{success,failure}` tests have been adapted and the openstack applicable tests are marked with an `openstack` mark.

Beyond that:

- `openstack_manager`: Now writes ssh key with correct permission and ubuntu user
- `openstack_userdata.j2`: Fix case where write_post_metrics function is not called on non-zero exit code of previous command

### Checklist


.github/workflows/e2e_test_run.yaml
```diff
@@ -21,7 +21,7 @@ on:
 jobs:
   e2e-test:
     name: End-to-End Test Run
-    runs-on: [self-hosted, linux, x64, "${{ inputs.runner-tag }}"]
+    runs-on: [self-hosted, linux, "${{ inputs.runner-tag }}"]
```
yhaliaw, line 24: Should we modify existing test that runs on `x64` or add new tests for `arm64`?
I think we should test for both `x64` and `arm64`.
Applies for the other workflow as well.

templates/openstack-userdata.sh.j2
```diff
@@ -97,7 +97,6 @@ write_post_metrics(){
 }
 
 # Run runner
-su - ubuntu -c "cd ~/actions-runner && /home/ubuntu/actions-runner/run.sh"
-write_post_metrics $?
+(set +e; su - ubuntu -c "cd ~/actions-runner && /home/ubuntu/actions-runner/run.sh"; write_post_metrics $?)
```
yhaliaw, line 100: Is this to ensure the `write_post_metrics` only execute on the `run.sh` completed successfully?
Would `.../run.sh && write_post_metrics $?` be more readable?
Would it be good to add a comment to explain it?

tests/conftest.py
```diff
@@ -57,3 +57,10 @@ def pytest_addoption(parser: Parser):
         action="store",
         help="The OpenStack clouds yaml file for the charm to use.",
     )
+    parser.addoption(
+        "--use-existing-app",
+        action="store",
+        help="The existing app to use. "
```
yhaliaw, line 63: I think there is an extra space after the `.`.

This option is a good idea. Great work here.

tests/integration/conftest.py
```diff
@@ -26,21 +27,27 @@
...
     deploy_github_runner_charm,
-    ensure_charm_has_runner,
     reconcile,
     wait_for,
 )
+from tests.integration.helpers.lxd import LXDInstanceHelper, ensure_charm_has_runner
+from tests.integration.helpers.openstack import OpenStackInstanceHelper
 from tests.status_name import ACTIVE
 
+nest_asyncio.apply()
```
yhaliaw, line 49: Would it be good to add a short comment on what is this for?
Maybe a link to docs?

tests/integration/conftest.py
```diff
@@ -247,33 +260,40 @@ async def app_openstack_runner(
...
+            constraints={
+                "root-disk": 50 * 1024,
+                "cores": 4,
+                "mem": 16 * 1024,
+                "arch": "arm64",
+            },
+            config={
+                OPENSTACK_CLOUDS_YAML_CONFIG_NAME: openstack_clouds_yaml,
+                # this is set by microstack sunbeam, see scripts/setup-microstack.sh
+                OPENSTACK_NETWORK_CONFIG_NAME: "demo-network",
```
yhaliaw, line 289: I hardcoded this due to using microstack previously. 
I would think this will need to be different for private endpoints. How did the test work with this hardcoded value.

This would need to be an option and pass via secret, which I can do in a follow-up PR.

yhaliaw: It seems this app was not used for testing. Which would make sense why the tests passed.

yhaliaw (commented):
Some comments.

.github/workflows/e2e_test_run.yaml
cbartz: Do you mean we should run an e2e test for both x64 and arm64 at the same time? I think in this case the end-to-end test needs to be adapted to build an arm charm. I am afraid this is beyond the scope of this PR. 
I can revert to `x64` if you prefer, but I think removing `x64` is not harmful because the `runner-tag` is sufficient to identify the runner.

templates/openstack-userdata.sh.j2
cbartz: No, this is to ensure that `write_post_metrics` runs even if run.sh fails. I could simplify this by using the `||` operator.

tests/conftest.py
cbartz: Thanks, I saw this in another PR: https://github.com/canonical/bind-operator/pull/5#discussion_r1566787699

tests/integration/conftest.py
cbartz: Yes, I replaced these config options locally with something appropriate for our staging environment, but I did not want to check this into the source code, so I kept the previous value for the sake of this PR. I figured you would change it in the next PR anyway.

tests/integration/test_charm_metrics_failure.py
```diff
@@ -114,18 +118,25 @@ async def test_charm_issues_metrics_for_abnormal_termination(
...
-        unit, workflow, branch=forked_github_branch, started_time=dispatch_time
+        unit,
+        workflow,
+        branch=forked_github_branch,
+        started_time=dispatch_time,
+        instance_helper=instance_helper,
     )
 
+    # Wait a bit to ensure pre-job script has been executed.
+    await sleep(5)
```
yhaliaw, line 129: What is needed for be waited for?
Would some form of readiness/health check be a more robust approach?

templates/openstack-userdata.sh.j2
yhaliaw: Oh it is `set +e` and not `set -e`. Still would it is better as `.../run.sh || write_post_metrics`, or adding a comment?

cbartz: Using `||` would prevent write_post_metrics from running in a happy case. I'll add a comment.

tests/integration/test_charm_metrics_failure.py
cbartz: I came across strange cases where the timestamp of RunnerStop was lower than that of RunnerStart. My suspicion is that the pre-job was executed after the post-job phase. I added a sleep to avoid this case.

.github/workflows/e2e_test_run.yaml
yhaliaw: It might be fine.

tests/integration/helpers/lxd.py
```diff
@@ -0,0 +1,257 @@
...
+    await run_in_unit(unit, "/usr/bin/systemctl start test-http-server")
+
+    # Test the HTTP server
+    for _ in range(10):
+        return_code, stdout, _ = await run_in_unit(unit, f"curl http://localhost:{port}")
+        if return_code == 0 and stdout:
+            break
+        await sleep(3)
+    else:
+        assert False, "Timeout waiting for HTTP server to start up"
```
yanksyoon, line 218: I think we can leverage `helpers.wait_for` to make it a little more standardized when we need to wait for something :D

yanksyoon (commented):
Reviewed until `helpers/lxd.py`, cool stuffs!

tests/integration/conftest.py
```diff
@@ -566,12 +588,49 @@ def get_branch():
...
+    if not existing_app:
+        grafana_agent = await model.deploy(
+            "grafana-agent",
+            application_name=f"grafana-agent-{basic_app.name}",
+            channel="latest/edge",
+            revision=108,
+        )
+        await model.relate(f"{basic_app.name}:cos-agent", f"{grafana_agent.name}:cos-agent")
+        await model.wait_for_idle(apps=[basic_app.name], status=ACTIVE)
+        await model.wait_for_idle(apps=[grafana_agent.name])
```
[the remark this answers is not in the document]
cbartz: Correct. I'll add more explanations to the config option help string.

tests/integration/helpers/lxd.py
```diff
@@ -0,0 +1,257 @@
...
+    return_code, stdout, _ = await run_in_unit(unit, f"lxc profile show {resource_profile_name}")
+    assert return_code == 0
+    assert stdout is not None
+    profile_content = yaml.safe_load(stdout)
+    assert f"{cpu}" == profile_content["config"]["limits.cpu"]
+    assert mem == profile_content["config"]["limits.memory"]
+    assert disk == profile_content["devices"]["root"]["size"]
+
+
+async def get_runner_names(unit: Unit) -> tuple[str, ...]:
```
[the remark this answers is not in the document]
cbartz: I think this is no longer necessary since python 3.9: https://peps.python.org/pep-0585/

tests/integration/helpers/lxd.py
```diff
@@ -0,0 +1,257 @@
...
+        return json.loads(stdout)
+
+    async def is_desired_num_runners():
+        """Return whether there are desired number of lxc instances running.
+
+        Returns:
+            Whether the desired number of lxc runners have been reached.
+        """
+        lxc_instances = await get_lxc_instances()
+        if lxc_instances is None:
```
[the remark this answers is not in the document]
cbartz: I am not the author (just moved it), but it seems that `None` is used to indicate an error, so we need to distinguish the None case from the empty list case (a caller could request `num==0`, which would be fine with an empty list).

.github/workflows/e2e_test_run.yaml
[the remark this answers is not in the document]
cbartz: Yes, but I'm afraid it's beyond the scope of this PR.

tests/integration/test_charm_metrics_failure.py
cbartz: Have removed the sleep.

tests/integration/helpers/common.py
```diff
@@ -52,6 +48,40 @@
 logger = logging.getLogger(__name__)
 
 
+class InstanceHelper(typing.Protocol):
+    """Helper for running commands in instances."""
+
+    # noqa
```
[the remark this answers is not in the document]
cbartz: I removed it, it was unnecessary.

yhaliaw (approved):

yhaliaw:
@yanksyoon I think you can merge this, when you approve it.

yanksyoon (approved):
LGTM! Thanks for the changes!

PR #286, opened 2024-05-23
Add chart displaying available applications

cbartz:
Applicable spec: n/a

### Overview

1. Rename flavour to application in the dashboard panels.
2. Add a chart close to the input variable listing all applications.

For the short-term board, list all available runners (idle runners after last reconciliation).

image

For the long-term board, show all runners created per application.
image

### Rationale

1. The term flavour is nowhere well defined, and we are really referring to the charm application that manages the runner instances.
2. The panel should help to list all available applications which can be filtered.

### Juju Events Changes

n/a

### Module Changes

n/a

### Library Changes

n/a

### Checklist

### DIFF
--- a/docs/how-to/debug-with-ssh.md
+++ b/docs/how-to/debug-with-ssh.md
@@ -7,7 +7,7 @@ for automatic configuration.
 
 ## Prerequisites
 
-To enhance the security of self-hosted runners and its infrastracture, only authorized connections
+To enhance the security of self-hosted runners and its infrastructure, only authorized connections
 can be established. Hence, action-tmate users must have
 [ssh-key registered](https://docs.github.com/en/authentication/connecting-to-github-with-ssh/adding-a-new-ssh-key-to-your-github-account)
 on the GitHub account.
--- a/docs/reference/cos.md
+++ b/docs/reference/cos.md
@@ -12,8 +12,8 @@ The "GitHub Self-Hosted Runner Metrics" metrics dashboard presents the following
 
 - General: Displays general metrics about the charm and runners, such as:
   - Lifecycle counters: Tracks the frequency of Runner initialisation, start, stop, and crash events.
-  - Jobs per Application: Shows the number of jobs per charm application.
-  - Idle runners after reconciliation: Reflects the count of Runners marked as idle during the last reconciliation event. Note: This data updates post-reconciliation events and isn't real-time.
+  - Available runners: A horizontal bar graph showing the number of runners available during the last reconciliation event. Note: This data is updated after each reconciliation event and is not real-time.
+  - Idle runners after reconciliation: A time series graph showing the number of runners marked as idle during the last reconciliation event over time. Note: This data is updated after each reconciliation event and is not real-time.
   - Duration observations: Each data point aggregates the last hour and shows the 50th, 90th, 95th percentile and maximum durations for:
       - Runner installation
       - Runner idle duration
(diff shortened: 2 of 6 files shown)


arturo-seijas (commented):
I think the jobs per application metric would be better displayed as a bar chart. The current chart doesn't seem to intuitive to me

arturo-seijas:
> @arturo-seijas
> 
> > I think the jobs per application metric would be better displayed as a bar chart. The current chart doesn't seem to intuitive to me
> 
> The problem with bar charts is that the focus is on the graph and not on the legend. The idea here is to present a list of available applications, so the focus should be on the text. My first version was as follows:
> 
> image

What about an horizontal bar chart, with the legend on the left side?

arturo-seijas:
> > What about an horizontal bar chart, with the legend on the left side?
> 
> @arturo-seijas This would look like
> 
> image
> 
> But as mentioned the emphasis would be a bit fewer on the application names.
> 
> There is also already a panel at the bottom of the dashboard that shows the application share and looks like this:
> 
> image
> 
> This could simply be reused and moved to the top to indicate the list of applications.
> 
> Perhaps @jdkandersson can give his opinion on this, as the new panel was requested by him.

When I see the application share I find too difficult to see the values for those that are not at the top nor bottom. In the screenshot above, what are the values for `large`? Again, I think bar charts are in order here, something like [this](https://www.reddit.com/media?url=https%3A%2F%2Fpreview.redd.it%2Fpossible-to-create-a-clustered-bar-chart-showing-percentage-v0-wp7fjk1egqlb1.png%3Fwidth%3D454%26format%3Dpng%26auto%3Dwebp%26s%3Decef192c5deae3e400ca5ef2d944d40cf098cac1)

cbartz:
> When I see the application share I find too difficult to see the values for those that are not at the top nor bottom. In the screenshot above, what are the values for `large`? Again, I think bar charts are in order here, something like [this](https://www.reddit.com/media?url=https%3A%2F%2Fpreview.redd.it%2Fpossible-to-create-a-clustered-bar-chart-showing-percentage-v0-wp7fjk1egqlb1.png%3Fwidth%3D454%26format%3Dpng%26auto%3Dwebp%26s%3Decef192c5deae3e400ca5ef2d944d40cf098cac1)

@arturo-seijas Ok, gotcha. I'm fine with using bar charts, but I think I need to re-confirm that this meets the acceptance criteria, as the original story was just to visualise the names of applications, not the number of jobs or anything. I just added it because the log panel only showing the names didn't look "pretty" to me.  See also the [jira ticket](https://warthogs.atlassian.net/jira/software/c/projects/ISD/boards/1304?selectedIssue=ISD-1845).

varshigupta12:
@cbartz I am wondering if all we need is an easy way for users to search by application- can we change the Flavors textbox to be a dropdown instead? Something like-

image

cbartz:
> @cbartz I am wondering if all we need is an easy way for users to search by application- can we change the Flavors textbox to be a dropdown instead? Something like-
> 
> image

@varshigupta12 The text box is automatically generated by Grafana and is not intended for further customisation. There are drop-down menus for "Query variables", but unfortunately these are not applicable here because the "flavor" is not a real label. It might be possible to turn it into a real label, which might require some customisation of the grafana-agent charm. I think we (@jdkandersson and I) decided not to put too much effort into solving this problem, but it could certainly be investigated together with the observability team.

jdkandersson:
> > @cbartz I am wondering if all we need is an easy way for users to search by application- can we change the Flavors textbox to be a dropdown instead? Something like-
> > image
> 
> @varshigupta12 The text box is automatically generated by Grafana and is not intended for further customisation. There are drop-down menus for "Query variables", but unfortunately these are not applicable here because the "flavor" is not a real label. It might be possible to turn it into a real label, which might require some customisation of the grafana-agent charm. I think we (@jdkandersson and I) decided not to put too much effort into solving this problem, but it could certainly be investigated together with the observability team.

Yeh we couldn't find a way to do a drop down and this is the alternative.

Perhaps we do the horizontal bar chart and make the number the number of available runners? I think that would be a useful chart anway

arturo-seijas (dismissed):

arturo-seijas (approved):

cbartz:
@jdkandersson Can you please review? Thanks.

jdkandersson (approved):

PR #295, opened 2024-06-06
fix: key error on charm state upgrade

yanksyoon:
Applicable spec: N/A

### Overview

Ignore KeyError on `immutable config changed`, since this is due to upgraded charm code not being aware of old saved JSON.
The code later on saves the new config correctly.

### Rationale

During charm upgrades, when we introduce a new persisted charm config that is immutable, we store them in a
persistent json file. When this is loaded, the previous state (the charm before the upgrade) has no idea about the added key and hence raises a KeyError. This PR fixes it.

### Module Changes

- `charm_state.py` - handles KeyError in `_check_immutable_config_changed`.

### Checklist

### DIFF
--- a/src/charm_state.py
+++ b/src/charm_state.py
@@ -846,23 +846,37 @@ class CharmState:
         prev_state = json.loads(json_data)
         logger.info("Previous charm state: %s", prev_state)
 
-        if prev_state["runner_config"]["runner_storage"] != runner_storage:
-            logger.error(
-                "Storage option changed from %s to %s, blocking the charm",
-                prev_state["runner_config"]["runner_storage"],
-                runner_storage,
+        try:
+            if prev_state["runner_config"]["runner_storage"] != runner_storage:
+                logger.error(
+                    "Storage option changed from %s to %s, blocking the charm",
+                    prev_state["runner_config"]["runner_storage"],
+                    runner_storage,
+                )
+                raise ImmutableConfigChangedError(
+                    msg=(
+                        "runner-storage config cannot be changed after deployment, "
+                        "redeploy if needed"
+                    )
+                )
+        except KeyError as exc:
+            logger.warning(
+                "Key %s not found, this will be updated to current config.", exc.args[0]
             )
-            raise ImmutableConfigChangedError(
-                msg="runner-storage config cannot be changed after deployment, redeploy if needed"
-            )
-        if prev_state["runner_config"]["base_image"] != base_image.value:
-            logger.error(
-                "Base image option changed from %s to %s, blocking the charm",
-                prev_state["runner_config"]["base_image"],
-                runner_storage,
-            )
-            raise ImmutableConfigChangedError(
-                msg="base-image config cannot be changed after deployment, redeploy if needed"
+
+        try:
+            if prev_state["runner_config"]["base_image"] != base_image.value:
+                logger.error(
+                    "Base image option changed from %s to %s, blocking the charm",
+                    prev_state["runner_config"]["base_image"],
+                    runner_storage,
+                )
+                raise ImmutableConfigChangedError(
+                    msg="base-image config cannot be changed after deployment, redeploy if needed"
+                )
+        except KeyError as exc:
+            logger.warning(
+                "Key %s not found, this will be updated to current config.", exc.args[0]
             )
 
     @classmethod
--- a/src/charm_state.py
+++ b/src/charm_state.py
@@ -860,9 +860,7 @@ class CharmState:
                     )
                 )
         except KeyError as exc:
-            logger.info(
-                "Key %s not found, this will be updated to current config.", exc.args[0]
-            )
+            logger.info("Key %s not found, this will be updated to current config.", exc.args[0])
 
         try:
             if prev_state["runner_config"]["base_image"] != base_image.value:
@@ -875,9 +873,7 @@ class CharmState:
                     msg="base-image config cannot be changed after deployment, redeploy if needed"
                 )
         except KeyError as exc:
-            logger.info(
-                "Key %s not found, this will be updated to current config.", exc.args[0]
-            )
+            logger.info("Key %s not found, this will be updated to current config.", exc.args[0])
 
     @classmethod
     def from_charm(cls, charm: CharmBase) -> "CharmState":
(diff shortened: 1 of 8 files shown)


tests/unit/test_charm_state.py
```diff
@@ -929,6 +929,41 @@ def mock_charm_state_data():
...
+)
+def test_check_immutable_config_key_error(
+    mock_charm_state_path: Path,
+    mock_charm_state_data: dict[str, typing.Any],
+    immutable_config: str,
+    monkeypatch: pytest.MonkeyPatch,
+    caplog: pytest.LogCaptureFixture,
+):
+    """
+    arrange: Mock CHARM_STATE_PATH and read_text method to return previous state with same config.
```
cbartz, line 947: Shouldn't this be phrased more like "to return previous state without particular config"?

tests/unit/test_charm_state.py
```diff
@@ -929,6 +929,41 @@ def mock_charm_state_data():
...
+    act: Call _check_immutable_config_change method.
+    assert: None is returned.
+    """
+    mock_charm_state_data["runner_config"].pop(immutable_config)
```
cbartz, line 951: Shouldn't we use a cloned object, since you reuse it in line 958 (and I think you want to test that the key is included in current state but not in the previous)?

cbartz (commented):
Looks good, but I have some questions about the unit test.

tests/unit/test_charm_state.py
yanksyoon: Good point, i thought a function scoped fixture would regenerate the object, let me test it!

cbartz: My point is not that it gets regenerated, but that you change it within the test itself. So the question would be, should the state in line 958 contain the `immutable_config` or not?

yanksyoon: I see what you mean, good point, the method under test is a class method so we don't quite need to instantiate the charm state at all in the first place.
Updated code, thank you!

tests/unit/test_charm_state.py
yanksyoon: Updated docstring to: `Mock CHARM_STATE_PATH and read_text method to return modified immutable config values` would this sound good?

cbartz (dismissed):
LGTM

src/charm_state.py
```diff
@@ -846,23 +846,37 @@ def _check_immutable_config_change(
...
+                    runner_storage,
+                )
+                raise ImmutableConfigChangedError(
+                    msg=(
+                        "runner-storage config cannot be changed after deployment, "
+                        "redeploy if needed"
+                    )
+                )
+        except KeyError as exc:
+            logger.warning(
```
jdkandersson, line 863: What is the reason that this is a warning please? Isn't it entirely expected if we upgrade from a previous version that didn't have the key?

In general, the way I think about warnings is that we should have some intention of making this an error later. So use cases include that we warn that a function is deprecated and will be removed in the future. Or that a retry has been triggered twice and, if it fails again, that it will result in an error. For whoever is reading the warning in the log there should be some action, otherwise what are we warning about? Warnings shouldn't be routinely ignored

yanksyoon: I think this warning doesn't happen routinely - only during charm upgrades with backward-incompatibilities. Hence I think it is worth noting that this action is taking place.
Would you think this would then be more suitable for a debug/info log?

jdkandersson: yeh info seems appropriate.

What is meant by routinely is that every time it would have appeared as a warning there is no action to be taken by the operator, hence they would ignore it

jdkandersson (dismissed):

src/charm_state.py
cbartz: Updated the code

cbartz (dismissed):

jdkandersson (dismissed):

jdkandersson (dismissed):

cbartz (dismissed):

jdkandersson (dismissed):

tests/integration/test_charm_upgrade.py
```diff
@@ -0,0 +1,82 @@
...
+    charm_file: str,
+    app_name: str,
+    path: str,
+    token: str,
+    http_proxy: str,
+    https_proxy: str,
+    no_proxy: str,
+):
+    """
+    arrange: given latest stable version of the charm.
```
cbartz, line 26: we are hardcoding rev 161, so the latest stable is probably not true.

tests/integration/test_charm_upgrade.py
```diff
@@ -0,0 +1,82 @@
...
+        branch="deadbeef",
+        hash_="hash",
+        id_="id",
+        revision=latest_stable_revision,
+        base=client.Base("22.04", "ubuntu"),
+    )
+
+    # upgrade the charm with current local charm
+    await application.local_refresh(path=charm_file, charm_origin=origin)
+    await model.wait_for_idle(
```
cbartz, line 76: I remember when developing https://github.com/canonical/github-runner-operator/blob/main/tests/integration/test_charm_no_runner.py#L203 that it can take some time for the upgrade-charm hook to really run, so you may need to use the idle_period kwarg or something like https://github.com/canonical/github-runner-operator/blob/main/tests/integration/test_charm_no_runner.py#L217-L227  to avoid having false positives (so to ensure that the upgrade-charm really ran).

jdkandersson (approved):

tests/integration/test_charm_upgrade.py
```diff
@@ -3,35 +3,65 @@
...
 ):
     """
-    arrange: given latest stable version of the charm.
+    arrange: given latest stable version of the charm (current 161).
     act: charm upgrade is called.
     assert: the charm is upgraded successfully.
     """
-    latest_stable_revision = 161
+    latest_stable_path = tmp_path / "github-runner.charm"
+    latest_stable_revision = 161  # update this value every release to stable.
```
cbartz, line 43: I don't think this will be followed (update on very release :D) unless there is an implemented check letting the test fail.

cbartz (approved):

tests/integration/test_charm_upgrade.py
yanksyoon: Lets see if this is required in further tests - currently it's just downloading from latest/stable jammy. This field was required to mark the original source of the charm during upgrade :)

PR #296, opened 2024-06-06
Openstack: Delete keypairs on timeout

cbartz:
Applicable spec: n/a

### Overview

Delete keypairs for runners that are deleted due to creation timeout

### Rationale

otherwise there will be a keypair <-> instance drift which can cause problems if the keypair quota is exceeded.

### Juju Events Changes
n/a

### Module Changes

`openstack_cloud.openstack_manager.OpenstackRunnerManager._create_runner`: add keypair cleanup code

### Library Changes

n/a

### Checklist

### DIFF
--- a/src/openstack_cloud/openstack_manager.py
+++ b/src/openstack_cloud/openstack_manager.py
@@ -805,8 +805,17 @@ class OpenstackRunnerManager:
                         instance_config.name,
                     )
                     conn.delete_server(name_or_id=instance_config.name, wait=True)
+                    try:
+                        conn.delete_keypair(instance_config.name)
+                    except openstack.exceptions.SDKException:
+                        logger.exception(
+                            "Unable to delete OpenStack keypair %s", instance_config.name
+                        )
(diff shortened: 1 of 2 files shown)


src/openstack_cloud/openstack_manager.py
```diff
@@ -805,8 +805,17 @@ def _create_runner(args: _CreateRunnerArgs) -> None:
                         instance_config.name,
                     )
                     conn.delete_server(name_or_id=instance_config.name, wait=True)
+                    try:
+                        conn.delete_keypair(instance_config.name)
```
amandahla, line 809: Should it ignore errors if keypair does not exist?

cbartz: I think this would be thrown as an SDKException and we would log an exception. I think this should be sufficient (I basically reused the same code we have in other parts of this module for cleaning up keypairs).

yhaliaw (approved):

jdkandersson (approved):

PR #297, opened 2024-06-10
Openstack COS: Issue metrics even if runner creation fails

cbartz:
Applicable spec: n/a

### Overview

1. Output runner/reconciliation metrics even  if runner creation fails
2. Increase server timeout to 5 minutes.
3. Also refactor the big `reconcile` method a bit.

### Rationale

1. Otherwise there is no observability
2. 2 minutes may be too short
3. method was too complex (linter also complained)

### Juju Events Changes

n/a

### Module Changes

`openstack_cloud.openstack_manager.OpenstackRunnerManager.reconcile` has been refactored

### Checklist

### DIFF
--- a/src/openstack_cloud/openstack_manager.py
+++ b/src/openstack_cloud/openstack_manager.py
@@ -1219,14 +1219,63 @@ class OpenstackRunnerManager:
         Args:
             quantity: The number of intended runners.
 
-        Raises:
-            OpenstackInstanceLaunchError: Unable to launch OpenStack instance.
-
         Returns:
             The change in number of runners.
         """
         start_ts = time.time()
+        try:
+            delta = self._reconcile_runners(quantity)
+        finally:
+            end_ts = time.time()
+            self._issue_reconciliation_metrics(
+                reconciliation_start_ts=start_ts, reconciliation_end_ts=end_ts
+            )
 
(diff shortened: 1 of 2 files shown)


jdkandersson (approved):

src/openstack_cloud/openstack_manager.py
```diff
@@ -794,7 +796,7 @@ def _create_runner(args: _CreateRunnerArgs) -> None:
                     security_groups=[SECURITY_GROUP_NAME],
                     userdata=cloud_userdata_str,
                     auto_ip=False,
-                    timeout=120,
+                    timeout=CREATE_SERVER_TIMEOUT,
```
amandahla, line 799: Out of curiosity: when it fails by timeout, the server creation is aborted? Or could it fail by timeout but the server is still being created and at some point, is done and ready?

yanksyoon (approved):

yhaliaw (approved):

src/openstack_cloud/openstack_manager.py
cbartz: I think it will not be cancelled, this is also the reason why we have a cleanup mechanism in those cases, but perhaps @yhaliaw can confirm.

PR #303, opened 2024-06-18
Add dispatch workflow for e2e test for openstack integration

yhaliaw:
### Overview

Add e2e test workflow for openstack integration tests.

### Rationale

This is needed on main branch for workflow dispatch.
The arch label is not needed as the runner_id should get a unique runner as the id is randomly gen.
Openstack integration e2e test does not need to test host IP access as there is no host.
The sctp support is not in place. This should be done by the image builder charm.

### Checklist


jdkandersson (approved):

tests/integration/helpers.py
```diff
@@ -546,7 +546,20 @@ async def dispatch_workflow(
     """
     start_time = datetime.now(timezone.utc)
 
+    # TODO: Debug
```
yanksyoon, line 549: Should we remove this TODO (if we implement strict linting this would throw an error)?

tests/integration/helpers.py
```diff
@@ -546,7 +546,20 @@ async def dispatch_workflow(
...
+    print("##################################################################")
```
yanksyoon, line 550: Would logging module be helpful here? Since I'm not sure if print gets captured in the log output.

yanksyoon (approved):
LGTM, minor questions just for my understanding thank you!

yhaliaw:
This is not needed as a new approach is used.

PR #305, opened 2024-06-21
Reactive: Consume jobs from mq

cbartz:
Applicable spec: [ISD-116](https://discourse.charmhub.io/t/specification-isd116-github-self-hosted-runners-reactive-scheduling/13755/1)

### Overview

Consume jobs from the message queue to allow reactive scheduling. This PR only adds consuming and not spawning in reactive mode which will be added in a future PR.

Due to the need for new logs, a refactoring and introduction of a new logrotate module has been included in this PR.

`src-docs` has been fixed/regenerated to correctly reflect the package structure. This is also one reason for the huge line diff of the PR.

### Rationale

To enable reactive scheduling, we need to consume the requested jobs from the MQ.

### Juju Events Changes

We add mongodb integration and trigger reconciliation on the database events (`database_created` , `endpoints_changed`).

#### Changed
`charm`: Instantiates `DatabaseRequires`  and calls new logrotate module.
`charm_state`: Add `ReactiveConfig` which extracts mq uri from database.
`metrics.events`: Remove logrotate handling code (but add logrotate configuration)
~`metrics.runner_logs`: Remove manual log rotation code and use new logrotate module instead~
`openstack_cloud.openstack_manager`: call reactive runner manager on reconciliation
`runner_manager`: call reactive runner manager on reconciliation

#### New
`logrotate`: Module that sets up logrotate accordingly.
`reactive.consumer`: Implements the consumer logic for a reactive runner. Currently no runner is spawned, only the consumed message is logged.
`reactive.runner`: 
`reactive.runner_manager`: Handles the logic of spawning scripts and keeping track of running processes. Also sets up logging accordingly.
 `scripts/reactive_runner`: This script is spawned by `reactive.runner_manager`. It sets up logging and calls the `reactive.runner` module.

### Library Changes

Add data_interfaces lib.

### Checklist

The documentation on charmhub will be updated when the reactive feature is in a usable state.


cbartz:
@jdkandersson or @gregory-schiano  Please provide initial feedback on this public interface

src-docs/mq.md
jdkandersson, line 1: Something to think about for this module - currently it is an abstraction for the message queue. You could consider making it an abstraction for a Job. So, I'm imagining some sort of interface which provides methods the charm would care about like:

* Get job, perhaps this blocks until a new job is available?
* Accept job
* Reject job

cbartz: @jdkandersson  Thanks. Have redesigned the module.

javierdelapuente (commented):
I had a first look, looks good. I added a few comments.

src/reactive/runner_manager.py
```diff
@@ -0,0 +1,178 @@
...
+        command,
+        shell=True,
+        env=env,
+        stdout=subprocess.DEVNULL,
+        stderr=subprocess.DEVNULL,
+        user=UBUNTU_USER,
+    )
+
+    try:
+        process.wait(0.001)
```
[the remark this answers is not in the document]
cbartz: No, we won't catch all errors here. The intention is to catch immediate exit errors.

scripts/reactive_runner.py
```diff
@@ -0,0 +1,46 @@
...
+
+"""Script to spawn a reactive runner."""
+import argparse
+import logging
+
+from reactive.runner import reactive_runner
+from reactive.runner_manager import REACTIVE_RUNNER_LOG_PATH
+
+
+def spawn_runner(mq_uri: str, queue_name: str) -> None:
```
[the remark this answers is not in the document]
cbartz: seems not, thanks

src/reactive/runner_manager.py
```diff
@@ -0,0 +1,178 @@
...
+    env = {"PYTHONPATH": "src:lib:venv"}
+    # We do not want to wait for the process to finish, so we do not use with statement.
+    # We trust the command.
+    command = " ".join(
+        [
+            TIMEOUT_COMMAND,
+            REACTIVE_RUNNER_TIMEOUT_STR,
+            PYTHON_BIN,
+            REACTIVE_RUNNER_SCRIPT_FILE,
+            f'"{reactive_runner_config.mq_uri}"',
```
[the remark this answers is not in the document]
cbartz: good idea, thanks

src/reactive/runner_manager.py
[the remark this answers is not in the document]
cbartz: I followed your suggestion and removed the wait. The delta is now determined after all new runners have been spawned. However, this means that in the case of `virtual-machines=1`, an immediate spawn failure will not be detected. @jdkandersson WDYT?

src/charm.py
```diff
@@ -125,6 +134,9 @@ def func_with_catch_errors(self: "GithubRunnerCharm", event: EventT) -> None:
             self.unit.status = BlockedStatus(
                 "Unauthorized OpenStack connection. Check credentials."
             )
+        except MissingIntegrationDataError as err:
```
yanksyoon, line 137: Rather than raise > catch > waitingstatus, would it be worth differentiating the status of the integration data by None?
What I've done for image-builder relation is that if the integration data is not yet ready, I'll return None values for the data fields. If the integration itself is missing, I'll return a None instead of the data struct.
The drawback of this implementation is that the type hints will include None in the fields:
```
class Dataclass:
    field: None | str
```
and it can become cumbersome doing None check for the fields each time even though it is guaranteed not to be None.
I'm thinking whether I should change the image relation implementation to raise like how you've done here.

src/charm_state.py
```diff
@@ -1015,6 +1070,8 @@ def _store_state(cls, state: "CharmState") -> None:
         # Convert pydantic object to python object serializable by json module.
         state_dict["proxy_config"] = json.loads(state_dict["proxy_config"].json())
         state_dict["charm_config"] = json.loads(state_dict["charm_config"].json())
+        if state.reactive_config:
+            state_dict["reactive_config"] = json.loads(state_dict["reactive_config"].json())
```
yanksyoon, line 1074: I'm wondering if this is an immutable config - if so, could you add this configuration to the [KeyError check](https://github.com/canonical/github-runner-operator/blob/597c1ba8a444ac77270fb29d4eedac71d6d11cfb/src/charm_state.py#L1083) in `def _check_immutable_config_change` please? (so that the upgrade charm does not throw an error)

src/reactive/runner.py
```diff
@@ -0,0 +1,32 @@
...
+
+"""Module which contains code to spawn a runner reactively."""
+import logging
+
+from reactive.job import Job, JobError, MessageQueueConnectionInfo
+
+logger = logging.getLogger(__name__)
+
+
+def reactive_runner(mq_uri: str, queue_name: str) -> None:
```
yanksyoon, line 12: Nit:
a little suggestion for naming :D
```suggestion
def spawn_reactive_runner(mq_uri: str, queue_name: str) -> None:
```

yanksyoon (commented):
Reviewed until `tests/integration/helpers/charm_metrics.py`.
Thanks for the review notes - really helps understand the PR better!

src/charm.py
```diff
@@ -68,14 +70,21 @@
...
+from reactive.runner_manager import REACTIVE_ERROR_LOGROTATE_CONFIG, REACTIVE_LOGROTATE_CONFIG
 from runner import LXD_PROFILE_YAML
 from runner_manager import RunnerManager, RunnerManagerConfig
 from runner_manager_type import FlushMode, OpenstackRunnerManagerConfig
 from utilities import bytes_with_unit_to_kib, execute_command, retry
 
 RECONCILE_RUNNERS_EVENT = "reconcile-runners"
 
+# This is currently hardcoded and may be moved to a config option in the future.
+REACTIVE_MQ_DB_NAME = "github-runner-webhook-router"
```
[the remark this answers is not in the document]
cbartz: The problem is that the paas-app charmer application hardcodes the db name to match the name of the github-runner-webhook-router charm application, which by default is `github-runner-webhook-router`.

src/logrotate.py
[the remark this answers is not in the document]
cbartz: Thanks for the comment, it also made me realise that logging is not multi-process safe: https://docs.python.org/3/howto/logging-cookbook.html#logging-to-a-single-file-from-multiple-processes
I changed the logging to output to stdout which is redirected to `/var/log/{pid}.log`. I think using pythons log rotation file handler makes no sense in this case, as each process will output to a different log file and logrotation will never be applied on another script execution.

scripts/reactive_runner.py
```diff
@@ -0,0 +1,44 @@
+#  Copyright 2024 Canonical Ltd.
```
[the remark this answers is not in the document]
cbartz: @yanksyoon  Can you elaborate on the second part of your comment? You mean having `reactive.runner.reactive_runner` in a script?

src/reactive/job.py
```diff
@@ -0,0 +1,192 @@
...
+        uri: The URI of the MQ.
+        queue_name: The name of the queue.
+    """
+
+    uri: AnyUrl
+    queue_name: str
+
+
+class JobSourceError(Exception):
+    """Raised when a job source error occurs."""
```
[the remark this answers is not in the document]
cbartz: I would prefer having the exceptions in the modules from now on. I think a module like `errors.py` is a [dependency magnet](https://github.com/jnguyen095/clean-code/issues/24). @jdkandersson @gregory-schiano WDYT?

src/charm_state.py
cbartz: Can I address this in an upcoming PR? At the moment the config is not immutable, but it will probably become so once we make spawning runners reactive.

src/charm.py
cbartz: I would prefer to have the exception control flow rather than handling `None` values.

scripts/reactive_runner.py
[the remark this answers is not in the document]
yanksyoon: In `openstack_manager.py` module, the `def _create_runner` [method](https://github.com/canonical/github-runner-operator/blob/main/src/openstack_cloud/openstack_manager.py#L613) is defined as a static method. I was wondering if this would also help if it was put here as a script as well :) (as to match the style in which we spawn sub processes)

src/charm.py
yanksyoon: I will modify the image-builder relation to do the same, thank you!

src/charm_state.py
yanksyoon: Perfectly fine with me :D thanks!

src/reactive/job.py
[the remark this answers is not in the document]
yanksyoon: TIL!

src/reactive/runner_manager.py
```diff
@@ -0,0 +1,167 @@
...
+    command = " ".join(
+        [
+            TIMEOUT_COMMAND,
+            REACTIVE_RUNNER_TIMEOUT_STR,
+            PYTHON_BIN,
+            REACTIVE_RUNNER_SCRIPT_FILE,
+            f'"{reactive_runner_config.queue_name}"',
+            ">>",
+            # $$ will be replaced by the PID of the process, so we can track the error log easily.
+            f"{REACTIVE_RUNNER_LOG_DIR}/$$.log",
```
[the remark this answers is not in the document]
weiiwang01: But it might take a long time for PID to wrap around.

tests/integration/test_reactive.py
```diff
@@ -0,0 +1,131 @@
...
+
+FAKE_URL = "http://example.com"
+
+
+@pytest.mark.openstack
+async def test_reactive_mode_consumes_jobs(ops_test: OpsTest, app_for_reactive: Application):
+    unit = app_for_reactive.units[0]
+    mongodb_uri = await _get_mongodb_uri_from_integration_data(ops_test, unit)
+    if not mongodb_uri:
+        mongodb_uri = await _get_mongodb_uri_from_secrets(ops_test, unit.model)
```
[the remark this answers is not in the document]
weiiwang01: I think it will be kept secret if it's Juju 3, or it will be related to data if it's Juju 2.

https://github.com/canonical/mongodb-operator/blob/b33d036173f47c68823e08a9f03189dc534d38dc/lib/charms/data_platform_libs/v0/data_interfaces.py#L827-L831

scripts/reactive_runner.py
```diff
@@ -0,0 +1,46 @@
...
+    argument_opts: dict = {"default": mq_uri} if mq_uri else {"required": True}
+    parser.add_argument(
+        "--mq-uri",
+        help="URI of the message queue database. This should include authentication information."
+        f"The argument is required but can also be set via the {MQ_URI_ENV_VAR} "
+        "environment variable."
+        " Example URI : mongodb://user:pw@10.36.24.114/github-runner-webhook-router"
+        "?replicaSet=mongodb&authSource=admin",
+        **argument_opts,
+    )
```
weiiwang01, line 42: If the URI is included in the command line arguments, everyone on the same system can see the URI, including the username and password.

src/metrics/events.py
```diff
@@ -8,16 +8,22 @@
...
 METRICS_LOG_PATH = Path("/var/log/github-runner-metrics.log")
-
+# Set rotate param to 0 to not keep the old metrics log file to avoid sending the
+# metrics to Loki twice, which may happen if there is a corrupt log scrape configuration.
+METRICS_LOGROTATE_CONFIG = LogrotateConfig(
+    name="github-runner-metrics",
+    log_path_glob_pattern=str(METRICS_LOG_PATH),
+    rotate=0,
+    create=False,
+)
```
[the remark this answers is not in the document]
cbartz: Yes, this is true, the comment indicates that this should keep us safe in case of external errors (corrupt log scrape configuration).

src/reactive/runner_manager.py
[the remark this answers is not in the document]
cbartz: I think it is easier for troubleshooting to have different log files per process ID.

scripts/reactive_runner.py
cbartz: Yes, this is true, but it is optional to run the script in this way. You prefer to remove it? We call the script using an env var.

src/reactive/runner_manager.py
```diff
@@ -0,0 +1,167 @@
...
+    Raises:
+        _SpawnError: If the runner fails to spawn.
+    """
+    env = {
+        "PYTHONPATH": "src:lib:venv",
+        MQ_URI_ENV_VAR: reactive_runner_config.mq_uri,
+    }
+    # We do not want to wait for the process to finish, so we do not use with statement.
+    # We trust the command.
+    command = " ".join(
```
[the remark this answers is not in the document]
cbartz: As mentioned via DM, I don't think we benefit from shlex.join here, as it would also quote `>>` and `2>&1` here.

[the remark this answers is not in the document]
weiiwang01: Yes, thanks for testing it!

scripts/reactive_runner.py
weiiwang01: Depends on you :)

weiiwang01 (dismissed):

src/openstack_cloud/openstack_manager.py
```diff
@@ -345,6 +350,22 @@ def reconcile(self, quantity: int) -> int:
...
+
+        Returns:
+            The difference between intended runners and actual runners. In reactive mode
+            this number is never negative as additional processes should terminate after a timeout.
+        """
+        logger.info("Reactive mode is experimental and not yet fully implemented.")
+        config = reactive_runner_manager.ReactiveRunnerConfig(
+            mq_uri=self._config.reactive_config.mq_uri, queue_name=self.app_name
+        )
+        return reactive_runner_manager.reconcile(quantity=quantity, config=config)
```
jdkandersson, line 367: Think about whether the config is needed or it can be passed as arguments directly

src-docs/reactive.runner_manager.md
```diff
@@ -0,0 +1,85 @@
...
+# <kbd>module</kbd> `reactive.runner_manager`
+Module for managing reactive runners. 
+
+**Global Variables**
+---------------
+- **MQ_URI_ENV_VAR**
+- **REACTIVE_RUNNER_SCRIPT_FILE**
+- **REACTIVE_RUNNER_TIMEOUT_STR**
+- **PYTHON_BIN**
+- **PS_COMMAND_LINE_LIST**
```
jdkandersson, line 14: ACTIVE_SCRIPTS_LIST_COMMAND

src-docs/reactive.runner_manager.md
```diff
@@ -0,0 +1,85 @@
...
+- **TIMEOUT_COMMAND**
```
jdkandersson, line 15: This seems inconsistent with the above naming convention for the Python binary

src-docs/reactive.runner_manager.md
```diff
@@ -0,0 +1,85 @@
...
+ 
+ - <b>`quantity`</b>:  The number of runners to spawn. 
+ - <b>`config`</b>:  The configuration for the reactive runner. 
+
+Raises a ReactiveRunnerError if the runner fails to spawn. 
+
+
+
+**Returns:**
+ The number of runners spawned.
```
jdkandersson, line 42: Scripts?

src/reactive/runner_manager.py
```diff
@@ -0,0 +1,156 @@
...
+    Args:
+        quantity: The number of runners to spawn.
+        config: The configuration for the reactive runner.
+
+    Raises a ReactiveRunnerError if the runner fails to spawn.
+
+    Returns:
+        The number of runners spawned.
+    """
+    actual_quantity = _determine_current_quantity()
```
jdkandersson, line 56: Perhaps line up the name and change to `get` instead of `determine`

src/reactive/runner_manager.py
```diff
@@ -0,0 +1,156 @@
...
+    logger.info("Actual quantity of reactive runner processes: %s", actual_quantity)
+    actual_delta = delta = quantity - actual_quantity
+    if delta > 0:
+        logger.info("Will spawn %d new reactive runner processes", delta)
+        _setup_logging()
```
jdkandersson, line 61: Perhaps `_setup_logging_for_processes`

src/reactive/runner_manager.py
```diff
@@ -0,0 +1,156 @@
...
+        for _ in range(delta):
+            try:
+                _spawn_runner(config)
+            except _SpawnError:
+                logger.exception("Failed to spawn a new reactive runner process")
+        actual_quantity_after_spawning = _determine_current_quantity()
+        actual_delta = actual_quantity_after_spawning - actual_quantity
```
jdkandersson, line 68: Up to you, perhaps just return `delta`

src/reactive/runner_manager.py
```diff
@@ -0,0 +1,156 @@
...
+        The number of reactive runners.
+
+    Raises:
+        ReactiveRunnerError: If the number of reactive runners cannot be determined
+    """
+    result = secure_run_subprocess(cmd=PS_COMMAND_LINE_LIST)
+    if result.returncode != 0:
+        raise ReactiveRunnerError("Failed to get list of processes")
+    commands = result.stdout.decode().rstrip().split("\n")[1:] if result.stdout else []
+    actual_quantity = 0
```
jdkandersson, line 95: `actual quantity = sum(1 for command in commands if commane.startswith...)

src/reactive/runner_manager.py
```diff
@@ -0,0 +1,156 @@
...
+def _determine_current_quantity() -> int:
+    """Determine the current quantity of reactive runners.
+
+    Returns:
+        The number of reactive runners.
+
+    Raises:
+        ReactiveRunnerError: If the number of reactive runners cannot be determined
+    """
+    result = secure_run_subprocess(cmd=PS_COMMAND_LINE_LIST)
```
jdkandersson, line 91: Potentially change this command to not include the header

src/reactive/runner_manager.py
```diff
@@ -0,0 +1,156 @@
...
+    if result.returncode != 0:
+        raise ReactiveRunnerError("Failed to get list of processes")
+    commands = result.stdout.decode().rstrip().split("\n")[1:] if result.stdout else []
```
jdkandersson, line 94: Potentially remove the `rstrip` since that line would be skipped

src/reactive/runner_manager.py
cbartz: I removed the `actual_quantity_after_spawning` and now just count the number of spawned runners (if there is an exception, I increment the counter) and remove that number.

src/reactive/job.py
jdkandersson, line 1: Algorithm:

1. Call a blocking function to get a message
2. Look at the info inside the message to check if it can be handled, if not reject it and send it back
3. try:
  a. Handle the message
  b. Ack the message
6. except: send it back

src/reactive/runner_manager.py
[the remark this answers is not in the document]
cbartz: We agreed not to consider the return code of the process.

jdkandersson (dismissed):

jdkandersson (approved):

weiiwang01 (approved):

scripts/reactive_runner.py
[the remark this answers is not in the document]
cbartz: We can discuss this further internally. I am resolving this conversation now.

PR #312, opened 2024-07-03
feat: image builder relation

yanksyoon:
Applicable spec: N/A

### Overview

- Adds integration w/ github-runner-image-builder
- Removes logic involving image building for Openstack.

### Rationale

- We now have a new OpenStack Image bulider charm!

### Juju Events Changes

- All events that involve openstack (except flush) now require image relation to be ready.
- The flush method is an exception since you should be able to flush runners regardless of image relation.

### Module Changes

- `charm.py`: event handlers now check for image availability before reconciling.
- `openstack_manager.py`: removes logic involved in building image, adds a passthrough of image_id from relation data.

### Library Changes

None

### Checklist


jdkandersson (dismissed):

yhaliaw (dismissed):

docs/how-to/openstack-runner.md
```diff
@@ -8,6 +8,19 @@ enabled the charm cannot be changed to use other virtualization methods.
...
+The OpenStack runners require image to be available via the `image` integration. Deploy the
+`github-runner-image-builder` charm and wait for the image to be successfully provided via the
+relation data.
+
+```bash
+juju deploy github-runner-image-builder
+juju integrate github-runner-image-builder github-runner
+```
+
+The image takes around 10-15 minutes to be built and ready via the relation.
```
cbartz, line 22:
```suggestion
The image will take about 10-15 minutes to build and be ready via the relation.
```

src/charm.py
```diff
@@ -504,6 +497,21 @@ def _on_start(self, _: StartEvent) -> None:
 
         self.unit.status = ActiveStatus()
 
+    def _get_set_image_ready_status(self) -> bool:
```
cbartz, line 500: nitpick: Should be placed below to comply with step-down rule

src/charm.py
```diff
@@ -1122,6 +1140,23 @@ def _on_debug_ssh_relation_changed(self, _: ops.RelationChangedEvent) -> None:
             state.runner_config.virtual_machine_resources,
         )
 
+    @catch_charm_errors
+    def _on_image_relation_changed(self, _: ops.RelationChangedEvent) -> None:
+        """Handle debug ssh relation changed event."""
```
cbartz, line 1145:
```suggestion
        """Handle image relation changed event."""
```

src/charm.py
```diff
@@ -1122,6 +1140,23 @@ def _on_debug_ssh_relation_changed(self, _: ops.RelationChangedEvent) -> None:
...
+        state = self._setup_state()
+
+        if state.instance_type != InstanceType.OPENSTACK:
+            return
```
cbartz, line 1149: Should we go into blocked state to inform the user that they cannot use this relationship with lxd?

cbartz (commented):
Great! Left some nitpicks.

src/charm.py
yanksyoon: I think it's okay - the LXD runners should build images itself. Going it to BlockedState suggests that the relation should be removed for the runners to work.
How about a warning log?

cbartz: When the user relates to the image relation, they have a certain expectation and that expectation is not met with LXD. 
I think BlockedState might be appropriate as it is very visible in the juju status output. An alternative would be to set a status message for the Active state.

The logs are usually not read if everything seems to be working fine.

I think BlockedState could also be used if the core functionality is still working, https://juju.is/docs/sdk/status#heading--possible-statuses leaves room for interpretation:

>Blocked: the device requires manual intervention by the Juju user, as specified by the charm author (see status notes or charm docs).

@gregory-schiano what do you think should happen in such cases?

yanksyoon: Perhaps a message on the current status would be fit.
```
StatusBase.from_name(self.unit.status.name, "LXD runners do not require image relation to be set.")
```
What do you think?

cbartz (dismissed):
There is one pending conversation about going into BlockedState, feel free to resolve.

src/charm.py
gregory-schiano: So the workload will still work properly but the relation is "useless"
I think a message in the active status is fine, but we should make sure the message is persistent so probably use a collect status event and put the logic there.
Beware that using collect status can override any status set before, so it has consequences on how you manages it

docs/how-to/openstack-runner.md
```diff
@@ -8,6 +8,19 @@ enabled the charm cannot be changed to use other virtualization methods.
 
 There are three configuration that the charm needs to be deployed with to enable OpenStack integration: `openstack-clouds-yaml`, `openstack-flavor`, and `openstack-network`.
 
+## Integration
+
+The OpenStack runners require image to be available via the `image` integration. Deploy the
```
erinecon, line 13: Nitpick: Add "the", as in, "The OpenStack runners require *the* image to be available..."

docs/how-to/openstack-runner.md
```diff
@@ -8,6 +8,19 @@ enabled the charm cannot be changed to use other virtualization methods.
...
+`github-runner-image-builder` charm and wait for the image to be successfully provided via the
+relation data.
+
+```bash
+juju deploy github-runner-image-builder
+juju integrate github-runner-image-builder github-runner
+```
+
+The image will take about 10-15 minutes to build and be ready via the relation.
```
erinecon, line 22: Assuming this is technically correct, I think it would be more concise to say "The image will take about 10-15 minutes to build and be fully integrated." 

You also might want to tell users how to check that the deployment/integration has finished, I assume with `juju status`

src/charm.py
yanksyoon: Will add collect status on a follow up PR, since it adds overhead of a review.

src/charm.py
```diff
@@ -1122,6 +1125,38 @@ def _on_debug_ssh_relation_changed(self, _: ops.RelationChangedEvent) -> None:
...
+        """Handle image relation changed event."""
+        state = self._setup_state()
+
+        if state.instance_type != InstanceType.OPENSTACK:
+            return
+        if not self._get_set_image_ready_status():
+            return
+
+        runner_manager = self._get_openstack_runner_manager(state)
+        # 2024/04/12: Should be flush idle.
```
nrobinaubertin, line 1139: Nitpick but maybe this comment should be prefixed with `TODO:` (and the date be ISO `2024-04-12`)

nrobinaubertin (approved):
One nitpick, otherwise LGTM

src/charm.py
yanksyoon: I'll also fix this as a follow up PR since the tests are pretty hard to make pass. Thank you for the review!

requirements.txt
```diff
@@ -3,12 +3,12 @@ jinja2
 fabric >=3,<4
 openstacksdk>=3,<4
 ops>=2.8
-pylxd @ git+https://github.com/canonical/pylxd
+pylxd @ git+https://github.com/canonical/pylxd@46b58e61a465a970937cd97bbbf93622f98caa8c
```
yhaliaw, line 6: Would it be good to comment why the version is pinned?

yhaliaw (approved):
One minor comment.
Can be done in a follow up PR.

PR #317, opened 2024-07-10
Grafana Dashboard: Combine 3 queries into one

cbartz:
Applicable spec: n/a

### Overview

Combine 3 queries into one using aggregation for the `Lifecycle Status` panel. The visualisation of the dashboard should not have changed:

Screenshot from 2024-07-10 16-12-58

### Rationale

Reduce the load on Loki by making fewer requests.

### Juju Events Changes

n/a

### Module Changes

n/a

### Library Changes

n/a

### Checklist

### DIFF
--- a/src/grafana_dashboards/metrics.json
+++ b/src/grafana_dashboards/metrics.json
@@ -24,7 +24,7 @@
   "editable": true,
   "fiscalYearStartMonth": 0,
   "graphTooltip": 0,
-  "id": 266,
+  "id": 293,
   "links": [],
   "liveNow": false,
   "panels": [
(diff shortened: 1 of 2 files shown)


javierdelapuente (dismissed):

merkata (approved):
LGTM

src/grafana_dashboards/metrics.json
```diff
@@ -1575,8 +1577,8 @@
       "type": "row"
     }
   ],
-  "refresh": false,
-  "schemaVersion": 37,
+  "refresh": "",
```
yhaliaw, line 1580: Is this intended?
The value was `false` and now `""`, which are different data types.

yanksyoon (approved):

src/grafana_dashboards/metrics.json
cbartz: I exported the json directly from Grafana, it seems that setting false or empty string has the same intention (no live refresh): https://github.com/grafana/grafonnet-lib/issues/45 . Maybe when this panel was first created some time ago, a different version of Grafana was used for exporting, which used false by default. I do not expect this code change to have any effect.

PR #324, opened 2024-07-17
Remove residual empty directories from last revision

weiiwang01:
Juju sometimes leaves empty directories from the previous revision uncleaned during charm upgrades. This issue is tracked at [https://bugs.launchpad.net/juju/+bug/2058335](https://bugs.launchpad.net/juju/+bug/2058335). Until the fix is integrated into Juju, we need to update the patch we applied to this charm to mitigate this problem for now.

### DIFF
--- a/src/openstack_cloud/__init__.py
+++ b/src/openstack_cloud/__init__.py
@@ -5,7 +5,6 @@
 
 import logging
 import os
-import shutil
 from pathlib import Path
 from typing import TypedDict, cast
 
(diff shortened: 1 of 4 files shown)


yanksyoon (dismissed):
Thank you!

src/openstack_cloud/__init__.py
```diff
@@ -19,31 +18,15 @@
...
-            "Please reach out to the charm dev team for further advice."
-        )
+def _remove_residual_venv_dirs() -> None:  # pragma: no cover
+    """Remove the residual empty directories from last revision if it exists."""
+    unit_name = os.environ["JUJU_UNIT_NAME"].replace("/", "-")
+    venv_dir = Path(f"/var/lib/juju/agents/unit-{unit_name}/charm/venv/")
+    for path in venv_dir.iterdir():
+        if path.is_dir() and not os.listdir(path):
+            logger.warning("Removing residual empty dir: %s", path)
+            path.rmdir()
```
cbartz, line 29: The code is no longer specifically targeting an openstack lib. Can you move this code somewhere else and make sure it is called at the top of the charm.py module? Otherwise it looks a bit odd to a reader who does not have the internal knowledge of why we are doing this.

weiiwang01: This operation has a performance penalty as it walks the entire `venv` directory, which can be quite large. Therefore, I think it might be best to call this function only when necessary.

weiiwang01: Okay, I've tested this in the real environment, and the entire process takes less than a second to finish, so I guess it's okay to run this every time a hook is fired. But please make sure it's removed once this problem is fixed in Juju. Thanks!

cbartz (approved):

yanksyoon (approved):

PR #329, opened 2024-07-23
chore: image bulider match style

yanksyoon:
Applicable spec: N/A

### Overview

From previous discussion, a charm state should raise an error if not valid and be caught, rather than returning validation result. This PR follows the pattern discussed.

### Rationale

To have consistent code styles.

### Juju Events Changes

None.

### Module Changes

- Removed `_get_set_image_ready_status` which sets charm states if image relation is not ready/ not provided.
- The charm now falls into BlockedStatus/WaitingStatus if the image relation is not provided regardless of whether the openstack charm deployment requires the image to be ready (e.g. flush doesn't require image to be ready)

### Library Changes

None.

### Checklist

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -58,7 +58,6 @@ from charm_state import (
     CharmState,
     GithubPath,
     InstanceType,
-    OpenstackImage,
     ProxyConfig,
     RunnerStorage,
     VirtualMachineResources,
@@ -66,6 +65,8 @@ from charm_state import (
 )
 from errors import (
     ConfigurationError,
+    IntegrationDataNotReadyError,
+    IntegrationNotFoundError,
     LogrotateSetupError,
     MissingMongoDBError,
     MissingRunnerBinaryError,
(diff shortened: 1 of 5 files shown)


src/charm.py
```diff
@@ -141,6 +142,12 @@ def func_with_catch_errors(self: "GithubRunnerCharm", event: EventT) -> None:
         except MissingMongoDBError as err:
             logger.exception("Missing integration data")
             self.unit.status = WaitingStatus(str(err))
+        except IntegrationNotFoundError as err:
+            logger.exception("Missing integration")
+            self.unit.status = BlockedStatus(str(err))
+        except IntegrationDataNotReadyError as err:
+            logger.exception("Missing integration data")
```
arturo-seijas, line 149: This is the same msg as in line 143. Is this intentional?

cbartz: I think @gregory-schiano did not like the abstract name (I also originally had "MissingIntegrationDataError" but then changed it to `MissingMongoDBError`). I am fine with using the abstract name and merging this with lines 142-144, but perhaps confirm with @gregory-schiano .

javierdelapuente:
@yanksyoon  do you think we can close this PR? it is a bit old and has several conflicts.

PR #332, opened 2024-07-23
Reactive: Add sigterm handler and process termination. 

cbartz:

### Overview

Follow-up to https://github.com/canonical/github-runner-operator/pull/305 :

- Add sigterm handler which lets kombu requeue consumed messages that have not been acknowledged.
- Remove Timeout command but instead kill unnecessary processes.
- Remove logrotation config for runner logs (has been forgotten in previous PR).

### Rationale

This is in preparation for the upcoming PR, which will spawn runners reactively.

### Juju Events Changes

n/a

### Module Changes

- `reactive/consumer.py`: Add sigterm handler .
- `reactive/runner_manager.py`: Add logic to kill processes.
- `logrotate.py`: Remove logrotate configuration for runner logs.

### Checklist

### DIFF
--- a/src/reactive/consumer.py
+++ b/src/reactive/consumer.py
@@ -47,7 +47,7 @@ def consume(mongodb_uri: str, queue_name: str) -> None:
     """
     with Connection(mongodb_uri) as conn:
         with closing(SimpleQueue(conn, queue_name)) as simple_queue:
-            _add_sigterm_handler(simple_queue, signal.SIGTERM)
+            _add_sigterm_handler(signal.SIGTERM)
             msg = simple_queue.get(block=True)
             try:
                 job_details = cast(JobDetails, JobDetails.parse_raw(msg.payload))
@@ -62,24 +62,25 @@ def consume(mongodb_uri: str, queue_name: str) -> None:
             msg.ack()
 
 
-def _add_sigterm_handler(queue: SimpleQueue, signal_code: signal.Signals) -> None:
-    """Add a signal handler to clean up and exit.
+def _add_sigterm_handler(signal_code: signal.Signals) -> None:
+    """Add a signal handler for clean up and exit.
 
     Args:
-        queue: The queue to clean up.
         signal_code: The signal code to handle.
     """
 
     def sigterm_handler(signal_code: int, _: FrameType | None) -> None:
         """Handle a signal.
 
-        Requeue the message and exit.
+        Call sys.exit with the signal code. Kombu should automatically
+        requeue unacknowledged messages.
 
         Args:
             signal_code: The signal code to handle.
         """
-        logger.error("Signal '%s' received. Requeueing message.", signal.strsignal(signal_code))
-        queue.channel.basic_recover(requeue=True)
+        print(
+            f"Signal '{signal.strsignal(signal_code)}' received. Will terminate.", file=sys.stderr
+        )
         sys.exit(signal_code)
 
     signal.signal(signal_code, sigterm_handler)
--- a/src/reactive/consumer.py
+++ b/src/reactive/consumer.py
@@ -2,12 +2,13 @@
 #  See LICENSE file for licensing details.
 
 """Module responsible for consuming jobs from the message queue."""
+import contextlib
 import logging
 import signal
 import sys
 from contextlib import closing
 from types import FrameType
-from typing import cast
+from typing import Generator, cast
 
 from kombu import Connection
 from kombu.simple import SimpleQueue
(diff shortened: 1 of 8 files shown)


src/reactive/runner_manager.py
```diff
@@ -41,45 +49,52 @@ def reconcile(quantity: int, mq_uri: str, queue_name: str) -> int:
...
         for _ in range(delta):
             _spawn_runner(mq_uri=mq_uri, queue_name=queue_name)
     elif delta < 0:
-        logger.info(
-            "%d reactive runner processes are running. "
-            "Will skip spawning. Additional processes should terminate after %s.",
-            current_quantity,
-            REACTIVE_RUNNER_TIMEOUT_INTERVAL,
-        )
+        logger.info("Will kill %d process(es).", -delta)
```
amandahla, line 62: What causes this difference? Are runners running too long?

cbartz: Yes, it is not implemented yet, but the upcoming PR will spawn runners in the subprocess and wait until a job has been picked up. 
The user can set the number of runner processes per config option - so it could happen that there are more processes running in the background than expected and we have to downsclae.
Initially a timeout approach was suggested, but after internal discussions with @jdkandersson we decided to kill superfluous processes.

src/reactive/consumer.py
```diff
@@ -56,3 +60,26 @@ def consume(mongodb_uri: str, queue_name: str) -> None:
...
+    def sigterm_handler(signal_code: int, _: FrameType | None) -> None:
+        """Handle a signal.
+
+        Requeue the message and exit.
+
+        Args:
+            signal_code: The signal code to handle.
+        """
+        logger.error("Signal '%s' received. Requeueing message.", signal.strsignal(signal_code))
+        queue.channel.basic_recover(requeue=True)
```
yanksyoon, line 82: Would this have some sort of a timeout?

yanksyoon (dismissed):
LGTM

src/reactive/consumer.py
```diff
@@ -56,3 +60,26 @@ def consume(mongodb_uri: str, queue_name: str) -> None:
...
+        sys.exit(signal_code)
```
weiiwang01, line 83: If I remember correctly, the Python signal handling function can execute at any point during the execution of python bytecode. This means there's no guarantee that the internal state of `queue` even `logger` will remain consistent entering the single handling function, potentially leading to unexpected behavior.

cbartz: https://docs.python.org/3.10/library/signal.html#execution-of-python-signal-handlers indicates that the handler function is not called during bytecode execution (but "after" the next bytecode instruction). So I would not expect to have an inconsistent state for the `queue' or `logger' object.

src/reactive/consumer.py
cbartz: Unfortunately there is no option to pass a timeout, but digging into the lib it seems that pymongo is being used, which seems to use some sort of timeout by default (just skimmed over it, so its more of a guess).

src/reactive/consumer.py
weiiwang01: Bytecode is a very small unit of execution, which means that the event handler can be called during the execution of an internal method in queue or logger, and the internal method can be in the process of manipulating a series of internal states. This is the same reason why catching `KeyboardInterrupt` is usually not a good idea. https://docs.python.org/3/library/exceptions.html#KeyboardInterrupt

yhaliaw (dismissed):

src/reactive/consumer.py
cbartz: You are right.
I changed the code to just exit. It seems that kombu will requeue the message in that case (tested it locally), due to https://github.com/celery/kombu/blob/46af1b94e061e098e0471efdffac46c1e8c567b0/kombu/transport/virtual/base.py#L274 and https://github.com/celery/kombu/blob/46af1b94e061e098e0471efdffac46c1e8c567b0/kombu/transport/virtual/base.py#L194

weiiwang01: By the way, is signal handling necessary only for the `consume` function? If so, should the signal handler be canceled before the function returns?

yanksyoon (approved):

weiiwang01 (dismissed):

src/reactive/consumer.py
cbartz: Have updated the code to restore the default handler.

yanksyoon (dismissed):

yanksyoon (approved):

src/reactive/consumer.py
```diff
@@ -44,15 +48,65 @@ def consume(mongodb_uri: str, queue_name: str) -> None:
...
+                    job_details = cast(JobDetails, JobDetails.parse_raw(msg.payload))
+                except ValidationError as exc:
+                    msg.reject(requeue=True)
+                    raise JobError(f"Invalid job details: {msg.payload}") from exc
+                logger.info(
+                    "Received job with labels %s and run_url %s",
+                    job_details.labels,
+                    job_details.run_url,
+                )
+                msg.ack()
```
weiiwang01, line 63: I think the message will only be rolled back on exit if it's not `ack`? Maybe this line, `msg.ack()`, needs to be placed outside of the `with` block?

cbartz: If we put it outside the block and we have an execution error in `msg.ack()`, this would mean that the message would not be recovered. So I think it should stay inside the block.

weiiwang01 (approved):

src/reactive/runner_manager.py
cbartz: I am resolving this conversation, feel free to reopen.

PR #335, opened 2024-07-24
Add security group rules by ID

yhaliaw:
### Overview

1. Add ID of security group to logging.
2. Add security rules by ID of security group.

### Rationale

1. More info for debugging.
2. Causes an exception when:
  - Multiple security groups exists
  - The security groups does not have all the rules setup.
 
### Checklist

### DIFF
--- a/src/openstack_cloud/openstack_manager.py
+++ b/src/openstack_cloud/openstack_manager.py
@@ -794,7 +794,7 @@ class OpenstackRunnerManager:
             for rule in existing_rules:
                 if rule["protocol"] == "icmp":
                     logger.debug(
-                        "Found ICMP rule in existing security group %s", SECURITY_GROUP_NAME
+                        "Found ICMP rule in existing security group %s of ID %s", SECURITY_GROUP_NAME, existing_security_group.id
                     )
                     rule_exists_icmp = True
                 if (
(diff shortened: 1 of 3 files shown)


src/openstack_cloud/openstack_manager.py
```diff
@@ -794,51 +794,57 @@ def _ensure_security_group(conn: OpenstackConnection) -> None:
...
-        if not rule_exists_tmate_ssh:
-            conn.create_security_group_rule(
-                secgroup_name_or_id=SECURITY_GROUP_NAME,
-                port_range_min="10022",
-                port_range_max="10022",
-                protocol="tcp",
-                direction="egress",
-                ethertype="IPv4",
-            )
+            if not rule_exists_icmp:
```
yhaliaw, line 823: The indentation is logically not needed, but the type checker is not smart enough to figure that out.

Reasoning: 
`rule_exists_icmp` and the others can only be `True` if `existing_security_group` is not `None`, but the  type checker does not know this.

cbartz: But what happens if `existing_security_group` is `None`? It will be created in line 788, but due to the change in logic and indentation, a newly created security group will not have the security rules created below. This seems wrong.

yhaliaw: Updated.
Can you check if this fix your concern?

cbartz: I think when the group is created in line 788, we need to set `existing_security_group` to the newly created group, otherwise no rules will be created for this newly created group.

yhaliaw: You are right. The reason that the typing system is throwing the error is due to this.
I have updated the code and the type checker is not complaining anymore.

yanksyoon (approved):

weiiwang01 (approved):

PR #344, opened 2024-08-01
feat: openstack flush mode

yanksyoon:
Applicable spec: <link>

### Overview

- Add flushmode support for Openstack.
- Add timeout to openstack runners being stuck in BUILD status for +10 min.

Flush logic:
1. Kill runner processes
  - IDLE: Dispatch ssh exec command to kill `Runner.Listener` process if `Runner.Worker` is not found.
  - BUSY: Dispatch ssh exec command to kill `Runner.Listener` and `Runner.Worker`.
2. Health check - since runner processes do not exist, health check should fail for that ones that were killed.
3. Cleanup - regular cleanup process & metrics extraction.

### Rationale

- Imagebuilder update kept flushing runners to death.

### Module Changes

- `openstack_manager.py` now supports two flush modes: `FlushIdle` and `FlushBusy`
- Other flush modes are unsupported due to not enough use-case. Will propose to remove `WAIT_FOR_REPO_CHECK` type flush modes.

### Checklist

### DIFF
--- a/src/openstack_cloud/openstack_manager.py
+++ b/src/openstack_cloud/openstack_manager.py
@@ -480,7 +480,7 @@ class OpenstackRunnerManager:
             conn: The Openstack connection instance.
             server_name: The name of the OpenStack server to health check.
             startup: Check only whether the startup is successful.
-            building_timeout_mins: How long to wait for BULIDING status until it is deemed
+            building_timeout_mins: How long to wait for BUILDING status until it is deemed
                 unhealthy.
 
         Returns:
@@ -1516,12 +1516,15 @@ class OpenstackRunnerManager:
         2. Get unhealthy runners after process purging.
         3. Delete unhealthy runners.
 
+        Args:
+            mode: The mode to determine which runner to flush.
+
         Returns:
             The number of runners flushed.
         """
         logger.info("Flushing OpenStack all runners")
         with _create_connection(self._cloud_config) as conn:
-            self._kill_runner_processes(conn=conn, mmode=mode)
+            self._kill_runner_processes(conn=conn, mode=mode)
             runner_by_health = self._get_openstack_runner_status(conn)
             remove_token = self._github.get_runner_remove_token(path=self._config.path)
             self._remove_runners(
@@ -1531,7 +1534,7 @@ class OpenstackRunnerManager:
             )
         return len(runner_by_health.unhealthy)
 
-    def _kill_runner_processes(self, conn: OpenstackConnection, mode: FlushMode):
+    def _kill_runner_processes(self, conn: OpenstackConnection, mode: FlushMode) -> None:
         """Kill runner application that are not running any jobs.
 
         Runners that have not picked up a job has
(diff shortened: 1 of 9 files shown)


src/openstack_cloud/openstack_manager.py
```diff
@@ -487,6 +494,11 @@ def _health_check(conn: OpenstackConnection, server_name: str, startup: bool = F
             return False
         if server.status not in (_INSTANCE_STATUS_ACTIVE, _INSTANCE_STATUS_BUILDING):
             return False
+        created_at = datetime.strptime(server.created_at, "%Y-%m-%dT%H:%M:%SZ")
+        current_time = datetime.now(created_at.tzinfo)
+        elapsed_min = (created_at - current_time).total_seconds() / 60
```
jdkandersson, line 499: What is the current timeout?

src/openstack_cloud/openstack_manager.py
```diff
@@ -1509,20 +1520,71 @@ def _issue_reconciliation_metric(
         except IssueMetricEventError:
             logger.exception("Failed to issue Reconciliation metric")
 
-    def flush(self) -> int:
+    def flush(self, mode: FlushMode = FlushMode.FLUSH_IDLE) -> int:
         """Flush Openstack servers.
 
+        1. Kill the processes depending on flush mode.
+        2. Get unhealthy runners after process purging.
+        3. Delete unhealthy runners.
```
jdkandersson, line 1528: Why not just delete the machine in OpenStack / LXD?

yanksyoon: There are follow-up processes such as metrics collection, key deletion. To reduce the responsibility of the function, the flush method will only handle deletion of the processes and let the normal flow continue.

src/openstack_cloud/openstack_manager.py
yanksyoon: 5 minutes is the current timeout. Good catch, there already is a constant for it.

jdkandersson (dismissed):

cbartz (dismissed):
Looks good, minor comment.

src/openstack_cloud/openstack_manager.py
```diff
@@ -487,6 +491,11 @@ def _health_check(conn: OpenstackConnection, server_name: str, startup: bool = F
             return False
         if server.status not in (_INSTANCE_STATUS_ACTIVE, _INSTANCE_STATUS_BUILDING):
             return False
+        created_at = datetime.strptime(server.created_at, "%Y-%m-%dT%H:%M:%SZ")
+        current_time = datetime.now(created_at.tzinfo)
+        elapsed_min = (created_at - current_time).total_seconds()
+        if server.status == _INSTANCE_STATUS_BUILDING:
+            return elapsed_min < CREATE_SERVER_TIMEOUT
```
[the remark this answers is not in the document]
yanksyoon: Added, thanks!

cbartz (dismissed):

jdkandersson (approved):

src/openstack_cloud/openstack_manager.py
```diff
@@ -1509,20 +1518,71 @@ def _issue_reconciliation_metric(
...
             self._remove_runners(
                 conn=conn,
-                instance_names=runners_to_delete,
+                instance_names=runner_by_health.unhealthy,
                 remove_token=remove_token,
             )
-            return len(runners_to_delete)
+        return len(runner_by_health.unhealthy)
+
+    def _kill_runner_processes(self, conn: OpenstackConnection, mode: FlushMode) -> None:
```
yhaliaw, line 1546: Is this needed?
Would the runner remove command that takes the remove token do this already?

yhaliaw (approved):
Question on whether an kill process is needed.
This should not be an issue, as the commend will be just be redundant/safeguard if not really needed.

src/openstack_cloud/openstack_manager.py
yanksyoon: Would you be referring to the `self._remove_runners` command?

The reason for having this is because from the moment where we perform health-checks via ssh until we actually delete the runners have quite a bit of a time lag in which the jobs can be picked up by the runners (since the deletion of runners are sequential and if OpenStack decides to take time to perform the action, the bigger the discrepancy from the truth).

This function reduces the racy implementation of health-check -> deletion by determining on spot whether the runner process should be stopped (thereby no longer receiving jobs), and then cleaning up via health check afterwards.

PR #346, opened 2024-08-06
test: add amd64/arm64 testing options

yanksyoon:
Applicable spec: N/A

### Overview

- Add test options to enable ARM64 tests

### Rationale

- Implement CI for ARM64

### Checklist

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -56,6 +56,7 @@ def pytest_addoption(parser: Parser):
         "--openstack-clouds-yaml",
         action="store",
         help="The OpenStack clouds yaml file for the charm to use.",
+        default="",
     )
     parser.addoption(
         "--use-existing-app",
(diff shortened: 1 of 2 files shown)


tests/conftest.py
```diff
@@ -56,6 +56,7 @@ def pytest_addoption(parser: Parser):
         "--openstack-clouds-yaml",
         action="store",
         help="The OpenStack clouds yaml file for the charm to use.",
+        default="",
```
yhaliaw, line 59: The default without  this would be `None` right?
Would changing to empty string have advantages?

yhaliaw (approved):
One minor comments. Not important.

tests/conftest.py
yanksyoon: I kept it so that it stays as type string and doesn't go to `None`. The default behavior without this would be an error saying the input is not given for this parameter.

cbartz (approved):

PR #349, opened 2024-08-08
fix: ssh health check failure status report

yanksyoon:
Applicable spec: <link>

### Overview

If health check fails on SSH health check, report status as unhealthy. We've let SSH health checks pass previously to determine why the network is unable to connect to instances.

### Rationale

SSH health check will return as healthy even if the machines cannot be connected via ssh.

From our usual Openstack console logs, we can see that the boot process takes around 30s (usually faster).

### Checklist

### DIFF
--- a/src/openstack_cloud/openstack_manager.py
+++ b/src/openstack_cloud/openstack_manager.py
@@ -502,6 +502,7 @@ class OpenstackRunnerManager:
         )
 
     @staticmethod
+    @retry(tries=5, delay=10, max_delay=60, backoff=2, local_logger=logger)
     def _ssh_health_check(conn: OpenstackConnection, server_name: str, startup: bool) -> bool:
         """Use SSH to check whether runner application is running.
 
@@ -524,7 +525,7 @@ class OpenstackRunnerManager:
             )
         except _SSHError as exc:
             logger.error("[ALERT]: Unable to SSH to server: %s, reason: %s", server_name, str(exc))
-            return True
+            return False
 
         result: invoke.runners.Result = ssh_conn.run("ps aux", warn=True)
         logger.debug("Output of `ps aux` on %s stderr: %s", server_name, result.stderr)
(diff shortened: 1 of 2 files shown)


arturo-seijas (dismissed):

src/openstack_cloud/openstack_manager.py
```diff
@@ -524,13 +532,17 @@ def _ssh_health_check(conn: OpenstackConnection, server_name: str, startup: bool
             )
         except _SSHError as exc:
             logger.error("[ALERT]: Unable to SSH to server: %s, reason: %s", server_name, str(exc))
-            return True
+            raise
```
yanksyoon, line 535: This needs to be raised in order for it to be retried.

yanksyoon (commented):
Review comments.

src/openstack_cloud/openstack_manager.py
```diff
@@ -497,11 +497,16 @@ def _health_check(
...
             return elapsed_min < CREATE_SERVER_TIMEOUT
-        return OpenstackRunnerManager._ssh_health_check(
-            conn=conn, server_name=server_name, startup=startup
-        )
+        try:
+            return OpenstackRunnerManager._ssh_health_check(
+                conn=conn, server_name=server_name, startup=startup
+            )
+        except _SSHError:
+            logger.warning("Health chedk failed, unable to SSH into server: %s", server_name)
```
cbartz, line 505:
```suggestion
            logger.warning("Health check failed, unable to SSH into server: %s", server_name)
```

cbartz (approved):

src/openstack_cloud/openstack_manager.py
```diff
@@ -497,11 +497,16 @@ def _health_check(
...
+        try:
+            return OpenstackRunnerManager._ssh_health_check(
+                conn=conn, server_name=server_name, startup=startup
+            )
+        except _SSHError:
+            logger.warning("Health check failed, unable to SSH into server: %s", server_name)
+            return False
 
     @staticmethod
+    @retry(tries=3, delay=5, max_delay=60, backoff=2, local_logger=logger)
```
yhaliaw, line 509: 3 tries with backoff of 2 and delay of 5.
This will result in waiting of 5, 10, 20 seconds.
It might be enough, I think we can try this.

yhaliaw (approved):
I am guessing this is the logic of this PR:

- The current `_get_ssh_connection` checks running the ssh by running `echo`. So if the ssh connection is returned then any further _SSHError can be treated as failure as the SSH was working.
- The `_get_ssh_connection` itself will be the only thing that needs to be retried. Since failure might be network issues, etc.

The retried timing feels on the shorter end, but might work fine. We can try it.

I think this should be fine. Approved.

PR #351, opened 2024-08-19
Use the refactored runner manager

yhaliaw:
### Overview

Use the refactored runner manager for OpenStack runner in charm.py.

### Rationale

The refactored interface should be used in the charm.

### Juju Events Changes

The `get-runners` juju action returns with some new fields ("busy", and "busy-runners").
This provides info on busy runners.
Added since it makes tests easier.

### Module Changes

Add a RunnerScaler class. This class uses the RunnerManager and provides the APIs the charm.py needs, e.g., scale up/down, flush runners, get runner infos.

### Checklist


src/manager/runner_scaler.py
```diff
@@ -0,0 +1,174 @@
...
+            if runner.github_state == GitHubRunnerState.OFFLINE
+            and runner.health == HealthState.HEALTHY
+        ]
+
+        try:
+            available_runners = set(runner.name for runner in idle_runners) | set(
+                runner.name for runner in offline_healthy_runners
+            )
+            logger.info(
+                "Current available runners (idle + healthy offline): %s", available_runners
```
jdkandersson, line 159: Perhaps log the healthy offline and healthy online as well?

jdkandersson (dismissed):

jdkandersson (dismissed):

jdkandersson (dismissed):

jdkandersson (dismissed):

src/charm.py
```diff
@@ -1231,39 +1231,62 @@ def _get_openstack_runner_manager(
...
             path = state.charm_config.path
 
-        # Empty image can be passed down due to a delete only case where deletion of runners do not
-        # depend on the image ID being available. Make sure that the charm goes to blocked status
-        # in hook where a runner may be created. TODO: 2024-07-09 This logic is subject to
-        # refactoring.
+        clouds = list(state.charm_config.openstack_clouds_yaml["clouds"].keys())
+        if len(clouds) > 1:
+            logger.warning(
+                "Multiple clouds defined in clouds.yaml. Using the first one to connect."
```
yanksyoon, line 1244: Minor nit: Should we print the first cloud name?

yanksyoon (commented):
Reviewed untiol tests/unit/test_charm.py, looks good so far! Thanks for working on it!

src/manager/github_runner_manager.py
```diff
@@ -72,14 +72,19 @@ def get_runners(
             Information on the runners.
         """
         runner_list = self.github.get_runner_github_info(self._path)
+        runner_list = [runner for runner in runner_list if runner.name.startswith(self._prefix)]
```
cbartz, line 75: I know you have not introduced a change here, but what happens if we have two deployments with the same application name, so perhaps the same unit names, registered to the same organisation? They would share the same prefix. Could this be a problem?

src/manager/runner_scaler.py
```diff
@@ -0,0 +1,215 @@
...
+        metric_stats = self._manager.cleanup()
+        delete_metric_stats = self._manager.flush_runners(flush_mode=flush_mode)
+        events = set(delete_metric_stats.keys()) | set(metric_stats.keys())
+        metric_stats = {
+            event_name: delete_metric_stats.get(event_name, 0) + metric_stats.get(event_name, 0)
+            for event_name in events
+        }
+        return metric_stats.get(metric_events.RunnerStop, 0)
+
+    def reconcile(self, quantity: int) -> int:
```
cbartz, line 111: This method looks quite big, takes a long time to read, and will probably have to be broken down soon because of its cognitive complexity. Perhaps we are already breaking it up.

pyproject.toml
```diff
@@ -13,6 +13,8 @@ omit = [
     # These are covered by `tests/integration/test_runner_manager_openstack.py`.
     "src/openstack_cloud/openstack_cloud.py",
     "src/openstack_cloud/openstack_runner_manager.py",
+    # Thin wrapper around GitHub API. Not a lot of value in unit tests.
+    "src/manager/github_runner_manager.py",
```
cbartz, line 17: Minor: Perhaps it would be better to put a `# pragma: no cover` in the classes of the source file, so that we are reminded every time we modify the module about the exclusion. Because the implementation may get more complex in the future, and we may want to have coverage, and by excluding the whole file we'll never get notified about missing coverage.

cbartz (commented):
Reviewed until tests

src/charm.py
yhaliaw: Can this be done as part of the replacing the openstack clouds yaml with individual openstack configurations?
Since the openstack-clouds-yaml configuration is getting removed, and multiple clouds will not be possible anymore.

pyproject.toml
yhaliaw: Good idea, I will update the code.

src/manager/github_runner_manager.py
yhaliaw: Yes, having two different deployments in different juju model, same application name with same unit number, registered on the same repo/org will cause problems. 
We will have to generation a unique string and store it to solve this issue. This was talked about during the early stages of rewrite of this charm, but StoredState was being removed.

Since we have a system to store old configuration as files, I think we can revisit this.

But I think this is out of scope of this PR.

src/manager/runner_scaler.py
yhaliaw: Are you planning on refactoring as part of your PR, or should I do it here.

cbartz: Its fine for me to let it as it is, but we'll probably have to refactor it once we touch it next time.

tests/unit/test_runner_scaler.py
```diff
@@ -0,0 +1,265 @@
...
+    # 2.
+    diff = runner_scaler.reconcile(1)
+    assert diff == 0
+    assert_runner_info(runner_scaler, online=1)
+
+    # 3.
+    runner_scaler.flush(flush_mode=FlushMode.FLUSH_IDLE)
+    assert_runner_info(runner_scaler, online=0)
+
+    # 3.
```
cbartz, line 197:
```suggestion
    # 4.
```

tests/unit/test_runner_scaler.py
```diff
@@ -0,0 +1,265 @@
...
+    Arrange: A RunnerScaler with no runners.
+    Act:
+        1. Reconcile to one runner.
+        2. Reconcile to one runner.
+        3. Flush idle runners.
+        4. Reconcile to one runner.
+    Assert:
+        1. Runner info has one runner.
+        2. No changes to number of runner.
+        3. Runner info has one runner.
```
cbartz, line 181: nit: need also something for 4.

cbartz (commented):
Some nit comments, otherwise looks good.

cbartz (approved):

jdkandersson (approved):

yanksyoon (approved):
LGTM!

PR #354, opened 2024-08-27
Feat/upgrade pydantic v2

alithethird:
Updated Pydantic version to v2.8.2
### Checklist


jdkandersson:
@alithethird could you please label the PR as noted in the description?

src/charm.py
```diff
@@ -1182,8 +1182,9 @@ def _on_image_relation_joined(self, _: ops.RelationJoinedEvent) -> None:
             return
 
         clouds_yaml = state.charm_config.openstack_clouds_yaml
-        cloud = list(clouds_yaml["clouds"].keys())[0]
-        auth_map = clouds_yaml["clouds"][cloud]["auth"]
+        # unsubscriptable-object / E1136 thinks cloud_yaml is kind of a list
+        cloud = list(clouds_yaml["clouds"].keys())[0]  # pylint: disable=E1136
+        auth_map = clouds_yaml["clouds"][cloud]["auth"]  # pylint: disable=E1136
```
jdkandersson, line 1187: Usually we put the pylint rule name rather than the ID to make it easier to undersantd. I realise it is above, just better not to use the ID at all

src/charm_state.py
```diff
@@ -862,16 +859,20 @@ def aproxy_address(self) -> Optional[str]:
...
+                aproxy_address = proxy_address.host
+
+            if proxy_address.port:
+                aproxy_address = f"{proxy_address.host}:{proxy_address.port}"
         else:
             aproxy_address = None
         return aproxy_address
 
-    @validator("use_aproxy")
+    @field_validator("use_aproxy")  # type: ignore
```
jdkandersson, line 875: What is the type: ignore here please?

src/charm_state.py
```diff
@@ -886,8 +887,14 @@ def check_use_aproxy(cls, use_aproxy: bool, values: dict) -> bool:
         Returns:
             Validated use_aproxy value.
         """
-        if use_aproxy and not (values.get("http") or values.get("https")):
-            raise ValueError("aproxy requires http or https to be set")
+        try:
+            if use_aproxy:
+                values_data = values.data  # type: ignore
```
jdkandersson, line 892: Here as well? Any linting disable rules should have an explanation why it is enabled (should be stated in our contributing guide)

src/metrics/runner.py
```diff
@@ -301,7 +301,11 @@ def _create_runner_stop(
         repo=runner_metrics.pre_job.repository,
         github_event=runner_metrics.pre_job.event,
         status=runner_metrics.post_job.status,
-        status_info=runner_metrics.post_job.status_info,
+        status_info=(
+            runner_metrics.post_job.status_info.dict()
```
jdkandersson, line 305: Is it no longer a dict?

src/openstack_cloud/openstack_runner_manager.py
```diff
@@ -194,7 +194,7 @@ def create_runner(self, registration_token: str) -> InstanceId:
                 instance_id=instance_id,
                 image=self._server_config.image,
                 flavor=self._server_config.flavor,
-                network=self._server_config.network,
+                network=str(self._server_config.network) if self._server_config.network else None,
```
jdkandersson, line 197: is it no longer a str?

src/repo_policy_compliance_client.py
```diff
@@ -42,7 +42,7 @@ def get_one_time_token(self) -> str:
         Returns:
             The one-time token to be used in a single request of repo policy compliance check.
         """
-        url = urljoin(self.base_url, "one-time-token")
+        url = urljoin(str(self.base_url), "one-time-token")
```
jdkandersson, line 45: I guess here and elsewhere, what is the reason to needing to do type casts please?

src/runner.py
```diff
@@ -716,7 +716,9 @@ def _configure_docker_proxy(self) -> None:
                 }
             }
         }
-        docker_client_proxy_content = json.dumps(docker_client_proxy)
+        docker_client_proxy_content = json.dumps(
+            docker_client_proxy, default=lambda x: x.unicode_string()
```
jdkandersson, line 720: What is the reasoning here please?

jdkandersson (commented):
@alithethird perhaps it would be better if we schedule a review call for this so we can take a look and discuss a bit 🙂

src/charm_state.py
alithethird: It is because mypy complains as:
```
src/charm_state.py:875: error: Value of type variable "_V2BeforeAfterOrPlainValidatorType" of function cannot be "Callable[[type[ProxyConfig], bool, dict[Any, Any]], bool]"  [type-var]
```

I should put it as 
```python
    @field_validator("use_aproxy")  # type: ignore[type-var]
```

right?

src/openstack_cloud/openstack_runner_manager.py
alithethird: It is no longer a str, it is a Url class or None

src/repo_policy_compliance_client.py
alithethird: AnyHttpUrl in pydantic v1 was returning str but in v2 they return Url class and has to be type casted to use in a lot of places unfortunately.

src/runner.py
alithethird: `docker_client_proxy` is a Url class and Url classes can not be serialized. To fix that we turn them into strings then do json.dump. The reason we use default and not just directly str cast is it can be None.

src/metrics/runner.py
alithethird: It was and still a `CodeInformation` class but in Pydantic v1 it did not complain. In Pydantic v2 it started to complain and fail integration tests.

src/charm_state.py
javierdelapuente: Your issue here is that `values` is not a `dict`, it is a `ValidationInfo`. Then, you can remove this exception and the next one in the function.

alithethird: I just tried that and can confirm we don't need the ignore comments when `ValidationInfo` type is used for `values`. Thanks @javierdelapuente !

src/charm.py
```diff
@@ -1182,8 +1182,9 @@ def _on_image_relation_joined(self, _: ops.RelationJoinedEvent) -> None:
             return
 
         clouds_yaml = state.charm_config.openstack_clouds_yaml
-        cloud = list(clouds_yaml["clouds"].keys())[0]
-        auth_map = clouds_yaml["clouds"][cloud]["auth"]
+        # unsubscriptable-object / thinks cloud_yaml is kind of a list
```
javierdelapuente, line 1185: Looking at charm_config it is `OpenStackCloudsYAML | None`, so it probably thinks it can be None, which it can be....

src/charm_state.py
```diff
@@ -496,15 +492,15 @@ def _parse_openstack_clouds_config(cls, charm: CharmBase) -> OpenStackCloudsYAML
 
         try:
             openstack_cloud.initialize(openstack_clouds_yaml)
-        except OpenStackInvalidConfigError as exc:
+        except (OpenStackInvalidConfigError, TypeError) as exc:
```
javierdelapuente, line 495: the linter has found something strange after the change in:
```
    openstack_clouds_yaml: OpenStackCloudsYAML | None
    openstack_clouds_yaml: OpenStackCloudsYAML | None = None
```

src/charm_state.py
```diff
@@ -862,18 +860,22 @@ def aproxy_address(self) -> Optional[str]:
             assert (
                 proxy_address is not None and proxy_address.host is not None
             )  # nosec for [B101:assert_used]
-            aproxy_address = (
-                proxy_address.host
-                if not proxy_address.port
-                else f"{proxy_address.host}:{proxy_address.port}"
-            )
+            if "http" in proxy_address.host:
```
javierdelapuente, line 863: is http or https in the host?

src/openstack_cloud/openstack_runner_manager.py
javierdelapuente: testing locally I do not need the change.

javierdelapuente (commented):
Added a few comments

cbartz:
@alithethird Is this still worked on?

alithethird:
Currently not working on this.

javierdelapuente:
@alithethird is it ok to close this PR? As there have been many changes in the code, maybe it would be better to start from scratch when we decide to go for this task again

PR #382, opened 2024-09-25
feat: Job label validation

cbartz:

### Overview

Adaptations to make the charm work with https://github.com/canonical/github-runner-manager/pull/17.

### Rationale

See rationale in  https://github.com/canonical/github-runner-manager/pull/17.

### Module Changes

`charm.py`: Pass over supported_labels.

### Library Changes

n/a

### Checklist

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -107,6 +107,8 @@ RECONCILE_RUNNERS_EVENT = "reconcile-runners"
 REACTIVE_MQ_DB_NAME = "github-runner-webhook-router"
 
 
+GITHUB_SELF_HOSTED_ARCH_LABELS = {"x64", "arm64"}
+
 logger = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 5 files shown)


src/charm.py
```diff
@@ -1281,18 +1283,39 @@ def _get_runner_scaler(
         )
         reactive_runner_config = None
         if reactive_config := state.reactive_config:
+            # The charm is not able to determine which architecture the runner is running on,
+            # so we add all architectures to the supported labels.
+            supported_labels = set(self._create_labels(state)) | GITHUB_SELF_HOSTED_ARCH_LABELS
```
jdkandersson, line 1288: Isn't that going to be a problem? How would it know that it can pick up `edge` for `x64` but not for `arm64`?

cbartz: This is not currently a problem, as we are assuming a different queue for each flavour in the first iteration of the reactive feature. This means that the x64/arm64 messages would be routed to a different queue, and the charms supporting arm64 would only connect to the queue for arm64. The label validation here is currently only there to try and catch programming or routing errors (but you are right that it wouldn't actually be able to detect those errors for the architectures).

Once we allow the charm to spawn different flavours (e.g. with custom images) this could become a problem, as we could end up with a queue of messages containing different architectures.

jdkandersson (approved):

yanksyoon (approved):
LGTM!

PR #390, opened 2024-10-14
feat: Adjustments for COS integration for reactive mode

cbartz:

### Overview

- Adaptions to make the charm work with https://github.com/canonical/github-runner-manager/pull/21

Also adapted the dashboard to show active runners:

Screenshot from 2024-10-11 23-50-23

### Rationale

- See rationale in https://github.com/canonical/github-runner-manager/pull/21
- Reactive mode shouldn't have any idle runners, so we also visualise active runners.

### Checklist

- [x] The changes do not introduce any regression in code or tests related to LXD runner mode.


src/shared_fs.py
```diff
@@ -19,11 +20,19 @@
 from utilities import execute_command
 
 DIR_NO_MOUNTPOINT_EXIT_CODE = 32
+METRICS_STORAGE_USER_CONFIG = SystemUserConfig(
+    user="ubuntu",
+    group="ubuntu",
+)
```
cbartz, line 26: For LXD mode, we still use the ubuntu user for metrics storage to reduce the risk of breaking something. This can be changed in a follow-up PR.

tests/integration/test_charm_metrics_success.py
```diff
@@ -59,6 +59,10 @@ async def test_charm_issues_runner_installed_metric(
     """
     await instance_helper.ensure_charm_has_runner(app)
 
+    # Set the number of virtual machines to 0 to speedup reconciliation
+    await app.set_config({VIRTUAL_MACHINES_CONFIG_NAME: "0"})
+    await reconcile(app=app, model=model)
```
cbartz, line 64: We need to reconcile because the runner installed metric is now issued in the next reconciliation and no longer directly in the build function (because the installed timestamp now lives in cloud-init).

tests/integration/test_charm_runner.py
```diff
@@ -155,13 +161,6 @@ async def test_flush_runner_and_resource_config(
         assert action.status == "completed"
         assert action.results["delta"]["virtual-machines"] == "0"
 
-        await wait_for(lambda: workflow.update() or workflow.status == "completed")
-        action = await app.units[0].run_action("flush-runners")
-        await action.wait()
-
-        assert action.status == "completed"
-        assert action.results["delta"]["virtual-machines"] == "1"
```
cbartz, line 163: This is not necessary to test (and its brittle, because flush runners could still be run on an "idle" runner, we would need to wait_for the runner to be considered offline/idle). I have added tear down code to the fixture above to ensure that a subsequent test is not affected by this test.

tests/integration/test_charm_runner.py
```diff
@@ -40,8 +42,12 @@ async def app_fixture(
     Ensure the charm has one runner before starting a test.
     """
     await instance_helper.ensure_charm_has_runner(basic_app)
+
     yield basic_app
 
+    await basic_app.set_config({VIRTUAL_MACHINES_CONFIG_NAME: "0"})
+    await reconcile(basic_app, basic_app.model)
```
cbartz, line 49: Ensure that follow-up tests are not affected by current live runners.

tests/integration/test_runner_manager_openstack.py
```diff
@@ -142,6 +143,8 @@ async def openstack_runner_manager_fixture(
         server_config=server_config,
         runner_config=runner_config,
         service_config=service_config,
+        # we assume the test runs as ubuntu user
+        system_user_config=SystemUserConfig(user="ubuntu", group="ubuntu"),
```
cbartz, line 147: This assumes that this test is run on the CI as a ubuntu user. I assume our tests are not intended to be run locally, so this should be fine.
An alternative would be to do more patching of internas (with the downside of not "really" doing an integration test).

tests/integration/test_runner_manager_openstack.py
```diff
@@ -202,22 +205,30 @@ def workflow_is_status(workflow: Workflow, status: str) -> bool:
...
     if len(runner_list) == num:
         return
 
     # The openstack server can take sometime to fully clean up or create.
-    await wait_for(lambda: len(runner_manager.get_runners()) == num)
+    await wait_for(
+        lambda: len(runner_manager.get_runners()) == num,
+        timeout=timeout,
+        check_interval=check_interval,
+    )
```
cbartz, line 231: Increased the timeout by default as the tests were quite brittle. Also increased the check_interval as there is so much logged on each run.

requirements.txt
```diff
@@ -12,4 +12,4 @@ cosl ==0.0.15
 # juju 3.1.2.0 depends on pyyaml<=6.0 and >=5.1.2
 PyYAML ==6.0.*
 pyOpenSSL==24.2.1
-github_runner_manager @ git+https://github.com/canonical/github-runner-manager.git@33bbaff42d7cc0f250006fdd08d24659cef364c9
+github_runner_manager @ git+https://github.com/canonical/github-runner-manager.git@feat/reactive-cos-ISD-1537
```
jdkandersson, line 15: Need to fix this before merge

src/charm.py
```diff
@@ -1033,6 +1043,33 @@ def _install_deps(self) -> None:
         logger.info("Installing charm dependencies.")
         self._apt_install(["run-one"])
 
+    @staticmethod
```
jdkandersson, line 1046: Could this be done as a function in the module rather than on the class if it doesn't require an instance?

src/shared_fs.py
jdkandersson: Please make sure we have a ticket for this

requirements.txt
cbartz: Sure, which is why I still mark this PR as a draft until https://github.com/canonical/github-runner-manager/pull/21 gets merged.

src/shared_fs.py
cbartz: Done

jdkandersson (approved):

src/charm.py
```diff
@@ -454,6 +458,12 @@ def _common_install_code(self, state: CharmState) -> bool:  # noqa: C901
             logger.error("Failed to install charm dependencies")
             raise
 
+        try:
+            _setup_runner_manager_user()
+        except SubprocessError:
+            logger.error("Failed to setup runner manager user")
```
Thanhphan1147, line 464: I think we can maybe use `logger.exception` here?

Thanhphan1147 (approved):
LGTM, I only have 1 small comment.

src/charm.py
cbartz: Thank you. The error will get re-raised in the line afterwards and will cause the charm to go into error state. I think the stack trace will be available in the logs then. This pattern is used also in the rest of the function (which might not be visible in the diff), so I wouldn't like to change this in this PR.

PR #415, opened 2024-12-04
fix: Continue reconciliation after a health check execution error

cbartz:
Applicable spec: n/a

### Overview

Catch health check execution errors and continue reconciliation.

### Rationale

We had an issue in production where the health check failed with an invoke.exception.CommandTimeout for an ssh command, and as this error was not caught, the whole reconciliation stopped and so no runners were cleaned up and rebuilt.

### Juju Events Changes

n/a

### Module Changes

- `github_runner_manager/openstack_cloud/health_checks.py`: Catch errors on ssh execution and raise a OpenstackHealthCheckError
- `github_runner_manager/openstack_cloud/openstack_runner_manager.py`: Catch the error and map it to an unknown health state. Do not perform cleanup tasks for runners in unknown health state.
- 
### Library Changes

n/a

### Checklist

- [x] The changes do not introduce any regression in code or tests related to LXD runner mode.

### DIFF
--- a/github-runner-manager/src/github_runner_manager/openstack_cloud/health_checks.py
+++ b/github-runner-manager/src/github_runner_manager/openstack_cloud/health_checks.py
@@ -8,6 +8,7 @@ from datetime import datetime, timedelta
 
 import invoke
 from fabric import Connection as SSHConnection
+
 from github_runner_manager.errors import KeyfileError, OpenstackHealthCheckError, SSHError
 from github_runner_manager.manager.cloud_runner_manager import CloudInitStatus, CloudRunnerState
 from github_runner_manager.openstack_cloud.constants import (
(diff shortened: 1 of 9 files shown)


github-runner-manager/src/github_runner_manager/openstack_cloud/health_checks.py
```diff
@@ -48,8 +52,11 @@ def check_runner(openstack_cloud: OpenstackCloud, instance: OpenstackInstance) -
...
         )
         return False
-    except SSHError:
-        logger.exception("SSH Failed on %s, marking as unhealthy.")
+    except _SSHError:
+        logger.exception(
+            "Unable to get SSH connection for instance %s, marking as unhealthy.",
+            instance.server_name,
+        )
         return False
```
cbartz, line 60: This is confusing and should probably not return False, but instead raise the health check error. The problem is that this would result in the server not being purged, and this could be a risky change.

jdkandersson (approved):

yhaliaw (approved):

PR #218, opened 2024-02-06
Use previous token to flush runners

cbartz:
### Overview

Use the previously saved token to remove existing runners. If the previous token has expired, use a force removal (ignoring gh api errors and just removing lxd instances).

In addition, any invalid token errors in the charm code will be caught and the charm will go into `BlockedStatus`, requiring an operator to update the token. A message is included in the `BlockedStatus`, so this PR should fix https://github.com/canonical/github-runner-operator/issues/89.

### Rationale

If the new token on config changed is no longer valid for the current organisation/repo, the removal of existing runners may fail (due to HTTP 403 Forbidden).

```
Traceback (most recent call last):
  File "/var/lib/juju/agents/unit-github-runner-1/charm/./src/charm.py", line 926, in <module>
    main(GithubRunnerCharm)
  File "/var/lib/juju/agents/unit-github-runner-1/charm/venv/ops/main.py", line 456, in main
    _emit_charm_event(charm, dispatcher.event_name)
  File "/var/lib/juju/agents/unit-github-runner-1/charm/venv/ops/main.py", line 144, in _emit_charm_event
    event_to_emit.emit(*args, **kwargs)
  File "/var/lib/juju/agents/unit-github-runner-1/charm/venv/ops/framework.py", line 351, in emit
    framework._emit(event)
  File "/var/lib/juju/agents/unit-github-runner-1/charm/venv/ops/framework.py", line 853, in _emit
    self._reemit(event_path)
  File "/var/lib/juju/agents/unit-github-runner-1/charm/venv/ops/framework.py", line 943, in _reemit
    custom_handler(event)
  File "/var/lib/juju/agents/unit-github-runner-1/charm/./src/charm.py", line 78, in func_with_catch_errors
    func(self, event)
  File "/var/lib/juju/agents/unit-github-runner-1/charm/./src/charm.py", line 505, in _on_config_changed
    prev_runner_manager.flush(FlushMode.FORCE_FLUSH_BUSY_WAIT_REPO_CHECK)
  File "/var/lib/juju/agents/unit-github-runner-1/charm/src/runner_manager.py", line 564, in flush
    remove_token = self._clients.github.get_runner_remove_token(self.config.path)
  File "/var/lib/juju/agents/unit-github-runner-1/charm/src/github_client.py", line 118, in get_runner_remove_token
    token = self._client.actions.create_remove_token_for_repo(
  File "/var/lib/juju/agents/unit-github-runner-1/charm/venv/ghapi/core.py", line 62, in __call__
[...]

PR #441, opened 2025-01-30
Some fixes for integration tests

javierdelapuente:
Applicable spec: <link>

### Overview

A few improvements to make integration tests less flaky, like:

* Bigger revision-history-limit for the image-builder.
* Bigger timeout for the CREATE_SERVER_TIMEOUT for the test_runner_manager_openstack.py
* Set the charm in Maintenance mode when the debug-ssh integration is added.
* More reasonable timeouts.
* hold autorefresh and wait for last refresh for snaps (setting aproxy can fail is the snap is refreshing).
* Flush runners on changing token, repo or labels. Store token on token change (otherwise the flush will be called every on config changed).

### Checklist

### DIFF
--- a/github-runner-manager/src/github_runner_manager/templates/openstack-userdata.sh.j2
+++ b/github-runner-manager/src/github_runner_manager/templates/openstack-userdata.sh.j2
@@ -7,7 +7,7 @@ hostnamectl set-hostname github-runner
 # Write .env contents
 su - ubuntu -c 'cd ~/actions-runner && echo "{{ env_contents }}" > .env'
 
-snap refresh --hold=2h
+snap refresh --hold=24h
 snap watch --last=auto-refresh?
 
 {% if aproxy_address %}
--- a/github-runner-manager/src/github_runner_manager/templates/openstack-userdata.sh.j2
+++ b/github-runner-manager/src/github_runner_manager/templates/openstack-userdata.sh.j2
@@ -7,6 +7,9 @@ hostnamectl set-hostname github-runner
 # Write .env contents
 su - ubuntu -c 'cd ~/actions-runner && echo "{{ env_contents }}" > .env'
 
+snap refresh --hold=2h
+snap watch --last=auto-refresh?
+
 {% if aproxy_address %}
 snap install aproxy --edge
 snap set aproxy proxy={{ aproxy_address }} listen=:54969
--- a/src/charm.py
+++ b/src/charm.py
@@ -341,7 +341,7 @@ class GithubRunnerCharm(CharmBase):
 
         flush_and_reconcile = False
         if state.charm_config.token != self._stored.token:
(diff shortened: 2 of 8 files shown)


src/charm.py
```diff
@@ -339,22 +339,26 @@ def _on_config_changed(self, _: ConfigChangedEvent) -> None:
...
         self._set_reconcile_timer()
 
+        flush_and_reconcile = False
         if state.charm_config.token != self._stored.token:
-            self._stored.token = None
+            self._stored.token = self.config[TOKEN_CONFIG_NAME]
+            flush_and_reconcile = True
         if self.config[PATH_CONFIG_NAME] != self._stored.path:
             self._stored.path = self.config[PATH_CONFIG_NAME]
+            flush_and_reconcile = True
```
cbartz, line 348: Just a heads up, this is not strictly related to this change (as it would also occur on upcoming reconciliation), but I think that reconcile fails if the token does not have enough permissions to unregister the current runners from the current path or to subscribe to the new path. https://github.com/canonical/github-runner-operator/blob/main/github-runner-manager/src/github_runner_manager/manager/runner_manager.py#L211 will raise an exception.

cbartz (approved):
Thank you!

src/charm.py
javierdelapuente: You are right.  

I did this mostly because there is a test called `test_charm_fork_path_change.py` changes the repo and then reconciles. I believe the runners should be flusnhed in that case. 

However, as you say that problem will happen with this change or without it, do you think I should keep this code?

cbartz: Yes I think you can keep the code. I found the bugfix I created in the past for LXD https://github.com/canonical/github-runner-operator/pull/218 . We need to create a separate bug ticket for Openstack.

github-runner-manager/src/github_runner_manager/templates/openstack-userdata.sh.j2
```diff
@@ -7,6 +7,9 @@ hostnamectl set-hostname github-runner
 # Write .env contents
 su - ubuntu -c 'cd ~/actions-runner && echo "{{ env_contents }}" > .env'
 
+snap refresh --hold=24h
+snap watch --last=auto-refresh?
```
jdkandersson, line 11: What is the question mark at the end here please?

javierdelapuente: it does not fail if there is no auto refresh in place:
```
      --last=          Select last change of given type (install, refresh, remove, try, auto-refresh, etc.). A question mark at the end of the type means to do nothing (instead of returning an error) if no change of the given type is found. Note the question mark could need protecting from the shell.
```

jdkandersson (approved):

PR #466, opened 2025-02-25
Use Just-In-Time tokens for GitHub runner registration

javierdelapuente:
Applicable spec: <link>

### Overview

This PR uses [JIT token](https://github.blog/changelog/2023-06-02-github-actions-just-in-time-self-hosted-runners/) for runner registration instead of normal registration tokens.

The main advantage of JIT tokens is that they can only be used one, so there is no risk if they are leaked.

Several changes have been done to the project, as to request a JIT token, it is necessary to send as parameters:
- The runner name.
- The labels. No labels are added by default, so "self-hosted" and "linux" have to be added manually.
- The runner_group_id (the charm currently has as a config option the group name, not the id).

As there is one JIT token per runner, a JIT token has to be requested per runner, and it is done just before creating the runner (that can be in a forked process).

To get the runner_group_id from the group name, a call has to be done to the GitHub API. This call is not currently supported by ghapi. See [issue](https://github.com/AnswerDotAI/ghapi/issues/189).

To simplify the process, the instance_id is now the same as the instance_name (all is instance_id), as the name is used for both OpenStack and GitHub.

The method `CloudRunnerInstance.get_runner` has been deleted, as it is not used not tested.

Runners with JIT for orgs have been tested with an org in here:
image

### Checklist

### DIFF
--- a/github-runner-manager/src/github_runner_manager/constants.py
+++ b/github-runner-manager/src/github_runner_manager/constants.py
@@ -4,7 +4,7 @@
 """Constants for the library."""
 
 GITHUB_SELF_HOSTED_ARCH_LABELS = {"x64", "arm64"}
-GITHUB_DEFAULT_LABELS = {"x64", "arm64"}
+GITHUB_DEFAULT_LABELS = {"self-hosted", "linux"}
 
 # Pending to be removed when the application works in standalone mode.
 RUNNER_MANAGER_USER = "runner-manager"
--- a/github-runner-manager/src/github_runner_manager/constants.py
+++ b/github-runner-manager/src/github_runner_manager/constants.py
@@ -4,6 +4,7 @@
 """Constants for the library."""
 
 GITHUB_SELF_HOSTED_ARCH_LABELS = {"x64", "arm64"}
+GITHUB_DEFAULT_LABELS = {"x64", "arm64"}
 
 # Pending to be removed when the application works in standalone mode.
 RUNNER_MANAGER_USER = "runner-manager"
--- a/github-runner-manager/src/github_runner_manager/github_client.py
+++ b/github-runner-manager/src/github_runner_manager/github_client.py
@@ -160,7 +160,7 @@ class GithubClient:
         return token["token"]
 
     @catch_http_errors
-    def get_runner_registration_token(
+    def get_runner_registration_jittoken(
         self, path: GitHubPath, instance_id: str, labels: list[str]
     ) -> str:
         """Get token from GitHub used for registering runners.
@@ -168,15 +168,16 @@ class GithubClient:
         Args:
             path: GitHub repository path in the format '<owner>/<repo>', or the GitHub organization
                 name.
-            instance_id: TODO
-            labels: TODO
+            instance_id: Instance ID of the runner.
+            labels: Labels for the runner.
 
         Returns:
             The registration token.
         """
         token: RegistrationToken
         if isinstance(path, GitHubRepo):
-            # JAVI not sure about this
+            # The supposition is that the runner_group_id 1 is the default.
+            # If the repo does not belong to an org, there is no way to get the runner_group_id.
             runner_group_id = 1
             token = self._client.actions.generate_runner_jitconfig_for_repo(
                 owner=path.owner,
@@ -186,9 +187,8 @@ class GithubClient:
                 labels=labels,
             )
         elif isinstance(path, GitHubOrg):
-            # JAVI we need the group id from the group in
-            # self.path.group
-            # TODO cache this.
+            # We cannot cache it in here, as we are running in a forked process.
+            # Pending to review.
             runner_group_id = self._get_runner_group_id(path)
             token = self._client.actions.generate_runner_jitconfig_for_org(
                 org=path.org,
@@ -202,12 +202,14 @@ class GithubClient:
         return token["encoded_jit_config"]
 
     def _get_runner_group_id(self, org: GitHubOrg) -> int:
-        """TODO. JUST TO TEST.
+        """Get runner_group_id from group name for an org.
 
-        LET'S SEE GHAPI IF THEY RELEASE A NEW VERSION BEFORE IMPROVING
-        THIS
+        Once this function is implemented in the ghapi library, we should
+        use that instead. See https://github.com/AnswerDotAI/ghapi/issues/189.
+
+        No pagination is used, so if there are more than 30 groups, this
+        function could fail.
         """
-        # Be careful, no pagination in here.
         url = f"https://api.github.com/orgs/{org.org}/actions/runner-groups"
         headers = {
             "Accept": "application/vnd.github+json",
--- a/github-runner-manager/src/github_runner_manager/github_client.py
+++ b/github-runner-manager/src/github_runner_manager/github_client.py
@@ -175,8 +175,6 @@ class GithubClient:
             The registration token.
         """
         token: RegistrationToken
-        instance_id  # pylint: disable=pointless-statement
-        labels  # pylint: disable=pointless-statement
         if isinstance(path, GitHubRepo):
             # JAVI not sure about this
             runner_group_id = 1
@@ -201,7 +199,7 @@ class GithubClient:
         else:
             assert_never(token)
 
-        return token["token"]
+        return token["encoded_jit_config"]
 
     def _get_runner_group_id(self, org: GitHubOrg) -> int:
         """TODO. JUST TO TEST.
--- a/github-runner-manager/src/github_runner_manager/github_client.py
+++ b/github-runner-manager/src/github_runner_manager/github_client.py
@@ -206,10 +206,10 @@ class GithubClient:
         Once this function is implemented in the ghapi library, we should
         use that instead. See https://github.com/AnswerDotAI/ghapi/issues/189.
 
-        No pagination is used, so if there are more than 30 groups, this
+        No pagination is used, so if there are more than 100 groups, this
         function could fail.
         """
-        url = f"https://api.github.com/orgs/{org.org}/actions/runner-groups"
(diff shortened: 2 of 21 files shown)


github-runner-manager/src/github_runner_manager/manager/runner_manager.py
```diff
@@ -369,13 +376,17 @@ def _issue_runner_metrics(self, metrics: Iterator[RunnerMetrics]) -> IssuedMetri
     class _CreateRunnerArgs:
         """Arguments for the _create_runner function.
 
+        This arguments are used in the forked processes and should be reviewed.
```
jdkandersson, line 379: What is meant here by should be reviewed?

javierdelapuente: Not related to this PR, but I think we should check the multiprocess thing. For example, we are using logging from the forked process, and it is not recommended.

github-runner-manager/src/github_runner_manager/github_client.py
```diff
@@ -159,27 +160,72 @@ def get_runner_remove_token(self, path: GitHubPath) -> str:
...
         Returns:
             The registration token.
         """
         token: RegistrationToken
         if isinstance(path, GitHubRepo):
-            token = self._client.actions.create_registration_token_for_repo(
-                owner=path.owner, repo=path.repo
+            # The supposition is that the runner_group_id 1 is the default.
+            # If the repo does not belong to an org, there is no way to get the runner_group_id.
+            runner_group_id = 1
```
yhaliaw, line 181: Does this have to be a separate line?
Can it be inlined to 186? Since the comment should be clear on what is happening.

github-runner-manager/src/github_runner_manager/github_client.py
```diff
@@ -159,27 +160,72 @@ def get_runner_remove_token(self, path: GitHubPath) -> str:
...
+            labels: Labels for the runner.
 
         Returns:
             The registration token.
         """
         token: RegistrationToken
         if isinstance(path, GitHubRepo):
-            token = self._client.actions.create_registration_token_for_repo(
-                owner=path.owner, repo=path.repo
+            # The supposition is that the runner_group_id 1 is the default.
```
yhaliaw, line 179: Is this endpoint tested with other values? Maybe the runner_group_id can be `None` or other values such as 0.

github-runner-manager/src/github_runner_manager/github_client.py
```diff
@@ -159,27 +160,72 @@ def get_runner_remove_token(self, path: GitHubPath) -> str:
...
+        No pagination is used, so if there are more than 30 groups, this
+        function could fail.
+        """
+        url = f"https://api.github.com/orgs/{org.org}/actions/runner-groups"
+        headers = {
+            "Accept": "application/vnd.github+json",
+            "Authorization": f"Bearer {self._token}",
+            "X-GitHub-Api-Version": "2022-11-28",
+        }
+        r = requests.get(url, headers=headers, timeout=30)
```
yhaliaw, line 219: I think we should generally avoid using single character variable names, such as `r`.

github-runner-manager/src/github_runner_manager/manager/runner_manager.py
```diff
@@ -118,10 +122,13 @@ def create_runners(self, num: int) -> tuple[InstanceId, ...]:
             List of instance ID of the runners.
         """
         logger.info("Creating %s runners", num)
-        registration_token = self._github.get_registration_token()
 
+        labels = list(self._labels)
+        # This labels are added by default by the github agent, but with JIT tokens
+        # we have to add them manually.
```
yhaliaw, line 128: I would think this is a issue on GitHub part? Should we report this to GitHub?

github-runner-manager/src/github_runner_manager/github_client.py
javierdelapuente: sure!

github-runner-manager/src/github_runner_manager/github_client.py
javierdelapuente: The field is necessary. With 0 I got 404, with other non-existent values 403 and with None 400. I believe 0 is our best bet

github-runner-manager/src/github_runner_manager/manager/runner_manager.py
javierdelapuente: Not sure about that, in the normal case you have the option of not using those labels with the [`--no-default-labels` flag](https://docs.github.com/en/actions/hosting-your-own-runners/managing-self-hosted-runners/using-self-hosted-runners-in-a-workflow#using-default-labels-to-route-jobs, But in this case, the labels are fixed when getting the registration token.

github-runner-manager/src/github_runner_manager/manager/runner_manager.py
jdkandersson: Ok, let's make a story for that

jdkandersson (approved):

yhaliaw (approved):

github-runner-manager/src/github_runner_manager/github_client.py
cbartz: There is always the Default runner group, and it seems to be implicit knowledge in github that this is always has id 1 . But yeah, this may be risky in the future if github changes something internally. 

You could use https://docs.github.com/en/rest/actions/self-hosted-runner-groups?apiVersion=2022-11-28#list-github-hosted-runners-in-a-group-for-an-organization and then pick the first id .

github-runner-manager/src/github_runner_manager/constants.py
```diff
@@ -4,6 +4,7 @@
 """Constants for the library."""
 
 GITHUB_SELF_HOSTED_ARCH_LABELS = {"x64", "arm64"}
+GITHUB_DEFAULT_LABELS = {"self-hosted", "linux"}
```
cbartz, line 7: what about the arch? Like x64 ? This is also automatically added by github.

github-runner-manager/src/github_runner_manager/manager/cloud_runner_manager.py
```diff
@@ -173,19 +159,16 @@ def name_prefix(self) -> str:
...
-        """Create a self-hosted runner.
-
-        Args:
-            registration_token: The GitHub registration token for registering runners.
-        """
+    def generate_instance_id(self) -> InstanceId:
+        """Generate an intance_id to name a runner."""
 
     @abc.abstractmethod
-    def get_runner(self, instance_id: InstanceId) -> CloudRunnerInstance | None:
```
cbartz, line 184: is this interface change required?

cbartz (commented):
Thanks. Looks good. Mostly minor comments.

github-runner-manager/src/github_runner_manager/constants.py
javierdelapuente: This is added by the tags in the image builder

github-runner-manager/src/github_runner_manager/manager/cloud_runner_manager.py
javierdelapuente: no, just removing unused/untested code.

github-runner-manager/src/github_runner_manager/github_client.py
```diff
@@ -159,27 +160,71 @@ def get_runner_remove_token(self, path: GitHubPath) -> str:
...
-        return token["token"]
+        return token["encoded_jit_config"]
+
+    def _get_runner_group_id(self, org: GitHubOrg) -> int:
+        """Get runner_group_id from group name for an org.
+
+        Once this function is implemented in the ghapi library, we should
+        use that instead. See https://github.com/AnswerDotAI/ghapi/issues/189.
+
+        No pagination is used, so if there are more than 30 groups, this
```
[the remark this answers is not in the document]
javierdelapuente: good point, I will add it

github-runner-manager/src/github_runner_manager/constants.py
cbartz: Can you double-check that "x64" is added? I thought only "amd64".

github-runner-manager/src/github_runner_manager/manager/cloud_runner_manager.py
javierdelapuente: the issue is that as I was changing things in the instance_id, probably I was also breaking this. As there are no tests and no users... better to remove it...

github-runner-manager/src/github_runner_manager/constants.py
javierdelapuente: We add amd64 manually as a label, the X64 comes from the image builder. As I will retest an org runner before merging I will review it (but I would say they are put here: https://github.com/canonical/github-runner-operator/blob/use-jit/src/charm.py#L577)

cbartz (commented):
Thanks. Just one remaining comment regarding the tags.

cbartz (approved):

github-runner-manager/src/github_runner_manager/github_client.py
javierdelapuente: I will add a story to tackle this. The issue is that a repo is not necessarily a part of an org, and outside of an org, there are no runner groups

github-runner-manager/src/github_runner_manager/constants.py
javierdelapuente: The X64 is correctly taken from the image-builder:

Screenshot from 2025-02-26 09-47-08

javierdelapuente:
Merging bypassing rules. All integ tests (except the forking ones) were tested locally.

PR #469, opened 2025-02-28
Some refactors to prepare for reactive / non-reactive combination

javierdelapuente:
Applicable spec: <link>

### Rationale

This PR is a combination of refactors for the github-runner-operator. As it is very difficult to pass the integration tests, I will not split it.

The instance name (also server_name or id in other contexts) is replaced by a new entity, the InstanceID that represents the instance name, both in the cloud provider and in the GitHub. The reasons to extract this new concept are:
- Currently it is the cloud provider the one creating it. It is not clear that this is the responsability of the cloud provider, as it is used everywhere (and the first use is to register the runner with GitHub using the a JIT token).
- In many parts, there are checks about the prefix (to see if some name matches the prefix). They are minor bugs, as it is not checked correctly (in some places the check included the '-' after the prefix, in others it didn't). With this entity, that function in centralized.
- The InstanceID real name, must be valid everywhere (cloud providers, GitHub, metrics file names...), so it sounds reasonable to have it as centralized as possible.
- Extra information can be included in the runner name. In this PR, the runner can be easily identified as reactive, both for tracing and to know the origin of the runner.

`expected_runner_quantity` is now also valid in reactive mode (in `runner_scaler.reconcile`), in preparation for the integration of reactive and non-reactive.

Hide ssh stdout and stderr. From logs. If we ever want to log it, we should do it on purpose, not on the library (this can be a minor security issue, as we are showing internal information from the runner in the logs).

Several TypeDicts, like RunnerApplication,SelfHostedRunnerList, RegistrationToken and RemoveToken are now classes (pydantic models), as otherwise we are using in our business model dictionaries from an external provider GitHub.

Changed the default config of `virtual-instances` to 0, as this field is deprecated now.

The file `tests/unit/mock_runner_managers.py` is deleted as it is not used (it is repeated).

Improvements to run  integration tests locally with the `--use-existing-app-suffix` option to not have to wait for the builder to prepare images in every iteration...

### Checklist


github-runner-manager/src/github_runner_manager/manager/models.py
```diff
@@ -0,0 +1,134 @@
...
+    @property
+    def name(self) -> str:
+        """Returns the name of the instance.
+
+        Returns:
+           Name of the instance
+        """
+        # Having a not number as a separator is ok, as the prefix should end
+        # with a number (it is the unit number).
+        return f"{self.prefix}{'r' if self.reactive else '-'}{self.suffix}"
```
jdkandersson, line 43: `f"{self.prefix}-{'r' if self.reactive else 'n'}-{self.suffix}"`

jdkandersson: We should check if the name is too long and crash

github-runner-manager/src/github_runner_manager/manager/models.py
```diff
@@ -0,0 +1,134 @@
...
+            suffix = name.removeprefix(f"{prefix}")
+        else:
+            raise ValueError(f"Invalid runner name {name} for prefix {prefix}")
+
+        # The separator from prefix and suffix indicates whether the runner is reactive.
+        reactive = False
+        separator = suffix[:1]
+        if separator == "r":
+            reactive = True
+        suffix = suffix[1:]
```
jdkandersson, line 69: Need to fix this as above

github-runner-manager/src/github_runner_manager/manager/models.py
```diff
@@ -0,0 +1,134 @@
...
+        The prefix must end with a number (it should be a unit name).
+
+        Args:
+           prefix: Application prefix (unit name of the charm).
+           name: Name of the runner instance.
+
+        Returns:
+           True if the instance name is part the applicatoin with the prefix.
+        """
+        if name.startswith(prefix):
```
jdkandersson, line 114: Need to change this

jdkandersson (approved):

github-runner-manager/src/github_runner_manager/manager/models.py
```diff
@@ -0,0 +1,149 @@
...
+    suffix: str
+
+    @property
+    def name(self) -> str:
+        """Returns the name of the instance.
+
+        Returns:
+           Name of the instance
+        """
+        return f"{self.prefix}-{'r' if self.reactive else 'n'}-{self.suffix}"
```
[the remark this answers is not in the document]
javierdelapuente: thanks! I did not see it!

github-runner-manager/src/github_runner_manager/openstack_cloud/health_checks.py
```diff
@@ -280,7 +282,7 @@ def _execute_ssh_command(ssh_conn: SSHConnection, command: str) -> invoke.runner
         _SSHError: If the command execution failed.
     """
     try:
-        return ssh_conn.run(command, warn=True, timeout=30)
+        return ssh_conn.run(command, warn=True, timeout=30, hide=True)
```
cbartz, line 285: :heart:

github-runner-manager/tests/unit/manager/test_models.py
```diff
@@ -0,0 +1,76 @@
...
+    prefix = "theprefix"
+    instance_id = InstanceID.build(prefix)
+
+    name = "wrong" + instance_id.name
+
+    with pytest.raises(ValueError):
+        _ = InstanceID.build_from_name(prefix, name)
+
+
+def test_build_instance_id_fails_when_very_long_name():
```
cbartz, line 68: shouldn't this test not also apply for build_from_name ? I didn't see a length check in the build_from_name method

cbartz (commented):
Thank you! One comment regarding the length check.

github-runner-manager/tests/unit/manager/test_models.py
javierdelapuente: I am not sure, a very name should not exist (or we are broken). If it was generated by something outside of our app, well, we cannot do anything in that case

cbartz (approved):

PR #475, opened 2025-03-10
Fix delete offline in github when building reactive

javierdelapuente:
Applicable spec: <link>

### Overview

A GitHub runner in offline mode is not necessarily offline, that means:
- A reactive runner in BUILD state in OpenStack is correct (it is in the creation process).
- A reactive runner that is ACTIVE, running the cloud init commands before running run.sh is correct.
- A runner ACTIVE that is rebooting/restarting the network is correct (busy runner that is offline).

Only the two first cases is tackled in this PR, as it is probably the most taxing. The other case will be fixed in future refactoring PRs.

Besides, in the GitHub client, typing was wrong (a dictionary was still being returned). Fixed to return the proper pydantic class. It could be reasonable to have both a typed dict for GitHub and then another "typed" class to use in the business logic, but that is not tackled in this PR.

### Checklist

### DIFF
--- a/github-runner-manager/src/github_runner_manager/manager/github_runner_manager.py
+++ b/github-runner-manager/src/github_runner_manager/manager/github_runner_manager.py
@@ -94,10 +94,10 @@ class GitHubRunnerManager:  # pragma: no cover
         )
 
     def delete_runners(self, runners: list[SelfHostedRunner]) -> None:
-        """TODO.
+        """Delete runners in GitHub.
 
         Args:
-            runners: TODO
+            runners: list of runners to delete.
         """
         for runner in runners:
             self.github.delete_runner(self._path, runner.id)
--- a/github-runner-manager/src/github_runner_manager/manager/github_runner_manager.py
+++ b/github-runner-manager/src/github_runner_manager/manager/github_runner_manager.py
@@ -78,7 +78,8 @@ class GitHubRunnerManager:  # pragma: no cover
             for runner in runner_list
             if InstanceID.name_has_prefix(self._prefix, runner.name)
         ]
-        # Calculate instance_id
+        # Calculate instance_id, as it is not calculated by the GithubClient as it
+        # does not have information about the prefix.
         for runner in runner_list:
             runner.calculate_instance_id(self._prefix)
 
--- a/github-runner-manager/src/github_runner_manager/manager/github_runner_manager.py
+++ b/github-runner-manager/src/github_runner_manager/manager/github_runner_manager.py
@@ -72,16 +72,7 @@ class GitHubRunnerManager:  # pragma: no cover
         Returns:
             Information on the runners.
         """
-        runner_list = self.github.get_runner_github_info(self._path)
-        runner_list = [
-            runner
-            for runner in runner_list
-            if InstanceID.name_has_prefix(self._prefix, runner.name)
-        ]
-        # Calculate instance_id, as it is not calculated by the GithubClient as it
-        # does not have information about the prefix.
-        for runner in runner_list:
-            runner.calculate_instance_id(self._prefix)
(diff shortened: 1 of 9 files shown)


github-runner-manager/src/github_runner_manager/manager/runner_manager.py
```diff
@@ -242,7 +242,26 @@ def cleanup(self) -> IssuedMetricEventsStats:
...
+        # We do not treat this case until better information is obtained from get_runners.
+        # get_runners returns unhealthy if the GitHub agent is not running and cloud init
+        # is running. If the unit is ACTIVE but stuck we could leave a runner forever.
+        cloud_instances_created = self.get_runners(cloud_states=[CloudRunnerState.CREATED])
+        cloud_instances_created_ids = [
+            cloud_instance.instance_id for cloud_instance in cloud_instances_created
+        ]
+        github_runners = self._github.get_runners([GitHubRunnerState.OFFLINE])
+        runners_to_delete = [
+            ghrunner
```
yhaliaw, line 260: `ghrunner` might not fit the naming convention for the team.
I think `gh_runner` might be better.

yhaliaw (approved):

github-runner-manager/src/github_runner_manager/manager/github_runner_manager.py
```diff
@@ -78,6 +78,10 @@ def get_runners(
             for runner in runner_list
             if InstanceID.name_has_prefix(self._prefix, runner.name)
         ]
+        # Calculate instance_id, as it is not calculated by the GithubClient as it
+        # does not have information about the prefix.
+        for runner in runner_list:
+            runner.calculate_instance_id(self._prefix)
```
jdkandersson, line 84: Should this be done by the class? E.g., what if someone forgets to call this method?

javierdelapuente: my plan is to have a "typedict" back in the GitHubClient, and in the GitHubRunnerManager to create the SelfHostedRunner class, without a name. But in this PR I just realized that we were still returning a dict from the GitHubClient, instead of the SelfHostedRunner

javierdelapuente: Thinking again about it, I am going to change it in this PR, I will pass the prefix to the client to filter

jdkandersson (approved):

github-runner-manager/src/github_runner_manager/manager/runner_manager.py
```diff
@@ -275,10 +275,9 @@ def _cleanup_github_offline_runners(self) -> None:
...
 
             # reactive runners.
-            # If there is no cloud runner, we do not remove  the GitHub runner.
-            # Although this could leave GitHub runners as offline forever, the
-            # other risk is to kill runners starting. Pending to analyze how to
-            # not leave runners in GitHub.
+            # If there is no cloud runner, we do not remove  the GitHub runner,
+            # as it can be a reactive runner being in the creation phase. The risk
+            # is that some offline runners could be left for a while in GitHub.
             if github_runner.instance_id not in cloud_instances_map:
```
yhaliaw, line 281: Is this the best way to check for reactive?
I thought we can check by naming convention or something.

javierdelapuente: at this point, the runners are reactive, as the previous check `if not github_runner.instance_id.reactive`.

yhaliaw: ok, resolved the comment.

yhaliaw (approved):

yanksyoon (approved):

PR #476, opened 2025-03-10
Fix backward compatibility in instance names

javierdelapuente:
Applicable spec: <link>

### Overview

Fix backward compatibility with previous unit names.

Otherwise, the old runners should be deleted manually when upgrading.

### Checklist

### DIFF
--- a/github-runner-manager/src/github_runner_manager/manager/models.py
+++ b/github-runner-manager/src/github_runner_manager/manager/models.py
@@ -43,12 +43,12 @@ class InstanceID:
            Name of the instance
         """
         if self.reactive is True:
-            reactive_indicator = "r-"
+            runner_type = "r-"
         elif self.reactive is False:
-            reactive_indicator = "n-"
+            runner_type = "n-"
         else:
-            reactive_indicator = ""
-        return f"{self.prefix}-{reactive_indicator}{self.suffix}"
+            runner_type = ""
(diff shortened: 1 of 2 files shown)


github-runner-manager/src/github_runner_manager/manager/models.py
```diff
@@ -42,7 +42,13 @@ def name(self) -> str:
         Returns:
            Name of the instance
         """
-        return f"{self.prefix}-{'r' if self.reactive else 'n'}-{self.suffix}"
+        if self.reactive is True:
+            reactive_indicator = "r-"
+        elif self.reactive is False:
+            reactive_indicator = "n-"
```
yhaliaw, line 48: Maybe `reactive_indicator` is not the best name as I feel "pre-spawned" is a better name than "non-reactive". In which case, thinking it as the type of runner might be better. So changing this variable to `runner_type` or something like that might be good.

If there is a name change. This can be follow-up in a different PR, as this is not critical.

yhaliaw (approved):

github-runner-manager/src/github_runner_manager/manager/models.py
javierdelapuente: let's put runner_type, it clearly looks better. Thanks!

jdkandersson (approved):

PR #492, opened 2025-03-21
Use of proxy for github runners

javierdelapuente:
Applicable spec: <link>

### Overview

In corporate environments, most of the communication is done through proxies. 

This PR adds several optional configuration options to work with proxies:
- manager-ssh-proxy-command
- runner-http-proxy
- use-runner-proxy-for-tmate

Both aproxy and socat will now be always installed in the image-builder. That way we save time (and possible connection issues).

If the `runner-http-proxy`  option is set, the runner will have that configuration for the proxy, instead of the configuration from the Juju model.

If `use-runner-proxy-for-tmate` is set to True, the connection from the runner to the tmate-ssh-server will be proxied using the runner proxy configuration.

The configuration option `manager-ssh-proxy-command` defines an optional ProxyCommand for the ssh connection from the manager to the runners.

For the tmate proxy, a new systemd service `tmate-proxy.service` will be added to the runner. However, it will not be started, in order not to interfere with ports used by actions run by the user of the runners. The `action-tmate` github action will start it when appropriate.

No integration tests are yet added for the new functionality, as they interfere with the openstack configuration and can be very dependent on infrastructure. Pending to reevaluate, as it would be nice to have some integration test for each feature.

Related to:
https://github.com/canonical/github-runner-image-builder-operator/pull/96
https://github.com/canonical/action-tmate/pull/11

**unit tests coverage**

The unit tests are excluding several files in "omit", which I believe are relevant to unit test. As so I have removed them from the omit list, and the coverage has decreased greatly.

### Checklist


github-runner-manager/src/github_runner_manager/configuration/base.py
```diff
@@ -131,12 +145,18 @@ class SSHDebugConnection(BaseModel):
         port: The SSH relay server port.
         rsa_fingerprint: The host SSH server public RSA key fingerprint.
         ed25519_fingerprint: The host SSH server public ed25519 key fingerprint.
+        use_runner_http_proxy: Whether to use runner proxy for the SSH connection.
+        local_http_proxy_host: Local host to use for proxying.
+        local_http_proxy_port: Local port to use for proxying.
```
cbartz, line 150:
```suggestion
        local_proxy_host: Local host to use for proxying.
        local_proxy_port: Local port to use for proxying.
```

iiuc its a general tcp proxy

cbartz (commented):
Some initial comments, lets have a call later.

github-runner-manager/pyproject.toml
```diff
@@ -39,16 +39,13 @@ skips = ["*/*test.py", "*/test_*.py", "*tests/*.py"]
...
-    "src/github_runner_manager/openstack_cloud/openstack_cloud.py",
-    "src/github_runner_manager/openstack_cloud/openstack_runner_manager.py",
     # Contains interface for calling repo policy compliance service. Tested in integration test
     # and end to end tests.
     "src/github_runner_manager/repo_policy_compliance_client.py",
 ]
 
 [tool.coverage.report]
-fail_under = 80
+fail_under = 73
```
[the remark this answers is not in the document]
javierdelapuente: it could be worse if I had also removed the repo policy file. I think we should not omit full files.

github-runner-manager/src/github_runner_manager/configuration/base.py
javierdelapuente: not sure, the proxy is http, but we use it to just tunnel a tcp connection....

config.yaml
```diff
@@ -128,3 +128,36 @@ options:
...
+      https://docs.fabfile.org/en/latest/api/connection.html#module-fabric.connection. It is equivalent
+      to ProxyCommand in ssh-config, but without a shell. For example, a value could be something like:
+      `sh -c "openssl s_client -quiet -connect 192.168.20.2:2222 -servername %h 2> /dev/null"`
+  runner-http-proxy:
+    type: string
+    default: ""
+    description: >-
+      The HTTP proxy value to use for the runners. If not specified, the model configuration option
+      juju-http-proxy will be used instead.
+  runner-https-proxy:
```
javierdelapuente, line 146: remove this config options, runner-https-proxy and runner-no-proxy

github-runner-manager/src/github_runner_manager/configuration/base.py
cbartz: but isn't 

```
socat -v TCP-LISTEN:{{ssh_debug_info.local_http_proxy_port}},fork,bind={{ssh_debug_info.local_http_proxy_host}} PROXY:{{runner_proxy_config.proxy_host}}:{{ssh_debug_info['host']}}:{{ssh_debug_info['port']}}{% if runner_proxy_config.proxy_port %},proxyport={{runner_proxy_config.proxy_port}}{% endif %}
```

creating a tunnel which listens on local_http_proxy_port , which is in theory independent of the http protocl? Shouldn't  the name not be rather `local_socat_tunnel_port` ?

At least for me the line `TMATE_SERVER_HOST={{ssh_debug_info.local_http_proxy_host}}` in env.j2 reads a bit confusing, as I would not expect the TMATE_SERVER_HOST to be a http proxy.

javierdelapuente: Updated to remove the http prefix

config.yaml
```diff
@@ -128,3 +128,23 @@ options:
...
       is supported at the moment. A valid configuration looks like "m1.small:small".
+  manager-ssh-proxy-command:
+    type: string
+    default: ""
+    description: >-
+      Optional ProxyCommand used by the ssh connection which is used to access the virtual
+      machine runners from the runner manager. It is equivalent to ProxyCommand in ssh-config,
+      but without a shell. For example, a value could be something like:
+      `bash -c "openssl s_client -quiet -connect 192.168.20.2:2222 -servername %h 2> /dev/null"`
+  runner-http-proxy:
```
yhaliaw, line 139: Does there need to be no proxy for runner?
If the runner is in a different environment as the charm, it would likely have a different set of no_proxy as well.

github-runner-manager/src/github_runner_manager/openstack_cloud/openstack_cloud.py
```diff
@@ -148,19 +148,28 @@ class OpenstackCloud:
...
+        proxy_command: str | None = None,
+    ):
         """Create the object.
 
         Args:
             credentials: The OpenStack authorization information.
             prefix: Prefix attached to names of resource managed by this instance. Used for
                 identifying which resource belongs to this instance.
             system_user: The system user to own the key files.
+            proxy_command: gateway argument for fabric Connection. Similar to ProxyCommand in
```
yhaliaw, line 165: The `gateway` should be in caps? Since it is the start of a sentence.
Maybe `The gateway argument ...`.

tests/integration/conftest.py
```diff
@@ -368,7 +368,7 @@ async def image_builder_fixture(
             "github-runner-image-builder",
             application_name=application_name,
             channel="latest/edge",
-            revision=61,
+            revision=68,
```
yhaliaw, line 371: We might want to have this as a constant. Maybe in it own file and have renovate to upgrade this as well.

yhaliaw: Not for this PR. Just a general comment. But we should record this somewhere.

javierdelapuente: that is a good idea... maybe even not setting the revision could do it

config.yaml
javierdelapuente: I discussed this yesterday with @cbartz. The runner proxy is currently just for aproxy, and aproxy does not make use of https-proxy or no-proxy. As they are not used, we removed it from the PR, to have fewer configuration options (if at some point they are used or are exposed to the runner, we could add them back).

yhaliaw: The close thing we would able to do is resolve the NO_PROXY to IPs and add them to the `private-ips` defined in the nftables rules: https://github.com/canonical/github-runner-image-builder-operator/blob/4d007e1ba95f030cb569198b0b54110b8e3183f3/app/src/github_runner_image_builder/templates/cloud-init.sh.j2#L31

This is a gap in aproxy feature set. I think this is fine for now.

tests/integration/conftest.py
yhaliaw: I would like it as PR that we can merge.
Else we will have to deal with workflows a PR we are working on, then suddenly unable to pass tests due to revision upgrade in a separate charm.

javierdelapuente: put back and we can discuss it elsewehere

docs/changelog.md
```diff
@@ -1,5 +1,14 @@
 # Changelog
 
+### 2025-03-27
+
+- Add several proxy configurations options for the manager and the runners:
```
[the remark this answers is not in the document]
javierdelapuente: explained a bit more

tests/integration/conftest.py
javierdelapuente: let's leave it for a future PR, and we can think how to do it everywhere.

github-runner-manager/src/github_runner_manager/configuration/base.py
cbartz: I think you removed proxy instead of http

javierdelapuente: ouch

tests/unit/test_charm_state.py
```diff
@@ -1061,3 +1029,88 @@ def test_charm_state__log_prev_state_redacts_sensitive_information(
 
     assert mock_charm_state_data["charm_config"]["token"] not in caplog.text
     assert charm_state.SENSITIVE_PLACEHOLDER in caplog.text
+
+
+@pytest.mark.parametrize(
+    "juju_http, juju_https, juju_no_proxy, runner_http, use_aproxy,"
+    "expected_proxy, expected_runner_proxy",
+    [
+        ("", "", "", "", False, ProxyConfig(), ProxyConfig()),
```
cbartz, line 1038: pytest.param with descriptive id would be nice for easier readability

cbartz (approved):
LGTM, small comment

PR #510, opened 2025-04-09
If a job is in progress, go for the following one without a new reactive process

javierdelapuente:
Applicable spec: <link>

### Overview

Currently reactive process exit if the message they get is for a process "in progress". That means we lose the reactive process and we have to wait for the next reconciliation loop (which is not very effective). If the message from the queue is in progress, just ack it and get another one.

Logs changed to INFO, as they are very verbose.

### Checklist


github-runner-manager/src/github_runner_manager/reactive/consumer.py
```diff
@@ -26,6 +26,8 @@
 
 Labels = set[str]
 
+END_PROCESSING_PAYLOAD = "__END__"
```
cbartz, line 29: Are we ever seeing this? Or do you plan to insert it in the queue?

cbartz (commented):
Looks good, one question

github-runner-manager/src/github_runner_manager/reactive/consumer.py
javierdelapuente: Not in principle. It is basically for testing.

cbartz: Could you add an explanatory comment? When I first read this, I thought this is some internal constant coming from kombu

javierdelapuente: added a comment for it

cbartz (approved):

yhaliaw (approved):

PR #511, opened 2025-04-09
Refactor platform interface. Use of metadata for tagging runners.

javierdelapuente:
Applicable spec: <link>

### Overview

Several things are tackled in this PR:
- New RunnerMetadata class to have "metadata" about a runner, that is platform name. Mainly id about the runner and the url.
 - Put metadata in the instance, to identify prefix, platform provider (github or jobmanager in the following pr), runner id and possible url for the platform provider.
 - Get that metadata every time one runnes is obtained from openstack, as we may need to use it to select the provider when there are more than one.
 - Put all the classes used in the cloud cloud_runner_manager module (the cloud api) in that file, as they are scattered in metrics modules, but it is really part of the interface. Maybe we can find a better place/naming in a following PR.

As it is now, there is an overlap in InstanceID and RunnerMetadata. Up to now, all matching between instances in cloud and platform are done with the InstanceID, but from now, the RunnerMetadata will be used in more places. Pending to review the uses, and maybe integrating both concepts of removing the InstanceID importance.

The following PR will create a new platform provider that will build upon the common interface for platforms.

### Rationale

This is another refactor for adapting the github-runner to be able to use both platform provider, that is, github and the jobmanager. To know whether a runner uses one provider of the other, we will store information in the openstack metadata. For the jobmanager we would also put there the URL of the job (the runner), so we can use it to get the token, health checks, metrics...

### Checklist

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -1,5 +1,9 @@
 # Changelog
 
+### 2025-04-09
+
+- Remove update-dependencies action. This actions is not needed for external cloud providers.
+
 ### 2025-03-27
 
 - Add proxy configuration options for charm to facilitate its use in corporate environments.
--- a/github-runner-manager/src/github_runner_manager/manager/models.py
+++ b/github-runner-manager/src/github_runner_manager/manager/models.py
@@ -157,7 +157,7 @@ class InstanceID:
 
 @dataclass
 class RunnerMetadata:
-    """This class container information about the runner and the platform it runs in.
+    """This class contains information about the runner and the platform it runs in.
 
     The information in this class is needed to link cloud runners with their
     platform ones.
--- a/github-runner-manager/src/github_runner_manager/manager/models.py
+++ b/github-runner-manager/src/github_runner_manager/manager/models.py
@@ -4,7 +4,7 @@
 """Module containing the main classes for business logic."""
 
 import secrets
-from dataclasses import dataclass
+from dataclasses import asdict, dataclass
 
 INSTANCE_SUFFIX_LENGTH = 12
 
@@ -153,3 +153,26 @@ class InstanceID:
             String with the representation of the InstanceID.
         """
         return f"InstanceID({self.name!r})"
+
+
+@dataclass
+class RunnerMetadata:
+    """TODO.
+
+    Attributes:
+        platform_name: TODO
+        runner_id: TODO
+        url: TODO for runner?? for platform? what does this mean??
+    """
+
+    platform_name: str = "github"
+    runner_id: str | None = None
+    url: str | None = None
+
+    def as_dict(self) -> dict[str, str]:
+        """TODO.
+
+        Returns:
+            TODO.
+        """
+        return {k: v for k, v in asdict(self).items() if v is not None}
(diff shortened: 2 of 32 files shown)


github-runner-manager/src/github_runner_manager/manager/models.py
```diff
@@ -153,3 +153,29 @@ def __repr__(self) -> str:
             String with the representation of the InstanceID.
         """
         return f"InstanceID({self.name!r})"
+
+
+@dataclass
+class RunnerMetadata:
+    """This class container information about the runner and the platform it runs in.
```
cbartz, line 160: nit
```suggestion
    """This class contains information about the runner and the platform it runs in.
```

github-runner-manager/src/github_runner_manager/manager/models.py
```diff
@@ -153,3 +153,29 @@ def __repr__(self) -> str:
...
+
+    The information in this class is needed to link cloud runners with their
+    platform ones.
+
+    Attributes:
+        platform_name: Platform name where the runner resides (github, jobmanager...)
+        runner_id: Id of the runner in the platform
+        url: URL for the runner.
```
cbartz, line 168: Can this be more precise? This is the job url?

cbartz (approved):
LGTM, have some nit comments.

Can you please fill out the Rationale? I think I understood the necessity but for people without context the reason for the changes may be hard to follow. Thanks.

github-runner-manager/src/github_runner_manager/openstack_cloud/openstack_cloud.py
```diff
@@ -175,11 +179,18 @@ def __init__(
...
     # added.
     def launch_instance(  # pylint: disable=too-many-arguments, too-many-positional-arguments
-        self, instance_id: InstanceID, image: str, flavor: str, network: str, cloud_init: str
+        self,
+        metadata: RunnerMetadata,
+        instance_id: InstanceID,
+        image: str,
+        flavor: str,
+        network: str,
+        cloud_init: str,
```
yhaliaw, line 188: I think we should re-organize the arguments, since there are 7 of them now.
Referring to the `too-many-arugments` lint ignore.

docs/changelog.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+### 2025-04-09
+
+- Remove update-dependencies action. This actions is not needed for external cloud providers.
```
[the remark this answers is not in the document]
javierdelapuente: This was done in the previous PR (and I forgot to add it :(, even as was suggested in a comment)

github-runner-manager/src/github_runner_manager/openstack_cloud/openstack_cloud.py
javierdelapuente: Thanks for the comment. I have used another class that already had exactly those fields (I also had to put it in another file and make it independent to avoid cyclic imports and make it part of the internal api of the client)

yhaliaw (approved):

PR #530, opened 2025-04-17
Feat/app integration

yhaliaw:
Applicable spec: ISD-202

### Overview

The github-runner-manager service is enabled in the charm now.

### Rationale

This allow us to check logs of the service, to monitor if the service is stable.

### Checklist

The changelog is not updated, since the usage of service does not impact the users.


src/charm.py
```diff
@@ -247,6 +250,12 @@ def _common_install_code(self) -> bool:
             logger.error("Failed to setup runner manager user")
             raise
 
+        try:
+            manager_service.install_package()
```
cbartz, line 254: What about upgrading? How is this handled?

tests/unit/test_manager_service.py
```diff
@@ -0,0 +1,142 @@
...
+    assert manager_service._SERVICE_SETUP_ERROR_MESSAGE in str(err.value)
+
+
+def test_install_package_success(mock_execute_command: MagicMock):
+    """
+    arrange: Mock execute command.
+    act: Run install_package.
+    assert: No errors raised.
+    """
+    manager_service.install_package()
```
cbartz, line 126: Is this test really adding value?

src/charm.py
```diff
@@ -255,6 +264,21 @@ def _common_install_code(self) -> bool:
 
         return True
 
+    def _setup_service(self, state: CharmState) -> None:
```
cbartz, line 267: nit: this method is called below, according to our standard it should appear lower in the file

cbartz (commented):
Thank you!

src/charm.py
yhaliaw: It is a local python package.
Installing it with code changes would update the installation.

The `_common_install_code` function is called on charm upgrade as well. So it should cover the charm upgrade event.

src/charm.py
yhaliaw: Updated.

src/charm.py
cbartz: So due to the `--ignore-installed` flag it would just install without caring about a previous installation?

yhaliaw: I have updated the `install_package` method to uninstall the packages prior to installing them.

src/manager_service.py
```diff
@@ -54,8 +59,22 @@ def install_package() -> None:
...
     """
+    logger.info("Upgrading pip")
     try:
         execute_command(["python3", "-m", "pip", "install", "--upgrade", "pip"])
+    except SubprocessError as err:
+        raise RunnerManagerApplicationInstallError(_INSTALL_ERROR_MESSAGE) from err
+
+    logger.info("Uninstalling previous version of packages")
+    try:
+        execute_command(["python3", "-m", "pip", "uninstall", GITHUB_RUNNER_MANAGER_PACKAGE])
```
cbartz, line 70: Maybe keep in mind for adaptions later, I think having an abstraction to interact with pip (similar as in https://github.com/canonical/github-runner-image-builder-operator/blob/main/src/pipx.py ) would make it the code easier testable and less coupled.

src/charm.py
gregory-schiano: In the upgrade scenario, I guess the service needs to be stopped before "uninstall / install"

src/manager_service.py
gregory-schiano: I'd use `pipx` instead of pip python module, but making a library is a bit overkill since the pip commands are only executed in this method, and nowhere else.

src/manager_service.py
```diff
@@ -0,0 +1,146 @@
...
+
+def _setup_config_file(state: CharmState, app_name: str, unit_name: str) -> Path:
+    """Write the configuration to file.
+
+    Args:
+        state: The charm state.
+        app_name: The Juju application name.
+        unit_name: The Juju unit.
+    """
+    config = create_application_configuration(state, app_name, unit_name)
```
gregory-schiano, line 119: Instead of passing the state from the charm to this module to the factory module, I'd rather generate the configuration in the charm and pass the configuration to this module

src/manager_service.py
```diff
@@ -0,0 +1,146 @@
...
+
+def _setup_service_file(config_file: Path) -> None:
+    """Configure the systemd service.
+
+    Args:
+        config_file: The configuration file for the service.
+    """
+    jinja = jinja2.Environment(loader=jinja2.FileSystemLoader("templates"), autoescape=True)
+
+    service_file_content = jinja.get_template("github-runner-manager.service.j2").render(
```
gregory-schiano, line 138: Since the template is trivial (no complex logic) and is pretty short, it's probably more readable to use a multiline f-string instead

src/manager_service.py
yhaliaw: For this, the package would need to be installed to all users.
Which pipx supports with this PR: https://github.com/pypa/pipx/pull/1281
However, version is not on apt yet.

I added a TODO comment for now:
```
# TODO: Use pipx over pip once the version that supports `pipx install --global` lands on apt.
```

tests/unit/test_manager_service.py
yhaliaw: It is checking the function is not raise unintended exceptions

yhaliaw: I have removed the test.

src/manager_service.py
```diff
@@ -56,7 +56,7 @@ def setup(state: CharmState, app_name: str, unit_name: str) -> None:
     _setup_service_file(config_file)
     _enable_service()
 
-
+# TODO: Use pipx over pip once the version that supports `pipx install --global` lands on apt.
```
cbartz, line 59: Please don't add TODOs to the code for things which will not be tackled immediately. These TODO's generally tend to live forever and confuse other developers.

yhaliaw: Added

PR #534, opened 2025-04-23
Remove SSH creation health checks

javierdelapuente:
Applicable spec: <link>

### Overview

This PR starts with the removal of the ssh health checks, specifically, on creating a new server, the machine is not accessed using ssh. Instead, the platform provider is queried to check if the server is online (of it is deletable).

For that:
- A new method was added in the Platform providers called get_runner_health. This method returns the information about the health of a runner. Probably a method similar to this one (but affecting more than one runner) should be used when removing the ssh health checks in the main reconcile.
- The RunnerManager calls the cloud manager to create the openstack machine and then waits until the platform provider replies that the runner is healthy. The waiting times (and number of retries) are in the variable RUNNER_CREATION_WAITING_TIMES.

Because of this change, up to 5 new requests can be made to the GitHub API. To compensate, I have removed 5 calls when checking if the job has been picked up. This has advantages and disadvantages. As the advantage, reactive runner will be spawned faster when there is a lot of pressure (the process will die sooner and the message requeued faster).

A few things pending (and out of scope for this PR):
- The PlatformProvider interface still has a few rough edges. A small cleanup will be done in following PRs. For example:
  - It is not clear the use of the SelfHostedRunner class (probably it is enough with the health check status).
  - The exceptions raised in the API are not totally clear, and places in different files.
- Actually, no waiting should be done on runner creation, but that is outside of scope for this PR. For that:
  - Runners can be deleted in the reconcile loop if they are offline after a long time since creation.
  - The waiting should only exist in the case of reactive runners (and a way without waiting should be thought).
- Currently, runners are flushed in reactive mode when there are no messages. It could be interesting to rethink the idea, and flush runners periodically (even for non-reactive).

 
- ### Rationale

### Checklist

### DIFF
--- a/github-runner-manager/src/github_runner_manager/github_client.py
+++ b/github-runner-manager/src/github_runner_manager/github_client.py
@@ -97,26 +97,20 @@ class GithubClient:
         self._client = GhApi(token=self._token)
 
     @catch_http_errors
-    def get_runner_info(self, path: GitHubPath, prefix: str, runner_id: int) -> SelfHostedRunner:
-        """TODO.
-
-        https://docs.github.com/en/rest/actions/self-hosted-runners?
-        apiVersion=2022-11-28#get-a-self-hosted-runner-for-an-organization
-        /orgs/{org}/actions/runners/{runner_id}
-        https://docs.github.com/en/rest/actions/self-hosted-runners?
-        apiVersion=2022-11-28#get-a-self-hosted-runner-for-a-repository
-        /repos/{owner}/{repo}/actions/runners/{runner_id}
+    def get_runner(self, path: GitHubPath, prefix: str, runner_id: int) -> SelfHostedRunner:
+        """Get a specific self-hosted runner information under a repo or org.
 
         Args:
-            path: TODO
-            prefix: TODO
-            runner_id: TODO
+            path: GitHub repository path in the format '<owner>/<repo>', or the GitHub organization
+                name.
+            prefix: Build the InstanceID with this prefix.
+            runner_id: Runner id to get the self hosted runner.
 
         Raises:
-            GithubRunnerNotFoundError: TODO
+            GithubRunnerNotFoundError: If the runner is not found.
 
         Returns:
-            TODO
+            The information for the requested runner.
         """
         try:
             if isinstance(path, GitHubRepo):
@@ -133,8 +127,8 @@ class GithubClient:
         return SelfHostedRunner.build_from_github(raw_runner, instance_id)
 
     @catch_http_errors
-    def get_runner_github_info(self, path: GitHubPath, prefix: str) -> list[SelfHostedRunner]:
-        """Get runner information on GitHub under a repo or org.
+    def list_runners(self, path: GitHubPath, prefix: str) -> list[SelfHostedRunner]:
+        """Get all runners information on GitHub under a repo or org.
 
         Args:
             path: GitHub repository path in the format '<owner>/<repo>', or the GitHub organization
--- a/github-runner-manager/src/github_runner_manager/github_client.py
+++ b/github-runner-manager/src/github_runner_manager/github_client.py
@@ -13,6 +13,9 @@ from typing import Callable, ParamSpec, TypeVar
 from urllib.error import HTTPError
 
 import requests
+
+# HTTP404NotFoundError is not found by pylint
+from fastcore.net import HTTP404NotFoundError  # pylint: disable=no-name-in-module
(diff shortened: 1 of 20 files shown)


github-runner-manager/src/github_runner_manager/manager/runner_manager.py
```diff
@@ -474,16 +483,59 @@ def _create_runner(args: _CreateRunnerArgs) -> InstanceID:
...
+                runner_health = platform_provider.get_runner_health(
+                    metadata=metadata, instance_id=instance_id
+                )
+            except PlatformApiError as exc:
+                logger.error("Error getting the runner health: %s", exc)
+                continue
+            if runner_health.online or runner_health.deletable:
+                break
+        else:
+            raise RunnerError(f"Runner {instance_id} did not get online")
```
cbartz, line 541: Just to understand. In this case, the openstack server would be cleaned up next reconcile? So by increasing the reconcile interval, we would still be able to spawn runners (e.g. for really slow clouds)?

github-runner-manager/src/github_runner_manager/reactive/consumer.py
```diff
@@ -237,7 +237,8 @@ def _spawn_runner(
         logger.error("Failed to spawn a runner. Will reject the message.")
         msg.reject(requeue=True)
         return
-    for _ in range(10):
+
+    for _ in range(5):
```
cbartz, line 241: This is probably to keep the amount of github api calls equal?

We would wait 2,5mins instead of 5mins for a job to be picked up by Github. I guess that this is still fine, but I think we should update the changelog to mention this (so we can refer back later quickly to this revision in case we see any issues out of it).

cbartz (commented):
Thanks! Reviewed up to tests.

github-runner-manager/src/github_runner_manager/manager/runner_manager.py
javierdelapuente: This pr does not change the functionality much, this code is quite similar to what was before. 

If the wait failed, it tries to remove the runner (from the platform provider). If it is not busy, it will be removed from the platform provider and cleaned up in the next reconcile loop.

The next step would be to remove this waiting in here (and also remove the waiting for the active runner), but it was not tackled in this pr (and it would be nice to remove all ssh checks before that)

github-runner-manager/src/github_runner_manager/reactive/consumer.py
javierdelapuente: I though a lot about the effects of this change. But I am not sure if it is something good or bad yet :). 

I will add it to the changelog. Thanks for the tip.

github-runner-manager/tests/unit/platform/test_github_provider.py
```diff
@@ -0,0 +1,117 @@
...
+)
+from github_runner_manager.types_.github import GitHubRunnerStatus, SelfHostedRunner
+
+
+def _params_test_get_runner_health():
+    """Parameterized data for test_get_runner_health."""
+    prefix = "unit-0"
+    runner_id = 3
+    metadata = RunnerMetadata(platform_name="github", runner_id=str(runner_id))
+    instance_id = InstanceID.build(prefix=prefix)
```
cbartz, line 25: nit: can you add "instance_id,metadata,self_hosted_runner,expected_online,expected_busy,expected_deletable" as a comment here to quickly understand the params without having to scroll down? Thank you.

cbartz (approved):

github-runner-manager/src/github_runner_manager/github_client.py
```diff
@@ -89,8 +97,38 @@ def __init__(self, token: str):
...
+        try:
+            if isinstance(path, GitHubRepo):
+                raw_runner = self._client.actions.get_self_hosted_runner_for_repo(
+                    path.owner, path.repo, runner_id
+                )
+            else:
+                raw_runner = self._client.actions.get_self_hosted_runner_for_org(
+                    path.org, runner_id
+                )
+        except HTTP404NotFoundError as err:
```
yhaliaw, line 124: Is this the only possible error here?
What about other status code or connection error?

yhaliaw (approved):
I left a comment on the edge case.
I do not expect the edge case to occur a lot, so you can cover it in a followup if needed.

github-runner-manager/src/github_runner_manager/github_client.py
javierdelapuente: That is already catched in the catch_http_errors decorator.

PR #545, opened 2025-05-07
[FIX] Add HTTP proxies setting for installation

yhaliaw:
### Overview

In the old LXD mode code, there HTTP proxy env variables are setup.

https://github.com/canonical/github-runner-operator/blob/3f99dd0fd608acdc9e8544a5669975fa7ad1abeb/src/runner_manager.py#L92

This PR adds this back to make the APT and pip installation work in proxied environments.

### Rationale

The github-runner-manager service is unable to be installed in proxied environments if this is not added.

### Checklist


src/charm_state.py
```diff
@@ -586,7 +586,7 @@ def from_charm(cls, charm: CharmBase) -> "OpenstackRunnerConfig":
         )
 
 
-def _build_proxy_config_from_charm() -> "ProxyConfig":
+def build_proxy_config_from_charm() -> "ProxyConfig":
```
cbartz, line 589: As we move from externalising the app (so not using as library anymore), I think we should no longer import these classes from the app path. Otherwise we are using the github-runner-manager as library and application at the same time, which could lead to confusion and weird mistakes in the future. The code of the app should be independent from the code of the charm.

WDYT @javierdelapuente ?

I think we should move to something like `ProxyConfig.from_charm` with ProxyConfig living in this module.

cbartz (approved):
LGTM, see my comment, you could already start moving ProxyConfig to the charm code.

src/charm_state.py
yhaliaw: I will do this in a follow up PR.
Thanks.

javierdelapuente (approved):

PR #561, opened 2025-05-23
feat: github-runner-manager service for reconcilations

yhaliaw:
Applicable spec: ISD-202

### Overview

The reconcile and flush runner of the github-runner charm is moved to using the github-runner-manager service.

The reconcile-interval is significantly reduce as now it is not tied to a juju event. It is now a background process that continuously reconciles. If there is GitHub rate-limit issues, this can be increase else it should be kept low.

### Rationale

Moving the business logic to a workload allows move flexible approach to features. E.g., easier to add prometheus metric, etc.

### Juju Events Changes

The `flush-runners` action no longer reconciles before returning.
The `reconcile-runners` action is removed. Since the reconcile will be ran in the background almost continuously.

Since the charm code no longer performs reconcile, reconciliation is removed from various juju events hooks.
e.g., image relation changed.

### Checklist


github-runner-manager/src/github_runner_manager/reconcile_service.py
```diff
@@ -3,24 +3,50 @@
...
-            logger.info("Acquired the lock for reconciling")
-            sleep(10)
-        logger.info("Released the lock")
+            logger.info(RECONCILE_START_MSG)
+            runner_scaler = get_runner_scaler(app_config)
+            delta = runner_scaler.reconcile()
+            logger.info("Change in number of runner after reconcile: %s", delta)
+        logger.info(RECONCILE_END_MSG)
+        # TODO: Think about this.
+        sleep(30)
```
yhaliaw, line 52: Bartz: tie this to config option.

docs/changelog.md
```diff
@@ -1,5 +1,12 @@
 # Changelog
 
+### 2025-06-04
+
+- Greatly reduce the reconcile-interval configuration. This is the interval between reconciling
```
cbartz, line 5: Do you mean by greatly the reduce from 10 to 5 minutes?

src/charm.py
```diff
@@ -397,83 +329,35 @@ def _on_check_runners_action(self, event: ActionEvent) -> None:
...
+        """Handle the action of flushing all runner and reconciling afterwards."""
+        self._manager_client.flush_runner()
 
     @catch_charm_errors
     def _on_update_status(self, _: UpdateStatusEvent) -> None:
         """Handle the update of charm status."""
-        self._ensure_reconcile_timer_is_active()
         self._log_juju_processes()
 
     def _setup_service(self, state: CharmState) -> None:
```
cbartz, line 345: will this method also be called after a unit reboot, so I guess the question would be if we can expect `config-changed` to be issued after a reboot. Otherwise we would need to set the service up on start (with a safeguard to check if image relation is already set up).

src/charm.py
```diff
@@ -513,25 +397,7 @@ def _log_charm_status(self) -> None:
...
-        """
-        self.unit.status = MaintenanceStatus("Reconciling runners")
-        try:
-            runner_scaler.reconcile()
-        except ReconcileError:
-            logger.exception(FAILED_TO_RECONCILE_RUNNERS_MSG)
-            self.unit.status = ActiveStatus(ACTIVE_STATUS_RECONCILIATION_FAILED_MSG)
-        else:
-            self.unit.status = ActiveStatus()
+        self._manager_client.flush_runner(busy=True)
```
cbartz, line 400: Wouldn't we also need to stop the service when we stop it to avoid new resources being created?

src/charm_state.py
```diff
@@ -395,15 +395,15 @@ def check_reconcile_interval(cls, reconcile_interval: int) -> int:
         Returns:
             The validated reconcile_interval value.
         """
-        # The EventTimer class sets a timeout of `reconcile_interval` - 1.
-        # Therefore the `reconcile_interval` must be at least 2.
-        if reconcile_interval < 2:
+        # The reconcile_interval should be greater than 0.
```
cbartz, line 398:
```suggestion
        # The reconcile_interval should be at least 1 minute.
```

src/event_timer.py
cbartz, line 1: :heart: 

Please also remove the ninja templates (`dispatch-.*`) under `templates`

tests/integration/test_jobmanager.py
```diff
@@ -102,16 +101,18 @@ async def app_fixture(
             PATH_CONFIG_NAME: "",
             BASE_VIRTUAL_MACHINES_CONFIG_NAME: "0",
             MAX_TOTAL_VIRTUAL_MACHINES_CONFIG_NAME: "1",
+            # Disable reconcile since the test controls when the reconcile happens.
```
cbartz, line 104:
```suggestion
            # Disable reconcile by setting it to a long time period since the test controls when the reconcile happens.
```

cbartz (commented):
Thank you! Looks good so far, have some suggestions/questions.

javierdelapuente (commented):
Looks good. A couple of minor comments

src/manager_client.py
```diff
@@ -123,11 +130,45 @@ def check_runner(self) -> dict[str, str]:
...
+    def wait_till_ready(self) -> None:
+        """Wait till the runner manager service is ready for requests."""
+        for _ in range(5):
+            try:
+                self.health_check()
+            except RunnerManagerServiceNotReadyError:
+                pass
+            else:
+                return
+            sleep(60)
```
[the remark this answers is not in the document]
yhaliaw: You are right that the startup should be fairly fast.
I will reduce the wait period.

github-runner-manager/src/github_runner_manager/reconcile_service.py
```diff
@@ -3,24 +3,66 @@
...
 from threading import Lock
 from time import sleep
 
 from github_runner_manager.configuration import ApplicationConfiguration
+from github_runner_manager.configuration.base import UserInfo
+from github_runner_manager.manager.runner_scaler import RunnerScaler
 
 logger = logging.getLogger(__name__)
 
+RECONCILE_ID_FILE = Path("/var/log/reconcile.id")
```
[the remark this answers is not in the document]
yhaliaw: This is used in integration tests right now.
I would like to avoid inter-test dependencies between the metric test and all other tests.

I think we can think about adding something to track the reconcile in the metric. But that would be in the metric story.

src/charm.py
yhaliaw: The systemd service is set to `enable`. So systemd should start the service on machine reboot.

In the case, of only restarting the juju agent, the systemd service would not be stop in the first place.

yhaliaw: I will test this, just to make sure the behavior is as expected.

src/charm_state.py
yhaliaw: Updated.

src/event_timer.py
yhaliaw: Done.

tests/integration/test_jobmanager.py
yhaliaw: Updated.

src/charm.py
yhaliaw: From manual testing, upon machine reboot  the service does gets started.
Restarting the juju agent with systemctl does not affect the service.

cbartz: and what happens if we remove the image relation? we should disable the service then I suppose?

docs/changelog.md
```diff
@@ -2,9 +2,9 @@
 
 ### 2025-06-04
 
-- Greatly reduce the reconcile-interval configuration. This is the interval between reconciling 
-the current and intended number of runners. The value should be kept low, unless Github API rate 
-limiting is encountered.
+- Reduce the reconcile-interval configuration from 10 minutes to 5 minutes. This is the interval 
+between reconciling the current and intended number of runners. The value should be kept low, 
+unless Github API rate limiting is encountered.
 - Removed the reconcile-runners juju action.
```
cbartz, line 8:
```suggestion
- Removed the reconcile-runners Juju action.
```

make vale happy ;-)

src/charm_state.py
```diff
@@ -395,8 +395,10 @@ def check_reconcile_interval(cls, reconcile_interval: int) -> int:
         Returns:
             The validated reconcile_interval value.
         """
-        # The reconcile_interval should be greater than 0.
-        if reconcile_interval < 1:
+        # The reconcile_interval should be at least 2.
+        # Due to possible race condition with the message acknowledgement with job status checking
+        # in reactive process.
+        if reconcile_interval <= 1:
```
cbartz, line 401: The code indicates it should be "at least 1" ?

src/manager_service.py
```diff
@@ -110,6 +107,24 @@ def install_package() -> None:
...
+    """Stop the GitHub runner manager service.
+
+    Raises:
+        RunnerManagerApplicationStopError: Failed to stop the service.
+    """
+    try:
+        if systemd.service_running(GITHUB_RUNNER_MANAGER_SERVICE_NAME):
+            systemd.service_stop(GITHUB_RUNNER_MANAGER_SERVICE_NAME)
+    except SystemdError as err:
+        raise RunnerManagerApplicationStopError(_SERVICE_STOP_ERROR_MESSAGE) from err
```
cbartz, line 125: Can you add a unit test for this behaviour?

src/charm_state.py
yhaliaw: The code should be if the reconcile_interval is 1 or less then raise error.

I think "at least 2" includes 2 right?

cbartz: You are right, as we cannot give fractions of minutes . I think `reconcile_interval < 2` would be more explicit in this case.

src/manager_service.py
```diff
@@ -151,6 +175,7 @@ def _setup_service_file(config_file: Path, log_file: Path) -> None:
         config_file: The configuration file for the service.
         log_file: The file location to store the logs.
     """
+    python_path = Path(os.getcwd()) / "venv"
     service_file_content = textwrap.dedent(
         f"""\
```
yanksyoon, line 180: Perhaps we can add this to the systemd service environment declaration, or perhaps a configuration directory might be the best practice according to this [stack overflow](https://serverfault.com/questions/413397/how-to-set-environment-variable-in-systemd-service) answer.

So perhaps something like:
```/etc/systemd/system/github-runner-manager.conf
PYTHONPATH={str(python_path)}
...<other configuration env vars that might be required>
```

and in the service unit file
```diff
+ EnvironmentFile=/etc/systemd/system/github-runner-manager.conf
```

The best case would be to rename the service dir `/etc/systemd/system/github-runner-manager.d`
and have the service unit file as well as the configuration in the same directory.

yanksyoon (approved):
Reviewed until charm.py, nice code deletions!

src/charm.py
```diff
@@ -397,83 +340,35 @@ def _on_check_runners_action(self, event: ActionEvent) -> None:
...
+                has errors.
         """
         try:
             manager_service.setup(state, self.app.name, self.unit.name)
         except RunnerManagerApplicationError:
             logging.exception("Unable to setup the github-runner-manager service")
-            # Not re-raising error for until the github-runner-manager service replaces the
-            # library.
+            raise
+        self._manager_client.wait_till_ready()
```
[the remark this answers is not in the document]
yhaliaw: The function is hardcoded to try for 8 times in 2 mins right now.
I do not think this function benefits from having these setting as arguments, since the startup time should be fairly short reasonable size juju machines.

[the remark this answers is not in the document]
yhaliaw: The startup is for a flask server. Generally, this will be fairly fast for all modern machines.

src/manager_service.py
```diff
@@ -51,7 +53,15 @@ def setup(state: CharmState, app_name: str, unit_name: str) -> None:
...
         unit_name: The Juju unit.
+
+    Raises:
+        RunnerManagerApplicationStartError: Setup of the runner manager service has failed.
     """
+    try:
+        if systemd.service_running(GITHUB_RUNNER_MANAGER_SERVICE_NAME):
+            systemd.service_stop(GITHUB_RUNNER_MANAGER_SERVICE_NAME)
+    except SystemdError as err:
+        raise RunnerManagerApplicationStartError(_SERVICE_SETUP_ERROR_MESSAGE) from err
```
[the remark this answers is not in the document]
yhaliaw: I did thought about this.
The difference is in the exception raised.
It would be a bit weird from the call viewpoint that this function that is trying to start the service raises a `RunnerManagerApplicationStopError`.

src/manager_service.py
yhaliaw: This is a tmp hack, to make the current reactive process work.
This PYTHONPATH related code will be removed with the reactive process re-design.

PR #567, opened 2025-06-11
feat: pre-spawned mode for jobmanager platform

cbartz:
Applicable spec: <link>

### Overview

Support prespawning runners with the jobmanager platform.

Note that the number of changed files is significant, but most of them are automatically generated (under `jobmanager/client`)

Also:

- add @yanksyoon to codeowners
- fix a bug that adding/leaving mongodb integration did not reconfigure the application.

### Rationale

Prespawned mode can provide faster job pickup.

### Juju Events Changes

- mongodb _relation_broken, database_endpoints_changed,database_created events now trigger a github runner manager application restart

### Module Changes

- charm : Add back mongodb events
- charm_state: Use path to detect whether jobmanager or github platform should be used (depending if an http url is configured or not)
- modules in manager application are adapted to use an updated api of the jobmanager, which basically behaves as github
- jobmanager client code has been regenerated 

### Library Changes
n/a

### Checklist

No documentation update, as the jobmanager platform is still considered as "hidden" feature.


.github/workflows/integration_test.yaml
```diff
@@ -12,22 +12,6 @@ concurrency:
...
-    with:
-      juju-channel: 3.1/stable
-      pre-run-script: scripts/setup-integration-tests.sh
-      provider: lxd
-      test-tox-env: integration-juju3.1
-      modules: '["test_charm_scheduled_events", "test_debug_ssh", "test_charm_upgrade"]'
-      extra-arguments: '-m openstack --log-format="%(asctime)s %(levelname)s %(message)s"'
-      self-hosted-runner: true
-      self-hosted-runner-label: stg-private-endpoint
-      test-timeout: 90
```
cbartz, line 30: no reason for those, they are also executed on private endpoint runners

.jujuignore
```diff
@@ -6,3 +6,4 @@ tests
 src-docs
 docs
 github-runner-operator
+clouds.yaml
```
cbartz, line 9: Avoid putting this in when building the charm locally

docs/changelog.md
```diff
@@ -31,9 +31,9 @@ network/infrastructure specific things.
 
 ### 2025-05-06
 
-- The ssh health checks are removed and the platform providers (GitHub or the JobManager) are used instead to get the runners health
+- The ssh health checks are removed and GitHub is used instead to get the runners health
 information. This implies many changes in both the structure of the project and its functionality. Potentially, many race conditions should
-disappear for the GitHub case. 
+disappear.
```
cbartz, line 36: remove jobmanager in changelog as this is still "hidden" and not well exposed to the user

github-runner-manager/src/github_runner_manager/platform/multiplexer_provider.py
cbartz, line 1: In theory we could also drop the multiplexer provider, as, it is now, the charm will never configure both platforms the same time. But we haven't yet decided if we want to do this in the future, and removing it is more effort than keeping, therefore I didn't drop.

github-runner-manager/src/github_runner_manager/reactive/consumer.py
```diff
@@ -178,8 +178,7 @@ def _build_runner_metadata(job_url: str) -> RunnerMetadata:
         logger.error("Invalid URL for a job. url: %s", job_url)
         raise ValueError(f"Invalid format for job url {job_url}")
     base_url = parsed_url._replace(path=match_result.group(1)).geturl()
-    runner_id = match_result.group(2)
```
cbartz, line 181: we no longer use the job id as runner id

jobmanager/client/.gitlab-ci.yml
cbartz, line 1: All the files under `jobmanager/client` are autogenerated and you shouldn't take effort to review those ;-)

src/charm.py
```diff
@@ -474,6 +480,24 @@ def _on_image_relation_changed(self, _: ops.RelationChangedEvent) -> None:
...
+        """Handle the MongoDB endpoints changed event."""
+        state = self._setup_state()
+        self._setup_service(state)
+
+    @catch_charm_errors
+    def _on_mongodb_relation_broken(self, _: ops.RelationDepartedEvent) -> None:
+        """Handle the MongoDB relation broken event."""
+        state = self._setup_state()
+        self._setup_service(state)
+
```
cbartz, line 500: necessary, otherwise the app will not be reconfigured on mongodb changes. especially important when we want to remove reactive mode.

src/charm_state.py
```diff
@@ -111,12 +110,17 @@ def from_charm(cls, charm: CharmBase) -> "GithubConfig | None":
...
         if not path_str:
             raise CharmConfigInvalidError(f"Missing {PATH_CONFIG_NAME} configuration")
+
+        # check if path_str is an url using pydantic
+        if path_str.startswith("http://") or path_str.startswith("https://"):
+            logger.info(
+                "Detected URL in %s configuration, will use experimental jobmanager mode",
+                PATH_CONFIG_NAME,
+            )
+            return None
```
cbartz, line 122: we now use the path for the "hidden" magic to detect the jobmanager mode and use the path for the jobmanager url until we have a proper design

tests/integration/conftest.py
```diff
@@ -743,3 +746,31 @@ async def instance_helper_fixture(request: pytest.FixtureRequest) -> OpenStackIn
     """Instance helper fixture."""
     openstack_connection = request.getfixturevalue("openstack_connection")
     return OpenStackInstanceHelper(openstack_connection=openstack_connection)
+
+
+@pytest.fixture(scope="module")
+def juju(request: pytest.FixtureRequest, model: Model) -> Generator[jubilant.Juju, None, None]:
```
cbartz, line 752: lets start using jubilant. stole this fixture from https://github.com/canonical/paas-charm ;-)

tests/integration/test_charm_scheduled_events.py
cbartz, line 1: we no longer use scheduled events and this test is basically already done in `test_reconcile_runners` in `test_charm_no_runner` .

cbartz (commented):
self-review

github-runner-manager/requirements.txt
```diff
@@ -6,4 +6,4 @@ openstacksdk==4.5.0
 pydantic < 2
 pymongo==4.13.0
 click==8.2.1
-flask==3.1.1
+flask==3.1.1
```
yhaliaw, line 9: Why remove the newline at the end of line?

cbartz: by accident

github-runner-manager/src/github_runner_manager/platform/__init__.py
```diff
@@ -2,3 +2,18 @@
...
+
+    Attributes:
+        GITHUB: GitHub platform.
+        JOBMANAGER: JobManager platform.
+        MULTIPLEXER: Multiplexer platform that uses several providers simultaneously.
+    """
+
+    GITHUB = "github"
+    JOBMANAGER = "jobmanager"
+    MULTIPLEXER = "multiplexer"
```
yanksyoon, line 19: I think `MULTIPLEXER` belongs rather to the `Mode` rather than the `Platform` enum class.

Perhaps we can create another dataclass for which `Mode` the application is running in, or, we make the "Job" of application be multiplexed by default in the future, and define what platform the job should be spawned for.

github-runner-manager/src/github_runner_manager/platform/jobmanager_provider.py
```diff
@@ -159,17 +177,29 @@ def get_runner_context(
...
+                    jobmanager_client.RegisterRunnerV1RunnerRegisterPostRequest(
+                        name=instance_id.name, labels=labels
+                    )
+                )
+
+                response = api_instance.register_runner_v1_runner_register_post(
+                    runner_register_request
+                )
+                if not response.id:
+                    raise PlatformApiError("No runner ID from jobmanager API")
```
yanksyoon, line 194: I'm wondering when this case is possible, that we ask it to create a resource and it does not return a resource ID. Should the API be returning 5XX error in this case?

yanksyoon (commented):
Reviewed until `reactive/types_.py`, LGTM so far. I'm wondering though, if we should be adding `MultiplexerProvider` as a provider - rather we should keep the platform as concrete platforms and multiplexer as a layer in which we create jobs. WDYT?

github-runner-manager/src/github_runner_manager/platform/jobmanager_provider.py
```diff
@@ -213,13 +243,26 @@ def check_job_been_picked_up(self, metadata: RunnerMetadata, job_url: HttpUrl) -
         Returns:
             True if the job has been picked up, False otherwise.
         """
-        configuration = jobmanager_client.Configuration(host=metadata.url)
+        configuration = jobmanager_client.Configuration(host=self._url)
+        #
```
yhaliaw, line 247: Why empty line comment?

github-runner-manager/src/github_runner_manager/platform/jobmanager_provider.py
```diff
@@ -213,13 +243,26 @@ def check_job_been_picked_up(self, metadata: RunnerMetadata, job_url: HttpUrl) -
...
+        # job_url has the path:
+        # "/v1/job/<job_id>"
```
yhaliaw, line 249: I think this comment should be documented in the caller function.
From the `check_job_been_picked_up` method, what is passed in as `job_url` is transparent to this function.

javierdelapuente (approved):

github-runner-manager/src/github_runner_manager/platform/__init__.py
cbartz: Also after discussion with @javierdelapuente , I've moved this enum to `multiplexer_provider.py`, as it makes only sense in the context of the multiplexer. Does this solve your concern, too?

github-runner-manager/src/github_runner_manager/platform/jobmanager_provider.py
simonepelosi: Updated it, ValidationError from FastAPI is returning 422.

yhaliaw (approved):

PR #569, opened 2025-06-16
fix: revert copy truncate

cbartz:
Applicable spec: <link>

### Overview

Revert copytruncate logrotate config for reactive processes.

### Rationale

copytruncate does not remove the files on the disk leading to a lot of zero sized files on the disk which leads to issue with the grafana agent.

### Juju Events Changes

n/a

### Module Changes

n/a

### Library Changes

n/a

### Checklist

### DIFF
--- a/src/logrotate.py
+++ b/src/logrotate.py
@@ -52,7 +52,7 @@ class LogrotateConfig(BaseModel):
     log_path_glob_pattern: str
     rotate: int
     create: bool
-    copytruncate: bool = False
+    copytruncate: bool = True
     notifempty: bool = True
     frequency: LogrotateFrequency = LogrotateFrequency.WEEKLY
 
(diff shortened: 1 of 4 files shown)


yhaliaw (approved):

src/logrotate.py
```diff
@@ -134,7 +137,7 @@ def _write_config(logrotate_config: LogrotateConfig) -> None:
         f"""{logrotate_config.log_path_glob_pattern} {{
 {logrotate_config.frequency}
 rotate {logrotate_config.rotate}
-copytruncate
+{"copytruncate" if logrotate_config.copytruncate else "nocopytruncate"}
```
yanksyoon, line 140: Would it be helpful to leave a comment on why copytruncate was used in the first place? (I don't quite remember why, I'm thinking whether we should remove `copytruncate` completely)

yanksyoon (approved):
Minor nitpicky question to help me understand, LGTM thank you!

src/logrotate.py
cbartz: I think @yhaliaw added it , probably because the external application doesn't implement a HUP to reload on a logrotate and would write to a deleted file. I also think we should move to WatchedFileHandlers,but for reactive processes this needs redesign and I don't want to tackle this in this PR.

PR #580, opened 2025-06-19
Forward arbitrary TCP traffic using aproxy

weiiwang01:
Applicable spec: <link>

### Overview

Update the GitHub runner to allow forwarding arbitrary TCP traffic to an HTTP proxy using aproxy. 

Adds two configuration options, `aproxy-exclude-addresses` and `aproxy-redirect-ports`, to filter which TCP traffic will be redirected to aproxy.

### Checklist


github-runner-manager/src/github_runner_manager/configuration/base.py
```diff
@@ -91,6 +94,8 @@ class SupportServiceConfig(BaseModel):
     proxy_config: "ProxyConfig | None"
     runner_proxy_config: "ProxyConfig | None"
     use_aproxy: bool
+    aproxy_exclude_addresses: str | None
+    aproxy_redirect_ports: str | None
```
cbartz, line 98: Would it make sense to model this as a python list? For me, the value should be independent of the output format.

src/charm_state.py
```diff
@@ -514,6 +521,8 @@ def from_charm(cls, charm: CharmBase) -> "CharmConfig":
             token=github_config.token if github_config else None,
             manager_proxy_command=manager_proxy_command,
             use_aproxy=use_aproxy,
+            aproxy_exclude_addresses=charm.config.get(APROXY_EXCLUDE_ADDRESSES_CONFIG_NAME),
+            aproxy_redirect_ports=charm.config.get(APROXY_REDIRECT_PORTS_CONFIG_NAME),
```
cbartz, line 525: we should validate the format somewhere to block directly , e.g. if an operator specifies an invalid format

cbartz (commented):
Thank you! Some comments regarding validation.

src/charm_state.py
weiiwang01: Yes, added some validation to these two fields, thanks!

cbartz (approved):

src/charm_state.py
```diff
@@ -3,7 +3,10 @@
 
 """State of the Charm."""
 
+# pylint: disable=too-many-lines
```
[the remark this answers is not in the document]
weiiwang01: I moved some of the data classes and type definitions from `charm_state.py` to `models.py`, making that file just under 1000 lines. What do you think of the change? If you don’t like it, I can revert the last commit.

weiiwang01:
Temporarily remove the following TCP proxy from the aproxy test in the end-to-end test suite, as the upstream proxy in the test environment is returning a 503 response. I [ran the test in the PS7 environment](https://github.com/canonical/aproxy/actions/runs/15870115150/job/44752361163?pr=28#step:13:7) and it worked without any problems.

```
          printf "" | timeout 60 nc github.com 22 | head -n 1 | grep SSH
```

cbartz:
We have issues with the CI setup, all test passed except the `test_runner_manager_openstack`. I ran it locally successfully
image
Will force merge the PR.

PR #582, opened 2025-06-23
[WIP] feat: prometheus metrics

yanksyoon:
Applicable spec: <link>

### Checklist

- [ ] The [charm style guide](https://juju.is/docs/sdk/styleguide) was applied.
- [ ] The [contributing guide](https://github.com/canonical/is-charms-contributing-guide) was applied.
- [ ] The changes are compliant with [ISD054 - Managing Charm Complexity](https://discourse.charmhub.io/t/specification-isd014-managing-charm-complexity/11619)
- [ ] The PR is tagged with appropriate label (`urgent`, `trivial`, `complex`).


scripts/setup-integration-tests.sh
yanksyoon, line 1: Moved to tests/integration/setup-integration-tests.sh

github-runner-manager/src/github_runner_manager/manager/runner_scaler.py
```diff
@@ -243,6 +248,8 @@ def get_runner_info(self) -> RunnerInfo:
                     offline += 1
                 case _:
                     unknown += 1
+        BUSY_RUNNERS_COUNT.labels(self._manager.manager_name).set(busy)
+        IDLE_RUNNERS_COUNT.labels(self._manager.manager_name).set(online)
```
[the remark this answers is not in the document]
yanksyoon: Moved, thanks!

yhaliaw (approved):

github-runner-manager/src/github_runner_manager/manager/runner_manager.py
```diff
@@ -354,6 +353,7 @@ def _delete_cloud_runners(
                 logger.error("No metrics returned after deleting %s", cloud_runner.instance_id)
             else:
                 extracted_runner_metrics.append(runner_metric)
+        CLEANED_RUNNERS_TOTAL.labels(self.manager_name).inc(len(cloud_runners))
```
cbartz, line 356: Couldn't we inc one by one in the loop (after line 328)? Otherwise we would again increase the amount also for non-deleted runners (if there is an exception).

github-runner-manager/src/github_runner_manager/http_server.py
```diff
@@ -45,11 +46,13 @@ def check_runner() -> tuple[str, int]:
...
         Information on the runners in JSON format.
     """
-    app_config = app.config[APP_CONFIG_NAME]
+    app_config: ApplicationConfiguration = app.config[APP_CONFIG_NAME]
     app.logger.info("Checking runners...")
     runner_scaler = get_runner_scaler(app_config)
     try:
         runner_info = runner_scaler.get_runner_info()
+        BUSY_RUNNERS_COUNT.labels(app_config.name).set(runner_info.busy)
+        IDLE_RUNNERS_COUNT.labels(app_config.name).set(runner_info.online)
```
cbartz, line 55: couldn't we set those in reconcile? `check_runner` is only triggered by a manual action.

yanksyoon: get_runner_info is only called from here and nowhere else

yanksyoon: Then doesn't that also go against the value of having side effects as the previous comment?

cbartz: I'm not sure if I understand you correctly, but I would expect a method called `_issue_reconciliation_metric` to have side effects. However, I would expect a method like `get_runner_info` to simply return a value without side effects or changing the state.

cbartz:
We decided to remove the integration test `test_runner_manager` , as those tests are covered in other integration test and it prevented merging due to its flakiness.

PR #587, opened 2025-07-01
feat: jobmanager token authentication

cbartz:
Applicable spec: <link>

### Overview

Support HTTP Bearer Token authentication with the jobmanager and make it required.

### Rationale

Communication with the jobmanager requires authentication.

### Juju Events Changes

n/a

### Module Changes

- charm state for parsing
- created a new module for interaction with the jobmanager api in the application and extracted the api interaction logic out of the jobmanager_provider
- this new module sets the HTTP Bearer Authentication header

### Library Changes

n/a

### Checklist

jobmanager support is still considered hidden feature, so there is no official documentation

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -48,7 +48,7 @@ options:
       to 'aproxy-exclude-addresses' or exclude the HTTP proxy port from 'aproxy-redirect-ports'.
   aproxy-exclude-addresses:
     type: string
-    default: "10.0.0.0/8, 171.16.0.0/12, 192.168.0.0/16"
+    default: "10.0.0.0/8, 172.16.0.0/12, 192.168.0.0/16"
     description: >-
       A comma-separated list of IP addresses that should be excluded from redirection to aproxy.
       127.0.0.0/8 are always excluded so you can omit if from the configuration.
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -2,6 +2,10 @@
 
 This changelog documents user-relevant changes to the GitHub runner charm.
 
+## 2025-07-15
+
+- Fix the incorrect default value of the aproxy-exclude-addresses configuration.
+
 ## 2025-07-09
 
 - Specify max supported nova compute api to be 2.91. This fixes an issue where the charm could fail


github-runner-manager/src/github_runner_manager/platform/jobmanager_provider.py
```diff
@@ -73,35 +82,32 @@ def get_runner_health(
         Returns:
            The health of the runner in the jobmanager.
         """
-        configuration = jobmanager_client.Configuration(host=self._url)
-        with jobmanager_client.ApiClient(configuration) as api_client:
```
cbartz, line 77: We no longer use a context manager. This means, the urllib connection pool will be reused until the reconciliation process dies.

github-runner-manager/src/github_runner_manager/jobmanager_api.py
cbartz, line 1: We called this module `jobmanager_api` due to a naming clash with the `jobmanager_client`

github-runner-manager/src/github_runner_manager/jobmanager_api.py
```diff
@@ -0,0 +1,170 @@
...
+
+class RunnerHealth(BaseModel):
+    """Represents the health status of a runner on the JobManagerAPI.
+
+    Attributes:
+        status: The health status of the runner.
+        deletable: Indicates if the runner can be deleted.
+    """
+
+    status: str
```
cbartz, line 75: idea would be to use the `RunnerStatus` enum, but right now, the openapi spec doesn't provide an enumeration, so I cannot foresee the potential values. The enum is still used in the jobmanager_provider to distinguish on the states.

src/charm_state.py
```diff
@@ -158,9 +160,13 @@ def from_charm(cls, charm: CharmBase) -> "JobManagerConfig | None":
         if not url_str:
             raise CharmConfigInvalidError(f"Missing {PATH_CONFIG_NAME} configuration")
 
+        token_str = cast(str, charm.config.get(TOKEN_CONFIG_NAME))
+        if not token_str:
+            raise CharmConfigInvalidError(f"Missing {TOKEN_CONFIG_NAME} configuration")
```
cbartz, line 165: We now require token to always be set also for jobmanager.

src/charm.py
```diff
@@ -321,7 +321,7 @@ def _on_config_changed(self, _: ConfigChangedEvent) -> None:
         state = self._setup_state()
 
         flush_runners = False
-        if state.charm_config.token != self._stored.token:
+        if self.config[TOKEN_CONFIG_NAME] != self._stored.token:
```
cbartz, line 324: this is to prevent an unnecessary flush on config change , because `charm_config.token` doesn't contain the jobmanager token value.

github-runner-manager/src/github_runner_manager/platform/jobmanager_provider.py
cbartz: Have changed it so we again use context managers per call.

javierdelapuente (approved):

github-runner-manager/src/github_runner_manager/jobmanager_api.py
```diff
@@ -0,0 +1,191 @@
...
+
+    status: str | None
+
+
+class RunnerStatus(str, Enum):
+    """Status of a runner on the JobManager.
+
+    Attributes:
+        IN_PROGRESS: Represents a job that is in progress.
+        PENDING: Represents a job that is pending.
```
yhaliaw, line 48: Line 47, 48: The docstring mentions `a job`. I think it should be `a runner`.

github-runner-manager/src/github_runner_manager/jobmanager_api.py
```diff
@@ -0,0 +1,191 @@
...
+from jobmanager_client.exceptions import ApiException, NotFoundException
+from pydantic import BaseModel
+from urllib3.exceptions import RequestError
+
+
+class JobManagerAPIError(Exception):
+    """Base exception for JobManager API errors."""
+
+
+class JobManagerAPINotFoundError(JobManagerAPIError):
```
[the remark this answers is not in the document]
yanksyoon: May I know why we might want to consolidate the errors into a single module? I'm wondering if each error should stay within the scope of the module.

github-runner-manager/src/github_runner_manager/platform/jobmanager_provider.py
```diff
@@ -168,56 +174,43 @@ def get_runner_context(
...
-                updated_metadata = RunnerMetadata(
-                    platform_name=metadata.platform_name, url=self._url
+        try:
+            response = self._jobmanager_api.register_runner(name=instance_id.name, labels=labels)
+            updated_metadata = RunnerMetadata(platform_name=metadata.platform_name, url=self._url)
+            updated_metadata.runner_id = str(response.id)
+            if token := response.token:
+                jobmanager_endpoint = f"{self._url}/v1/runners/{updated_metadata.runner_id}/health"
+                # For now, use the first label
+                label = "undefined"
```
yhaliaw, line 186: What is this for?

github-runner-manager/src/github_runner_manager/platform/jobmanager_provider.py
```diff
@@ -168,56 +174,43 @@ def get_runner_context(
...
-                        "builder-agent"
-                    )
-                    return (
-                        RunnerContext(
-                            shell_run_script=command_to_run,
-                            ingress_tcp_ports=[8080],
+                return (
+                    RunnerContext(
+                        shell_run_script=command_to_run,
+                        ingress_tcp_ports=[8080],
```
yanksyoon, line 197: It looks like we're hardcoding 8080, should this be in the setup of cloud security group during the initialization of the application?

github-runner-manager/src/github_runner_manager/platform/jobmanager_provider.py
```diff
@@ -168,56 +174,43 @@ def get_runner_context(
...
-                            status=GitHubRunnerStatus.OFFLINE,
-                        ),
-                    )
-                raise PlatformApiError("Empty token from jobmanager API")
-            except (ApiException, RequestError, ValidationError) as exc:
-                logger.exception("Error calling jobmanager api.")
-                raise PlatformApiError("API error") from exc
+                        busy=False,
+                        id=int(updated_metadata.runner_id),
+                        labels=[SelfHostedRunnerLabel(name=label) for label in labels],
```
yanksyoon, line 206: Is the `SelfHostedRunnerLabel` class necessary? The only value inside is the "name" parameter, we should probably just use it as a type, i.e.
`SelfHostedRunnerLabel = str`

github-runner-manager/src/github_runner_manager/jobmanager_api.py
```diff
@@ -0,0 +1,191 @@
...
+        Returns:
+            RunnerRegistration: The registration details of the runner, including ID and token.
+
+        Raises:
+            JobManagerAPIError: If there is an error registering the runner.
+        """
+        # The job manager api uses an autogenerated api client that uses urllib3 connection pools
+        # that are not multiprocessing safe: https://github.com/urllib3/urllib3/issues/850
+        # Therefore, we create a new ApiClient for each request and close resources
+        # to avoid issues with multiprocessing in the application.
```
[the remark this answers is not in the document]
cbartz: I thought about this initially, but I think its a decision the caller and not the callee does. So where would the reader of the code would look for it? 

To avoid duplication I suggest adding this comment to the top of the class.

github-runner-manager/src/github_runner_manager/jobmanager_api.py
```diff
@@ -0,0 +1,191 @@
...
+
+
+class Job(BaseModel):
+    """Represents a job on the JobManagerAPI.
+
+    Attributes:
+        status: The status of the job.
+    """
+
+    status: str | None
```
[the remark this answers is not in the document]
cbartz: Same as for the RunnerStatus, I originally thought using `JobStatus`, but I have no enumeration from `openapi.json` , so I cannot guarantee for any potential statuses (cc @simonepelosi)

github-runner-manager/src/github_runner_manager/jobmanager_api.py
```diff
@@ -0,0 +1,191 @@
+#  Copyright 2025 Canonical Ltd.
```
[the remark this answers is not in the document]
cbartz: Main value is to provide an API abstraction to  `jobmanager_provider` , similar what the `github_client` is doing.

github-runner-manager/src/github_runner_manager/platform/jobmanager_provider.py
cbartz: I think the way this was done was historically necessary to dynamically support JobManager vs. GitHub mode depending on the messages in the queue. We could do it during application initialization now since the mode is predetermined. However, I think it's out of scope for this PR.

yanksyoon: Fine with me, I'll add this to the refactor notes.

github-runner-manager/src/github_runner_manager/jobmanager_api.py
```diff
@@ -0,0 +1,191 @@
...
+        """
+        # The job manager api uses an autogenerated api client that uses urllib3 connection pools
+        # that are not multiprocessing safe: https://github.com/urllib3/urllib3/issues/850
+        # Therefore, we create a new ApiClient for each request and close resources
+        # to avoid issues with multiprocessing in the application.
+        with self._create_api_client() as api_client:
+            jobs_api = jobmanager_client.JobsApi(api_client=api_client)
+            try:
+                response = jobs_api.get_job_v1_jobs_job_id_get(job_id)
+            except NotFoundException as err:
```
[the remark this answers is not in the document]
yanksyoon: Just to help me with my learning - why does it lead to flakier code here? Also since this is a API wrapper, it seems like an appropriate place to do so?

github-runner-manager/src/github_runner_manager/platform/jobmanager_provider.py
```diff
@@ -262,16 +253,13 @@ def check_job_been_picked_up(self, metadata: RunnerMetadata, job_url: HttpUrl) -
...
-            except (ApiException, RequestError, ValidationError) as exc:
-                logger.exception("Error calling jobmanager api to get job information.")
-                raise PlatformApiError("API error") from exc
+        try:
+            job = self._jobmanager_api.get_job(job_id)
+            if job.status != JobStatus.PENDING:
+                return True
+        except JobManagerAPIError as exc:
+            logger.exception("Error calling jobmanager api to get job information.")
+            raise PlatformApiError("API error") from exc
```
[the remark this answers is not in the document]
yanksyoon: > Currently, the reactive consumer process errors out in this case.

I'm wondering if we should merge the code given that the code paths do allow this case to happen. I'm afraid we're adding undetermined code that doesn't cover all the cases which is a clear technical debt.

Furthermore, we're going to be catching the NotFound exception in the job call API and also raise it as PlatformApiError. May I suggest that we handle this more gracefully? I don't see anywhere in this PR that handles PlatformAPIError, which means that our application will crash when querying for jobs.

Given the situation above, what I can think of is to add the error handler for JobManagerAPIError from the caller if you insist that we should not be returning `None` for not found jobs.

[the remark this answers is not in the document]
yanksyoon: Sorry - misunderstood your comment. Perhaps the only question I have left is whether we should handle the `JobNotFound` error separately from the `PlatformAPIError`.

Also, I believe that the `JobAPIError` should inherit from the generic `PlatformAPIError`? This is because the specific implementations should inherit from the interfacing generics.

yanksyoon (commented):
A minor comment on inheritance, thanks!

src/models.py
```diff
@@ -19,18 +18,6 @@ class AnyHttpsUrl(AnyHttpUrl):
     allowed_schemes = {"https"}
 
 
-class Arch(str, Enum):
```
cbartz, line 22: no longer used

src/charm_state.py
```diff
@@ -708,41 +714,6 @@ def _build_runner_proxy_config_from_charm(charm: CharmBase) -> "ProxyConfig":
...
-        Arch: Current machine architecture.
-    """
-    arch = platform.machine()
-    match arch:
-        case arch if arch in ARCHITECTURES_ARM64:
-            return Arch.ARM64
-        case arch if arch in ARCHITECTURES_X86:
-            return Arch.X64
-        case _:
-            raise UnsupportedArchitectureError(arch=arch)
```
cbartz, line 743: no longer used

github-runner-manager/src/github_runner_manager/platform/jobmanager_provider.py
```diff
@@ -38,13 +39,15 @@
 class JobManagerPlatform(PlatformProvider):
     """Manage self-hosted runner on the JobManager."""
 
-    def __init__(self, url: str):
+    def __init__(self, url: str, jobmanager_api: JobManagerAPI):
```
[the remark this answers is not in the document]
cbartz: Done

github-runner-manager/src/github_runner_manager/platform/jobmanager_provider.py
```diff
@@ -168,56 +165,43 @@ def get_runner_context(
...
-                raise PlatformApiError("Empty token from jobmanager API")
-            except (ApiException, RequestError, ValidationError) as exc:
-                logger.exception("Error calling jobmanager api.")
-                raise PlatformApiError("API error") from exc
+                        busy=False,
+                        id=int(updated_metadata.runner_id),
+                        labels=labels,
+                        status=GitHubRunnerStatus.OFFLINE,
+                    ),
+                )
```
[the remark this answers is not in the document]
cbartz: Lets consider this in another PR

github-runner-manager/tests/unit/platform/test_multiplexer_provider.py
```diff
@@ -138,7 +138,7 @@ def test_get_runners_health_returns_non_requested_runners_always(
         ),
         pytest.param(
             None,
-            JobManagerConfiguration(url="http://jobmanager.example.com"),
+            JobManagerConfiguration(url="http://jobmanager.example.com", token="token"),
```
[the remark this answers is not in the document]
cbartz: Lets consider this in another PR

github-runner-manager/tests/unit/test_jobmanager_api.py
```diff
@@ -0,0 +1,258 @@
+# Copyright 2025 Canonical Ltd.
```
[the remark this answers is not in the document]
cbartz: Are you referring to a simple Fake implementation of the  autogenerated jobmanager_client? Because the JobManagerAPI is the class under test, so I don't think I should write a fake for the class under test.

github-runner-manager/pyproject.toml
```diff
@@ -80,7 +80,7 @@ select = ["E", "W", "F", "C", "N", "R", "D", "H"]
 # Ignore D107 Missing docstring in __init__
 ignore = ["W503", "D107", "E203"]
 # D100, D101, D102, D103, D104: Ignore docstring style issues in tests
-per-file-ignores = ["tests/*:D100,D101,D102,D103,D104,D205,D212"]
+per-file-ignores = ["tests/*:D100,D101,D102,D103,D104,D205,D212, DCO020, DCO030, DCO050"]
```
yhaliaw, line 83: Should there be a comment explaining why DCO020, DCO030, DCO050 are ignored?

yhaliaw (approved):

github-runner-manager/pyproject.toml
cbartz: Updated

PR #595, opened 2025-07-15
[ISD-3570] Flush runner will kill reactive processes as well

yhaliaw:
Applicable spec: <link>

### Overview

- Flush runner will kill the reactive process.
- Retried reactive job messages are delayed by 60 secs before processing.
- Flush are done on upgrade charm.

### Rationale

- Fix the issue where the old reactive process persist after a charm upgrade.
- Fix rapid retry of reactive job messages on error.

### Checklist


github-runner-manager/src/github_runner_manager/reactive/consumer.py
```diff
@@ -125,6 +127,14 @@ def consume(  # noqa: C901
                 if msg.payload == END_PROCESSING_PAYLOAD:
                     msg.ack()
                     break
+
+                msg.headers[PROCESS_COUNT_HEADER_NAME] = (
+                    msg.headers.get(PROCESS_COUNT_HEADER_NAME, 0) + 1
+                )
```
yanksyoon, line 133: Out of curiosity, does setting the message headers get reflected back in the queue? If not, should we keep a local variable for it?

The reason is because this snippet of code seems to suggest that the message will be modified - which is going to be shared over other workers that receive the message, but I don't quite see that happening further down the code.

github-runner-manager/src/github_runner_manager/reactive/runner_manager.py
```diff
@@ -128,3 +128,8 @@ def reconcile(
     )
 
     return ReconcileResult(processes_diff=processes_created, metric_stats=metric_stats)
+
+
+def flush_reactive_processes() -> None:
+    """Flush all the reactive processes."""
+    process_manager.kill_reactive_processes()
```
yanksyoon, line 135: I'm wondering if you want to make a classes for this pipeline and inject the dependencies, 
i.e. ReactiveManager(process_manager: ProcessManager)

The reason is because during testing, monkeypatching receives a string of things to patch and cannot be refactored easily with the help of IDE's intelliJ. Also, having the clear dependencies of classes will help us see the domain components easily - to help see the workflow clearer.

github-runner-manager/src/github_runner_manager/reactive/consumer.py
yanksyoon: As I see it - I don't quite see a case where we reject the message with a requeue. Would this be intended? Or would this be a preparation for the future case where we would requeue?

javierdelapuente (approved):

github-runner-manager/src/github_runner_manager/reactive/runner_manager.py
yhaliaw: - I feel using this style only due to testing, would be a large change. This would need to be done at a team-level. 
- Under the current method the using mocking would be `monkeypatch.setattr("github_runner_manager.reactive.runner_manager.kill_reactive_processes", mock_func)`. I think this not worse that the functional programming injection, since both mock at the function-level.

I think we need to bring this up in team discussion. I have added this to agenda.

github-runner-manager/src/github_runner_manager/reactive/consumer.py
yhaliaw: The rejection is done at two places:
- sigterm signal handler.
- In the spawning runner function at the of this function. It would wait for the reactive job to be picked up, if not it gets requeued.

PR #752, opened 2026-03-16
fix: update queue duration calculation

yanksyoon:
Applicable spec: <link>

### Checklist

- [ ] The [charm style guide](https://juju.is/docs/sdk/styleguide) was applied.
- [ ] The [contributing guide](https://github.com/canonical/is-charms-contributing-guide) was applied.
- [ ] The changes are compliant with [ISD054 - Managing Charm Complexity](https://discourse.charmhub.io/t/specification-isd014-managing-charm-complexity/11619)
- [ ] The PR is tagged with appropriate label (`urgent`, `trivial`, `complex`).
- [ ] The application version number is updated in `github-runner-manager/pyproject.toml`.

### DIFF
--- a/github-runner-manager/src/github_runner_manager/github_client.py
+++ b/github-runner-manager/src/github_runner_manager/github_client.py
@@ -464,7 +464,18 @@ class GithubClient:
         # datetime strings should be in ISO 8601 format, but they can also use Z instead of +00:00,
         # which is not supported by datetime.fromisoformat
         created_at = datetime.fromisoformat(job["created_at"].replace("Z", "+00:00"))
-        started_at = datetime.fromisoformat(job["started_at"].replace("Z", "+00:00"))
+        queued_at_raw = job.get("queued_at")
+        queued_at = (
+            datetime.fromisoformat(queued_at_raw.replace("Z", "+00:00"))
+            if queued_at_raw
+            else None
+        )
+        started_at_raw = job.get("started_at")
+        started_at = (
+            datetime.fromisoformat(started_at_raw.replace("Z", "+00:00"))
+            if started_at_raw
+            else None
+        )
         # conclusion could be null or an empty dictionary per api schema, so we need to handle
         # that though we would assume that it should always be present, as the job should be
         # finished.
@@ -474,6 +485,7 @@ class GithubClient:
         job_id = job["id"]
         return JobInfo(
             job_id=job_id,
+            queued_at=queued_at,
             created_at=created_at,
             started_at=started_at,
             conclusion=conclusion,
--- a/github-runner-manager/src/github_runner_manager/metrics/runner.py
+++ b/github-runner-manager/src/github_runner_manager/metrics/runner.py
@@ -550,7 +550,11 @@ def _issue_runner_start(
         else float("inf")
     )
     RUNNER_IDLE_DURATION_SECONDS.labels(flavor).observe(idle_duration)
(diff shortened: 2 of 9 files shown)


github-runner-manager/src/github_runner_manager/github_client.py
```diff
@@ -464,7 +464,18 @@ def _to_job_info(job: dict) -> JobInfo:
         # datetime strings should be in ISO 8601 format, but they can also use Z instead of +00:00,
         # which is not supported by datetime.fromisoformat
         created_at = datetime.fromisoformat(job["created_at"].replace("Z", "+00:00"))
-        started_at = datetime.fromisoformat(job["started_at"].replace("Z", "+00:00"))
+        queued_at_raw = job.get("queued_at")
+        queued_at = (
+            datetime.fromisoformat(queued_at_raw.replace("Z", "+00:00"))
+            if queued_at_raw
+            else None
+        )
```
yhaliaw, line 472: The GitHub Actions Jobs REST API ("Get a job for a workflow run") doesn't return a `queued_at` field at all — its response schema only has `created_at`/`started_at`/`completed_at`. That means `job.get("queued_at")` here always evaluates to `None` in production.

As a result, `queued_or_created_at = job_info.queued_at or job_info.created_at` in `metrics/github.py:53` will always fall back to `created_at`, so the queue-duration value is unchanged from before this PR. The PR's stated goal — measuring queue duration from the true queue-entry timestamp — isn't actually achieved; only the missing-`started_at` handling is a real behavioral change. This is easy to miss because every unit test injects a synthetic `queued_at` key into the mocked response, so the suite can't detect the divergence from the real API shape.

Either drop the `queued_at` plumbing and scope this PR to the missing-`started_at` fix, or call out explicitly in the PR description that GitHub doesn't currently expose this field and the plumbing is forward-looking for a future data source (e.g. a `workflow_job` webhook payload, if that's the intent).

*🤖 AI-assisted*

github-runner-manager/src/github_runner_manager/metrics/runner.py
```diff
@@ -614,14 +618,18 @@ def _create_runner_start(
...
             "Queue duration for runner %s is negative: %f. Setting it to zero.",
             runner_metrics.instance_id,
             job_metrics.queue_duration,
         )
-    queue_duration = max(job_metrics.queue_duration, 0) if job_metrics else None
+    queue_duration = (
+        max(job_metrics.queue_duration, 0)
+        if job_metrics and job_metrics.queue_duration is not None
+        else None
+    )
```
yhaliaw, line 632: The new `queue_duration is None` branches in `_issue_runner_start` (falls back to `float("inf")`) and `_create_runner_start` (falls back to `None`, skipping the negative-clamp/warning) have no test coverage. A regression here — e.g. reverting to `job_metrics.queue_duration if job_metrics else float("inf")` — would raise a `TypeError` on `None < 0` and wouldn't be caught by the existing suite, since every `GithubJobMetrics(...)` construction in `tests/unit/metrics/test_runner.py` currently passes a concrete `queue_duration`.

Could you add cases constructing `GithubJobMetrics(queue_duration=None, ...)`, asserting the `float("inf")` observation for `_issue_runner_start` and `RunnerStart.queue_duration is None` for `_create_runner_start`?

*🤖 AI-assisted*

yhaliaw (approved):

Note: the queued_at field may already exist on the GitHub Jobs API (mentioned previously) — if so, disregard the comment questioning whether it exists.

---

PR #790, opened 2026-04-20
test: prefer github app auth in integration fixtures

cbartz:
#### What this PR does

Teach the integration test fixtures to prefer GitHub App authentication when credentials are available, falling back to PAT otherwise. Along the way, clean up the surrounding plumbing:

- The shared `github_config` fixture (both root charm and `github-runner-manager`) returns a `GitHubConfig` populated with app credentials when all three of `GITHUB_APP_CLIENT_ID`, `GITHUB_APP_INSTALLATION_ID`, and `GITHUB_APP_PRIVATE_KEY` are present. Otherwise token auth is used. The ad-hoc override inside `test_e2e.py` is gone.
- Drop the `test_charm_fork_path_change` integration test and its workflow entry, along with the now-dead `INTEGRATION_TOKEN_ALT` / `--github-token-alt` / `alt_token` plumbing that only existed for that test.
- `INTEGRATION_TOKEN` and `GITHUB_APP_PRIVATE_KEY` are read from env vars only — no more `--github-token` pytest CLI flag — keeping secrets off the process command line.
- Replace the per-slot "which env var lives here" comments in the workflows with a pointer to `CONTRIBUTING.md`, which is the authoritative place for the slot-to-var mapping. The inline comments rot silently when the mapping is reshuffled in repo settings.
- Wire three additional secret slots (`INTEGRATION_TEST_SECRET_ENV_VALUE`, `_11`, `_12`) in `test_github_runner_manager.yaml` to carry the app credentials.
- `integration_test.yaml` now forwards `--github-app-client-id` from `vars.INTEGRATION_TEST_GITHUB_APP_CLIENT_ID`, matching the pattern already in `e2e_test.yaml`.

#### Why we need it

PATs are bound to individual people, so any rotation or role change risks breaking CI. Using a GitHub App is cleaner: credentials belong to the project, not a person, and the same auth path is exercised that the charm uses in production.

#### Checklist

- [x] I added or updated the documentation (if applicable)
- [x] I updated \`docs/changelog.md\` with user-relevant changes
- [ ] **If this is a Grafana dashboard:** I added a screenshot of the dashboard
- [ ] **If this is Terraform:** \`terraform fmt\` passes and \`tflint\` reports no errors
- [ ] **If the github-runner-manager application has been changed:** The application version number is updated in \`github-runner-manager/pyproject.toml\`.


github-runner-manager/tests/conftest.py
```diff
@@ -13,21 +13,21 @@ def pytest_addoption(parser):
         parser: Pytest parser.
     """
     parser.addoption(
-        "--github-token",
+        "--github-repository",
         action="store",
-        help="GitHub personal access token for integration tests.",
-        default=os.getenv("INTEGRATION_TOKEN"),
+        help="The GitHub repository in <owner>/<repo> format for integration tests.",
```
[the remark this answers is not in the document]
cbartz: GITHUB_REPOSITORY is a seductive default because Actions sets it automatically, but it names the repo running the workflow — not the target repo the integration tests should exercise. Our CI points the tests at a dedicated test repo through    
  --github-repository=${{ vars.INTEGRATION_TEST_PATH }}, which would be silently overridden if $GITHUB_REPOSITORY were a fallback. I'd rather keep missing config loud than mask it with the wrong repo. I've narrowed the docstring instead so it no  longer implies this option has an env fallback.

github-runner-manager/tests/integration/factories.py
```diff
@@ -54,7 +64,7 @@ class OpenStackConfig:
     auth_url: str
     project: str
     username: str
-    password: str
+    password: str = field(repr=False)
```
cbartz, line 67: security: never show password

github-runner-manager/tests/conftest.py
```diff
@@ -13,21 +13,21 @@ def pytest_addoption(parser):
         parser: Pytest parser.
     """
     parser.addoption(
-        "--github-token",
```
cbartz, line 16: security: don't allow passing --github-token via cli

tests/integration/test_charm_fork_path_change.py
cbartz, line 1: as discussed in the team - keeping this test is overkill , it comes with a too high cost (classic pats or github app with separate created repository) for the value it is bringing

tests/integration/test_charm_upgrade.py
```diff
@@ -31,15 +31,6 @@
 pytestmark = pytest.mark.openstack
 
 
-@pytest.mark.skip(
```
cbartz, line 34: skip is no longer needed - test_charm_upgrade should work again

tests/conftest.py
```diff
@@ -179,12 +179,6 @@ def pytest_addoption(parser: Parser):
         help="The GitHub App installation ID for GitHub App authentication testing.",
         default=os.environ.get("GITHUB_APP_INSTALLATION_ID"),
     )
-    parser.addoption(
```
cbartz, line 182: security: don't allow passing private key via cli

.github/workflows/test_github_runner_manager.yaml
```diff
@@ -50,6 +49,8 @@ jobs:
           ${{ vars.INTEGRATION_TEST_SECRET_ENV_NAME_8 }}: ${{ secrets.INTEGRATION_TEST_SECRET_ENV_VALUE_8 }}
           ${{ vars.INTEGRATION_TEST_SECRET_ENV_NAME_9 }}: ${{ secrets.INTEGRATION_TEST_SECRET_ENV_VALUE_9 }}
           ${{ vars.INTEGRATION_TEST_SECRET_ENV_NAME_10 }}: ${{ secrets.INTEGRATION_TEST_SECRET_ENV_VALUE_10 }}
+          ${{ vars.INTEGRATION_TEST_SECRET_ENV_NAME_11 }}: ${{ secrets.INTEGRATION_TEST_SECRET_ENV_VALUE_11 }}
+          ${{ vars.INTEGRATION_TEST_SECRET_ENV_NAME_12 }}: ${{ secrets.INTEGRATION_TEST_SECRET_ENV_VALUE_12 }}
```
yanksyoon, line 53: Operator supports upto ENV_VALUE_12, perhaps we should add another slot in the operator workflows first

yanksyoon (approved):
LGTM, thanks for catching the security related issues!

.github/workflows/test_github_runner_manager.yaml
cbartz: yeah, this test is bypassing operator workflows, this is why it works using the slot . for the charm integration test I am passing the client id via cli due to that restriction.

yhaliaw (approved):

PR #800, opened 2026-06-23
chore: apply recipe use-github-runners-for-tests

seb4stien:
Use Github Runners for simple tests

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -11,6 +11,4 @@ jobs:
     uses: canonical/operator-workflows/.github/workflows/test.yaml@main
     secrets: inherit
     with:
-      self-hosted-runner: true
-      self-hosted-runner-label: edge
       vale-style-check: false


.github/workflows/test.yaml
```diff
@@ -9,8 +9,10 @@ permissions:
...
   unit-tests:
     uses: canonical/operator-workflows/.github/workflows/test.yaml@main
+    permissions:
+      contents: read
+      pull-requests: write
     secrets: inherit
     with:
-      self-hosted-runner: true
-      self-hosted-runner-label: edge
+      runs-on-base: ubuntu-24.04
```
weiiwang01: I think this is probably unnecessary, as the default value is already `ubuntu-24.04`. Using the default can also make it so that the next time we update the `runs-on-base` value in the upstream `operator-workflows`, it will automatically update here.

seb4stien: Thanks. I checked for the default value, but did it on my local repo which was not updated.
I removed this change.

weiiwang01 (approved):

javierdelapuente (approved):

PR #802, opened 2026-06-26
fix(runner-manager): clean up VMs in error state immediately

cbartz:
#### What this PR does

Cleans up runners whose cloud VM is in an `ERROR` state immediately, instead of waiting for the creation timeout.

`_get_platform_runners_to_cleanup` previously decided cleanup purely from platform health (`online`/`busy`/`deletable`) plus VM age, never consulting the cloud `VMState`. An ERROR VM was therefore treated like one still booting and only collected once older than `RUNNER_MAXIMUM_CREATION_TIME` (~23 min). This adds an explicit rule: any requested runner whose mapped VM has `state == VMState.ERROR` is eligible for cleanup regardless of age.

#### Why we need it

An OpenStack VM reported as `ERROR` never recovers, but it still counted toward the desired runner total, so the slot stayed occupied with a dead VM for ~23 minutes and no replacement was created until cleanup finally evicted it. This leaves a flavor with zero usable capacity for that whole window.

#### Checklist

- [ ] I followed the [contributing guide](https://github.com/canonical/is-charms-contributing-guide)
- [x] I added or updated the documentation (if applicable) — updated `docs/changelog.md`
- [x] I updated `docs/changelog.md` with user-relevant changes
- [x] I used AI to assist with preparing this PR
- [x] I added or updated tests as needed (unit and integration) — added a unit case covering immediate cleanup of a recent ERROR-state VM
- [ ] **If this is a Grafana dashboard:** I added a screenshot of the dashboard — n/a
- [ ] **If this is Terraform:** `terraform fmt` passes and `tflint` reports no errors — n/a
- [x] **If the github-runner-manager application has been changed:** The application version number is updated in `github-runner-manager/pyproject.toml`. — bumped 0.18.1 → 0.18.2

PR #803, opened 2026-06-29
test(integration): de-flake otel pre-job config test

cbartz:
#### What this PR does

De-flakes `test_otel_collector_endpoint_pre_job_installs_config`. Instead of dispatching a quick workflow, waiting for it to *complete*, then SSHing into the runner, it now dispatches the long-running wait workflow and inspects the runner **while its job is in progress**. The config read is polled to absorb the timing of the pre-job hook write.

#### Why we need it

The runner is ephemeral, so its OpenStack VM is torn down on job completion. Reading the otel config *after* completion races the runner-manager cleanup loop — when cleanup wins, `get_single_runner` finds zero VMs and the test fails with `found more than one runners or no runners: []`. This was observed failing on the 22.04 base while passing on 24.04 in the same run (e.g. PR #802 CI). Keeping the job in progress guarantees the VM is alive during inspection.

#### Checklist

- [ ] I added or updated the documentation (if applicable) — N/A, test-only change
- [ ] I updated `docs/changelog.md` with user-relevant changes — N/A, no user-facing change
- [ ] **If this is a Grafana dashboard:** I added a screenshot of the dashboard — N/A
- [ ] **If this is Terraform:** `terraform fmt` passes and `tflint` reports no errors — N/A
- [ ] **If the github-runner-manager application has been changed:** version updated in `github-runner-manager/pyproject.toml` — N/A, not changed


yanksyoon (approved):

yhaliaw (approved):

tests/integration/test_charm_runner.py
```diff
@@ -216,21 +216,36 @@ def test_otel_collector_endpoint_pre_job_installs_config(
...
         branch=test_github_branch,
         github_repository=github_repository,
         conclusion="success",
-        workflow_id_or_name=DISPATCH_TEST_WORKFLOW_FILENAME,
-        dispatch_input={"runner": app},
+        workflow_id_or_name=DISPATCH_WAIT_TEST_WORKFLOW_FILENAME,
+        dispatch_input={"runner": app, "minutes": "5"},
+        wait=False,
     )
+    wait_for(lambda: workflow.update() or workflow.status == "in_progress")
```
yhaliaw: An AI has flagged the follwoing, which seems reasonable:

The lambda: workflow.update() or workflow.status == "in_progress" expression does not reliably wait for the run to reach in_progress. PyGithub's CompletableGithubObject.update() returns True whenever it performs a successful HTTP GET (200 OK) and False only on a 304 Not Modified. Because update() is the left operand of or, any 200 response short-circuits the expression before status is ever consulted. On the first loop iteration the run object was returned from a list call and has no stored etag, so the first GET is unconditional — it returns True immediately, regardless of what status actually is.

The guard therefore degrades to "refresh once and continue" rather than "wait until the job is running." This undercuts the invariant that lines 219–221 rely on — that the VM is guaranteed to be alive when the inspection begins. The downstream _read_otel_config poll (120 s) masks this in typical conditions, but the structural guarantee the PR intends is lost and the effective startup budget is reduced to a single wait.

The fix is to call update() for its side effect and evaluate status in a separate step:
```
def _in_progress() -> bool:
    workflow.update()
    return workflow.status == "in_progress"

 wait_for(_in_progress, timeout=300, check_interval=10)
```

yhaliaw (approved):
Approved

---

tests/integration/test_charm_runner.py
cbartz: Valid — `CompletableGithubObject.update()` returns True on a successful 200 GET, so the `or` short-circuited before `status` was checked. Both call sites now use the existing `wait_for_status` helper (refresh-then-check via `_has_workflow_run_status`). Also merged latest main into the branch.

**AI generated message**

PR #806, opened 2026-07-03
feat: pressure non streaming mode

yanksyoon:
#### What this PR does

Introduce non streaming mode for pressure

#### Why we need it

Flaky network infrastructure will always fail on pressure stream

#### Checklist

- [x] I added or updated the documentation (if applicable)
- [x] I updated `docs/changelog.md` with user-relevant changes
- [ ] **If this is a Grafana dashboard:** I added a screenshot of the dashboard
- [ ] **If this is Terraform:** `terraform fmt` passes and `tflint` reports no errors
- [ ] **If the github-runner-manager application has been changed:** The application version number is updated in `github-runner-manager/pyproject.toml`.


yhaliaw (approved):

github-runner-manager/src/github_runner_manager/planner_client.py
```diff
@@ -63,6 +63,42 @@ def __init__(self, config: PlannerConfiguration) -> None:
...
+            PlannerConnectionError: On transient connection failures or timeouts.
+            PlannerApiError: On HTTP errors or invalid payloads.
+        """
+        base = str(self._config.base_url).rstrip("/") + "/"
+        url = urljoin(base, f"api/v1/flavors/{name}/pressure")
+        try:
+            response = self._session.get(
+                url,
+                headers={"Authorization": f"Bearer {self._config.token}"},
+                timeout=self._config.timeout,
```
yhaliaw, line 85: `get_pressure()` uses `self._config.timeout`, whose default is `5 * 60` (300s, line 42) — a value chosen for long-lived streaming connections, not one-shot polls. `build_pressure_reconciler` constructs `PlannerConfiguration` without overriding it, so request mode inherits the streaming timeout unchanged.

A slow or hung planner can block the create loop for up to 300 seconds per poll attempt, compounded by the session's `urllib3.Retry(total=3, backoff_factor=0.3)` on 5xx responses. That makes the effective poll cadence wildly variable under exactly the flaky-network conditions this PR exists to handle, and delays the fallback-to-`min_pressure` behavior meant to keep runners flowing when the planner is unreachable.

Pass a short explicit timeout for the one-shot request path (a few seconds, consistent with the 5-second poll interval) — either a separate `request_timeout` field on `PlannerConfiguration` or an explicit argument at the `get_pressure` call site.

*🤖 AI-assisted*

tests/unit/factories.py
```diff
@@ -31,6 +31,7 @@
     OPENSTACK_FLAVOR_CONFIG_NAME,
     OPENSTACK_NETWORK_CONFIG_NAME,
     PATH_CONFIG_NAME,
+    PLANNER_PRESSURE_MODE_CONFIG_NAME,
     PLANNER_INTEGRATION_NAME,
```
yhaliaw, line 35: `PLANNER_PRESSURE_MODE_CONFIG_NAME` is imported before `PLANNER_INTEGRATION_NAME`, breaking the alphabetical ordering isort enforces here (every other name in this block is correctly ordered).

The `lint` tox env runs `isort --check-only --diff`, so this will fail CI lint as-is.

```suggestion
    PLANNER_INTEGRATION_NAME,
    PLANNER_PRESSURE_MODE_CONFIG_NAME,
```

*🤖 AI-assisted*

yhaliaw (approved):

---

PR #810, opened 2026-07-23
fix(test): wait for config-changed hook after juju.config() in set_app_runner_amount

yanksyoon:
## Problem

`test_check_runner` in `test_charm_runner.py` fails with `Timeout waiting for 2 runner(s) to be ready`. Investigation on the debug tmate session revealed:

- The service config file had `base_virtual_machines: 0` throughout the test
- Service logs showed `desired=0 current=0` every reconcile cycle
- Only one `config-changed` hook fired (at 15:34:20), none after the test called `juju.config()`
- `jubilant.Juju.config()` is fire-and-forget — it returns before the hook fires

The test polls `check-runners` while the service is still running with the previous `base_virtual_machines` value.

## Fix

Three changes in two files:

1. **Verify config was applied** — read back the config after `juju.config()` and assert
2. **Wait for model settle** — `juju.wait(all_active)` after config change so the hook completes
3. **Pre-poll grace period** — 30s sleep before first `check-runners` poll in `wait_for_runner_ready`

## Testing

- Unit tests: 214 passed
- Integration: [monitoring CI]


yanksyoon:
Re-enabled tmate debug for the integration test jobs to inspect the runner and OpenStack VM state during failures.

tests/integration/helpers/openstack.py
```diff
@@ -157,9 +157,26 @@ def set_app_runner_amount(self, app_name: str, num_runners: int) -> None:
...
+        logger.info("CLI config output after setting runner amount: %s", cli_output)
         if num_runners == 0:
             return
+        # Wait for the config-changed hook to complete and the service to be restarted
+        # with the new config. Without this, the test polls check-runners while the
+        # service is still running with the previous base_virtual_machines value.
+        self.juju.wait(
+            lambda status: jubilant.all_active(status, app_name),
+            timeout=60 * 5,
+        )
```
yhaliaw, line 179: `self.juju.wait(lambda status: jubilant.all_active(status, app_name), timeout=60*5)` doesn't guarantee `config-changed` has actually run before `wait_for_runner_ready` starts polling. The app is typically still `active` from before the config change, and `all_active` only inspects workload status — it can't distinguish "never left active" from "went through maintenance and came back."

Note that `jubilant.wait` already defaults to `successes=3` (three consecutive polls must satisfy the predicate) with `delay=1.0`, so this isn't a bare one-shot check — but that's still only a ~2s settle window. If the unit is already `active` when the wait starts and `config-changed` hasn't kicked off yet, three quick polls of "active" can pass well before the hook actually runs, leaving the same race this PR is meant to close, just narrower.

Worth asserting on something that actually reflects the new config having been applied — e.g. a workload status message set post-config-changed, or the rendered service config on the unit — rather than relying on `all_active` alone. The read-back in the block above (L162-167) doesn't help either, since it only confirms the value in the Juju controller's datastore, not that the charm consumed it.

*🤖 AI-assisted*

yhaliaw (approved):

---

